PR #1, opened 2015-09-04
Typo correction

eugeneloza:
GPU-compressed date -> GPU-compressed data


michaliskambi:
Merged, thank you.

This way I have also tested (and fixed:) the synchronization between GitHub GIT and SourceForge SVN to be two-way. So commits to GIT repository on https://github.com/castle-engine/castle-engine can be applied back to SourceForge SVN. The GIT commit https://github.com/castle-engine/castle-engine/commit/acc12f2534807a3e39967929b0c4adff75172d89 was automatically synchronized as https://sourceforge.net/p/castle-engine/code/14676/ .

PR #3, opened 2015-10-16
Joystick support stumb / untested

tsr8:
Foundation of joystick support based on zengl.

### DIFF
--- a/src/ui/castlejoysticks.pas
+++ b/src/ui/castlejoysticks.pas
@@ -1,0 +1,514 @@
+{
+  Copyright 2015 Tomasz Wojtyś.
+  Based on zgl_joystick.pas by Andrey Kemka.
+
+  This file is part of "Castle Game Engine".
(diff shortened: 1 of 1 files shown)


michaliskambi:
Cool! Merged, thank you!

Waiting for more!:)

Small notes:

1. I added notes clarifying the license stuff. ZenGL is on zlib license http://www.zengl.org/license.html , and we allow LGPL/GPL. So everything is cool, I just documented it clearly at the beginning of castlejoysticks.pas.

2. Nitpicking: remove the trailing whitespace:)

Note that because of our current two-way SVN<->GIT synchronization, the hashes of GIT commit may change after merging. In short, if you work on GitHub repo, be sure to switch to our master afterwards, and make new fork for new pull requests. (This discomfort on our side should disappear soon, we'll just make SVN read-only soon, and use GIT for main development.)

PR #5, opened 2015-10-29
Another bunch of joystick support code

tsr8:


michaliskambi:
Thank you, merged!

I made small corrections (see https://sourceforge.net/p/castle-engine/code/14895/ ) :

* do not add "{ TInputListener }" comment (I guess it was added by Lazarus automatically?). It's not useful and it hides the good comment "{ Base class for things..." that is used by PasDoc.

* I changed the code to call JoyAxisMove / JoyButtonPress on all controls to 
  1. check the result, and stop processing as soon as something retuns true, and
  2. call it recursively, as each TUIControl may now have children (this is a new thing in engine SVN, and I'm in the process of adjusting 100% of CastleUIControls to it).

Future idea: I would probably remove the JoyButtonPress event, instead I would extend TInputPressRelease object and TInputPressReleaseType to handle also joystick events. This allows to receive presses using the TUIControl.Press method, and eventually it allows to make inputs configurable (see CastleInputs and TInputShortcut object with it's IsEvent method). That's e.g. how http://castle-engine.sourceforge.net/castle.php allows to customize 100% of it's inputs. User could then assign any joystick button to any input.

But this can be deferred for later:)

michaliskambi:
There were quite a few changes to CastleUIControls today, related to passing events recursively to all controls. Be sure to get the latest version to work on :)

PR #8, opened 2015-11-15
Implemented rotation of TGLImage

tsr8:


michaliskambi:
Sorry for a delay in answering. I have some small comments about it before merging:

1. Using glTranslate/glRotate will only work in old OpenGL fixed-function pipeline. Right now this includes all desktops (but not mobile - Android, iOS), but eventually even modern desktops will be excluded from this (as we'll switch TGLImage rendering to use shaders even on desktops).

  The right solution is to rotate the 4 quad corners on CPU. (Or implement the same rotations as you do by glTranslate/glRotate inside the TGLImage shaders, but this seems like a worse solution.)

2. The API does not allow rotating images drawn by Draw3x3 or Draw3x1. Instead of adding new (optional) parameters to a *lot* of methods, I would prefer introducing instead properties on TGLImage - like CenterX, CenterY, Rotation (exactly like TCastleImageControl properties). Setting them should be extra-cheap, so e.g. TCastleImageControl.Render can just set them before every call to TGLImage.Draw.

3. Don't you think CenterX/Y should be by default (0.5, 0.5), not (0, 0)? I'm not sure about this, but a pivot in the middle seems more expected for me.

4. Since CenterX / CenterY / Rotation are only supported by TCastleImageControl, they should be defined at TCastleImageControl, not at TUIControl.

The 2 and 4 points are a must, IMHO... The 3rd --- I don't know, I'm not sure. The 1st point can be something to be finished/changed later, as it's an implementation detail.

tsr8:
Ad 1. I have to improve my knowelage about this :) Do you have some good materials?  Ad 2. OK.  Ad 3. At first I think that, but (0.5, 0.5) will influence current default behavior, becouse center of rotation is also a point of reference for drawing.  Ad 4. I thought about it as every control could be rotated but it will need to be implemented in controls too. When some component not support rotations then nothing wrong happens. Do you still want to move this properties from TUIControl? 
 

  
    Dnia 26 listopada 2015 01:36 Michalis Kamburelis &amp;lt;notifications@github.com&amp;gt; napisał(a):
  
  
    Sorry for a delay in answering. I have some small comments about it before merging:



Using glTranslate/glRotate will only work in old OpenGL fixed-function pipeline. Right now this includes all desktops (but not mobile - Android, iOS), but eventually even modern desktops will be excluded from this (as we&amp;#39;ll switch TGLImage rendering to use shaders even on desktops).

The right solution is to rotate the 4 quad corners on CPU. (Or implement the same rotations as you do by glTranslate/glRotate inside the TGLImage shaders, but this seems like a worse solution.)

The API does not allow rotating images drawn by Draw3x3 or Draw3x1. Instead of adding new (optional) parameters to a lot of methods, I would prefer introducing instead properties on TGLImage - like CenterX, CenterY, Rotation (exactly like TCastleImageControl properties). Setting them should be extra-cheap, so e.g. TCastleImageControl.Render can just set them before every call to TGLImage.Draw.
Don&amp;#39;t you think CenterX/Y should be by default (0.5, 0.5), not (0, 0)? I&amp;#39;m not sure about this, but a pivot in the middle seems more expected for me.
Since CenterX / CenterY / Rotation are only supported by TCastleImageControl, they should be defined at TCastleImageControl, not at TUIControl.


The 2 and 4 points are a must, IMHO... The 3rd --- I don&amp;#39;t know, I&amp;#39;m not sure. The 1st point can be something to be finished/changed later, as it&amp;#39;s an implementation detail.

michaliskambi:
> Ad 1. I have to improve my knowelage about this :) Do you have some good materials? 

The idea is to add a function like RotatePoint2D(Point, Center, Rotation) to calculate a rotation in 2D. Equations to do it should be on wikipedia and google, we also have a RotatePointAroundAxisDeg function in CastleVectors (but it's for 3D, you could probably implement something more efficient specific for 2D).

> Ad 3. At first I think that, but (0.5, 0.5) will influence current default behavior, becouse center of rotation is also a point of reference for drawing.

Oh, if that's the case, then this needs to be corrected. Center of rotation should *not* be a point of reference for drawing, in other words: when Rotation = 0, then Center values should have no visible effect in the image. (actually, the case of Rotation = 0 could also be specially optimized, since it will be true in majority of cases).

It should be possible by something as simple as

~~~~
if Rotation <> 0 then
begin
  glTranslatef(X, Y, 0);
  glRotatef(Rotation, 0, 0, 1);
  glTranslatef(-X, -Y, 0);
end;
~~~~

And then the default center could be (0.5,0.5) without any problems:)

> Ad 4. I thought about it as every control could be rotated but it will need to be implemented in controls too. 

That's a future plan. When it will happen (rotations will be implemented for other controls), we can move the CenterX/Y/Rotation properties to TUIControl. But until then, it's better to define them only where they are really supported, at TCastleImageControl.

Moving the properties from TCastleImageControl to TUIControl in the future is easy. And, until it happens, users will not be confused by CenterX/Y/Rotation properties on TUIControl, that only work with TCastleImageControl descendant:)

tsr8:
Ad 3. At first I think that, but (0.5, 0.5) will influence 
current default behavior, becouse center of rotation is also a point of 
reference for drawing.

     Oh, if that's the case, then this needs to be corrected. Center of 
rotation should /not/ be a point of reference for drawing, in other 
words: when Rotation = 0, then Center values should have no visible 
effect in the image. (actually, the case of Rotation = 0 could also be 
specially optimized, since it will be true in majority of cases).

It is only a matter of contractual. Not a problem to me.

Work in progress...

tsr8:
Required changes done.

michaliskambi:
Thank you!

I made small corrections in https://sourceforge.net/p/castle-engine/code/14971/ :

- The shift by CenterX/Y is done differently, to avoid breaking the case when GLImageUseShaders is defined (Android, iOS). We don't modify now PixCenterX/Y, whole shift now happens by

  ~~~~
glTranslatef( ShiftX,  ShiftY, 0);
glRotatef(FRotation, 0, 0, 1);
glTranslatef(-ShiftX, -ShiftY, 0);
~~~~

- I also made the Draw3x3 working with rotations nicely. One Draw3x3 call is 9 Draw calls, and they all must share the same center point (in screen coords) to rotate the whole image correctly.

- I also added VisibleChanged to the TCastleImageControl setters. This way we make sure the window is redrawn when Rotation or CenterX/Y changed.

Screenshot of examples/2d/controls_demo.lpr with everything rotated by 45 degrees attached:)
zrzut_ekranu-controls_demo
zrzut_ekranu-controls_demo-1

PR #11, opened 2015-12-07
Implemented dmBlendSmart drawing mode in TRGBAlphaImage

eugeneloza:


michaliskambi:
Excellent! Thank you! I applied your patch, and did small modifications:

- To use the same variables as other versions, so just use DestY, PSource, PDest etc., no need to define new iy, SourceAlphaColor,ThisAlphaColor for similar purpose.

- There's no need to check ranges, since DrawFrom() checks it always before calling DrawFromCore().

- I changed some style stuff:
  - indent by 2 spaces, not 1
  - start "begin" on new line, like
    ~~~~
if Blah then
begin
  
end;
~~~~
  
  - put spaces around operators like " := "

I also added better exception messages in my code to clearly report when the given Mode is not implemented for specified source/destination classes.

PR #12, opened 2015-12-07
Added TCastleSprite2D class

tsr8:
I have created nice class to run 2D animations called sprites.


michaliskambi:
Nice!

Last month, for my own game, I implemented something similar called TVideoAtlas. I did not manage yet to commit it to the engine. Your class has similar purpose, so I'll probably just abandon my TVideoAtlas, and merge my needs with yours:) For your curiosity, I attach my implementation of TVideoAtlas class, feel free to use it: [gamevideoatlas.pas.txt](https://github.com/castle-engine/castle-engine/files/56274/gamevideoatlas.pas.txt) .

Comments:

1. if not Assigned(FImage) then Exit; //todo: error 

  This line should be removed. The result of constructor (TGLImage.Create in line above) is never nil, this does not need to be checked. (And if if it could be nil, such errors should be signalled by raising an exception.)

2. Some renames/changes are needed, to make property names consistent with existing TVideo class (from CastleVideos unit) and TGLVideo (and TGLVideo2D, TGLVideo3D) classes from CastleGLImages unit:

  - Default FPS should be DefaultFramesPerSecond constant from CastleVideos unit, not 10. It should also be called FramesPerSecond, not FPS.

  - Looped should be renamed to TimeLoop.

  - ASmoothScaling should be true by default, or have no default value (be forced to be specified at constructor call). In previous engine releases, some stuff had SmoothScaling=false by default, and lately I came to the conclusion that it was a bad decision. SmoothScaling=false gains some speed (in theory, but in practice --- not really, even in brutal cases), but in 99% you actually want smooth scaling and it's too easy to forget it. So in next engine version (current SVN) SmoothScaling is either true by default or has to be specified explicitly.

  - Represent constantly changing time (FAnimTime) type as TFloatTime, not Double. Yeah, TFloatTime is just equal to Double, it's used throughout the engine (to make it clear *why* it's double --- because it needs to be, because it's constantly-updated time value, so precision matters for large time values and/or many accumulating time additions).

  - When calculating current frame in TCastleSprite2D.Update , use existing method TVideo.FrameIndexFromTime . It's also used by TGLVideo, and makes sure this is always calculated the same (for example rounding is the same), and works also for negative time, allows TimeBackwards option.

3. Feature requests:

  - Allow setting time explicitly. Simply adding a property "property AnimTime read FAnimTime write FAnimTime", to make FAnimTime writeable, seems ok for me. Also, I would rename it to just "Time" and "FTime", to keep names simple. Usecase: For example to reset the time to zero. Or to set it explicitly, like "Sprite.Time := Dragon.CurrentTime - Dragon.AnimationStartTime", when the sprite is shared across many other objects (Dragons in this example), and then keep track of time --- different for each dragon, so they just set the time before every render call.

  - You auto-calculate the FFrameWidth based on image width and columns count. While it's often comfortable, for some uses this automatic calculation is unhandy. 

    The usecase: When you use GPU compressed images, sometimes a rectangular image needs to be turned into square, and then the width of the image is expanded. At runtime, you decide (based on GPU capabilities) whether to use GPU compressed image or not. But the GPU-compressed version has wider width than the uncomressed, so you must pass the width explicitly to the class, not depend on autocalculated value.

    Proposed solution: hm, just make a property to set FFrameWidth and FFrameHeight ? Like you do already with FDrawingWidth / FDrawingHeight?

4. Finally, about the name TCastleSprite2D: don't use "TCastle" prefix (we usually don't use it, unless it's a component registered on Lazarus palette or something generic, e.g. TCastleImage is called like that because Lazarus already has TImage). I think that TSprite2D, or maybe even just TSprite, is simple and elegant:)

michaliskambi:
> Allow setting time explicitly.

Addendum: Setting "Time", in "SetTime", should also recalculate the current "FFrame" value. So that TCastleSprite2D.Update can be implemented simply as

~~~~
procedure TCastleSprite2D.Update(ATime: Single);
begin
  if Playing then
    Time := Time + ATime;
end;
~~~~

... and SetTime does the job of updating FFrame.

I'm not sure should the Frame value be settable. It desynchronizes the frame number with current time (and we cannot autosynchronize it, e.g. we cannot calculate Time from Frame, because the time may loop). Maybe it's safer to leave Frame property read-only? Up to you, depending on your usecases.

tsr8:
More features coming soon.

TVideo.FrameIndexFromTime takes time as Single. Shouldn't it be TFloatTime?

> I'm not sure should the Frame value be settable.
I think it should. You may don't want to use animation but just change to frame exactly you need to.

michaliskambi:
> TVideo.FrameIndexFromTime takes time as Single. Shouldn't it be TFloatTime?

It should:) Changed in the engine, thanks for pointing this!

>>  I'm not sure should the Frame value be settable.

> I think it should. You may don't want to use animation but just change to frame exactly you need to.

Good point, OK, let it be settable.

tsr8:
All your requests already met.

tsr8:
Added custom animations feature. Now is possible to make animations with frames any order and individual durations.

michaliskambi:
Excellent work, thank you! Merged.

PR #14, opened 2015-12-20
Tiled maps support

tsr8:
Created this request to show my progress with implementation of TMX files support (TCastleTiledMap component). Please review and comment. Decided to write it to separate .inc (like in CastleGLImages) file because CastleControls is so big.


michaliskambi:
Thoughts:

- Possibly, a separate unit for this would be suitable. Like CastleTileMap. Existing CastleControls unit is probably too large already:)

- Possibly, my initial suggestion to base this on TUIControl is not so simple. I mean, you want to base your map control on TUIControl. But in this case, it may be nice to have a 

  1. class like TTileMap, that *does not* contain render method at all and descends from standard TComponent, or even just standard TObject. It can read map from file.
  2. a class like TIsometricMap that uses assigned TTileMap instance for rendering.
  3. a class like TOrthogonalMap that uses assigned TTileMap instance for rendering,
  4. a class like THexagonalMap that uses assigned TTileMap instance for rendering.

  Only the classes TIsometricMap, TOrthogonalMap, THexagonalMap would descend from TUIControl, and use information in TTileMap for rendering. Or maybe (if it will be beneficial for implementation) they should all descend from TAbstractMap, and TAbstractMap descends from TUIControl.

  All this makes the implementation of course more complicated, but also more versatible. Looking at http://www.mapeditor.org/, it supports various types of maps, so it would be cool to have support for them. Or at least "leave room" to add this support later --- of course it's normal that at the beginning you would not implement all three TIsometricMap, TOrthogonalMap, THexagonalMap.

- All this would be very cool. I have quite a few games where I used (and always implemented "from scratch") some kind of 2D map --- 
  1. there's castle_game_engine/examples/isometric_game/ , 
  2. there's "Hydra Battles" with isometric game ( http://michaliskambi.itch.io/hydra-battles , complete source code and data on https://github.com/castle-engine/hydra-battles ), 
  3. there's "Orcs and Volcanoes" (not open-source unfortunately, because the data files are not mine, and I cannot publish them), screens on the sidebar of this post: http://castle-engine.sourceforge.net/news.php?id=2013-09-15 . This uses 2D map, but (for some tricks) it even rotates it in 3D, so it would probably needs it's own map implementation anyway (where your TTileMap would be converted to X3D nodes graph, and so rendered using our TCastleScene).

  This is just to show you what I'm thinking about. In an ideal world, your work would be useful for *all* these types of games! Which would be absolutely cool!

- I don't know how my proposals fit with the concept of "Rectangle, ellipse, polygon and image objects can be placed with pixel precision." mentioned on http://www.mapeditor.org/ . I don't know how these objects are represented (and how are they supposed to be drawn). Are they on top of isometric/orthogonal/hexagonal map? In this case they could be drawn inside TAbstractMap, and each TIsometricMap, TOrthogonalMap, THexagonalMap could use it by calling "inherited" from it's specialized "Render" method.

All of the above are just "general guidelines", my thoughts how I would approach it, after some thinking.

Small code stuff:

- Instead of using "record", it's often useful to use old-style Pascal objects, using the "object" keyword. You can then add procedures/functions to them, which is sometimes comfortable.

- Hm, I would probably changed TTileset and TImage to be just normal, new, classes, using the "class" keyword. Using "record" (or old-style Pascal "object") is justified only for performance reasons/memory layout. If you're going to use pointers to them (like PImage / PTileset) then *usually* it's better to just make them normal classes. A list of them is TFPGObjectList.

- Avoid reintroducing constructor of component. So instead of

  ~~~~
constructor Create(AOwner: TComponent; AURL: string); reintroduce;
~~~~

  better override normal constructor (without AURL param), and provide a property to set URL (or just a method like Load(AURL: string)). This is admittedly more difficult (preparing your class to be able to load many times), but usually this API is just better in the long run. In the first implementation, it's OK to say "you can only use the Load method once, and it must be called immediately after constructing the class" --- this is acceptable, and it leaves room to make it more flexible later.

tsr8:
What is the reason to make 3 classes like TIsometricMap, TOrthogonalMap, THexagonalMap?
Map type is defined in TMX file so programmer don't have to create for instance TIsometricMap explicitly. 

I think that one class like TTiledMap based on TUIControl will do rendering depends of TMX content. Eventually make another class to load a TMX file. It's not so hard for me to change it with your suggestions but it must be think over :)

>I don't know how my proposals fit with the concept of "Rectangle, ellipse, polygon and image objects can be placed with pixel precision." mentioned on http://www.mapeditor.org/ . I don't know how these objects are represented...

It is just list of points or a rectangle bounds (for ellipse). It shouldn't be difficult to implement. Drawing this should be optional because such object can be used to only mark something on map. I keep this at the end of work.

michaliskambi:
> What is the reason to make 3 classes like TIsometricMap, TOrthogonalMap, THexagonalMap? Map type is defined in TMX file so programmer don't have to create for instance TIsometricMap explicitly.

You're right! Since the format chooses the rendering mode (and also, maybe it can change at runtime --- at least in Tiled, I see that you can change map "orientation" at any point), it would be uncomfortable to split it into 3 separate classes.

> I think that one class like TTiledMap based on TUIControl will do rendering depends of TMX content. Eventually make another class to load a TMX file. It's not so hard for me to change it with your suggestions but it must be think over :)

I think that splitting into one rendering/UI class like TCastleTiledMapControl (descends from TUIControl) and one class that holds the information like TCastleTiledMap (descends from TComponent) may still be nice. 

- Consider the example with "Orcs and Volcanoes" http://castle-engine.sourceforge.net/news.php?id=2013-09-15 where I could happily use the information contained in TCastleTiledMap, but I don't want it's 2D rendering, I want specialized 3D rendering.

- Also, I imagine that TCastleTiledMap can eventually grow with some useful methods for processing maps. Like in https://github.com/castle-engine/hydra-battles/blob/master/code/gameabstractmap.pas I implemented utilities like Neighbors , HandleNeighbors, PositionToTile... All these could be implemented in a generic way (handling all 3 map orientations) in TTiledMap. So it will be beneficial to separate the "information processing" class TCastleTiledMap from "display/input" class TCastleTiledMapControl.

Anyway, treat it only as a guideline ("how I would do it") at this point. The decision whether to do the split may be more obvious later in the implementation process:)

> Pointers are necessary to make specialized lists (can't use records directly - compiler errors).

If you want to have a list of records, use our own TGenericStructList. It can be used quite like TFPGList (in fact, it was based on FGL sources), but it handles records without any problems:) This also gives you performance --- all the consecutive records will be in one memory stream.

michaliskambi:
BTW, I was playing with Tiled map editor yesterday, and I'm very impressed. It would be cool to have it supported by the engine out-of-the-box. So, in case it wasn't obvious, I am excited about your work:)

If you're looking at which library to use:
- for JSON, we already use FpJson, JSONParser in X3DLoadInternalSpine unit. That's for reading Spine 2D animations, http://esotericsoftware.com/spine-json-format .
- for gzip and base64, FPC includes ready decoding classes. We already use TGZFileStream and TBase64DecodingStream , mostly in units in src/net/ .

tsr8:
There are many many todos yet but for now we have a milestone! Simple orthogonal map rendering works as you can see in demo application.

BTW Is there any methods to draw primitives like: line, circle, rectangle etc.?

michaliskambi:
> There are many many todos yet but for now we have a milestone! Simple orthogonal map rendering works as you can see in demo application.

Excellent news -- I'll try it out as soon as I'll catch a little free time!:)

> BTW Is there any methods to draw primitives like: line, circle, rectangle etc.?

Only for rectangles now: DrawRectangle, from CastleGLUtils unit.

Lines, circles have to be drawn by adding a T2DScene with TLineSetNode or TIndexedLineSetNode to the T2DSceneManager. This may be an overkill for some purposes, I know --- it's much more complicated than a simple DrawXxx method would be. If needed, I can easily add to CastleGLUtils routines DrawLines, DrawTriangles. And then DrawCircle, using DrawLines or DrawTriangles (filled or not filled) would be done using them. Do you need them?

tsr8:
> Do you need them?
It will be also useful for debugging in games. There is such feature in Tiled called object layer - I only need lines for drawing that layer. I will study T2DScene more.

Kagamma:
Hi,

I noticed currently tmx reader only read files directly from hard drive so I created a patch to make it work with other protocols supported by CGE. [castletiledmap.zip](https://github.com/castle-engine/castle-engine/files/415154/castletiledmap.zip)

michaliskambi:
>  If needed, I can easily add to CastleGLUtils routines DrawLines, DrawTriangles. And then DrawCircle, using DrawLines or DrawTriangles (filled or not filled) would be done using them. Do you need them?

I have added a procedure DrawPrimitive2D to CastleGLUtils unit that can be used to draw points, lines and triangles in 2D. More complex stuff can of course be constructed from these primitives. Current DrawRectangle is a simple call to DrawPrimitive2D to draw 2 triangles.

I did not test them extensively, so sorry in advance for any bugs:) I'll be on a small vacation for a few days, but when I get back I'll test and improve them more.

Regards!

michaliskambi:
I improved the DrawPrimitive2D a little, and added DrawRectangleOutline, DrawCircle, DrawCircleOutline to the CastleGLUtils. This should make drawing simple 2D shapes easier than having to create T2DScene and T2DSceneManager instances.

I also added a control to wrap them: TCastleShape, this is somewhat equivalent to Lazarus LCL TShape control.

It's demonstrated by castle-engine/examples/2d/controls_demo.lpr . Hope this is useful:)

src/ui/castletiledmap.pas
```diff
@@ -791,7 +794,7 @@ procedure TTiledMap.LoadTMXFile(AURL: string);
 begin
   Doc := nil;
   try
-    ReadXMLFile(Doc, URIDeleteProtocol(AURL));  //todo: what is proper way to open files? (AbsoluteURI ?)
+    ReadXMLFile(Doc, Download(AURL));
```
michaliskambi: @Kagamma , @tsr8 : Note that this line makes a memory leak, as "Download(AURL)" returns a new stream instance that is nowhere freed.

The easiest solution is to use "URLReadXML(Doc, AURL)". It is defined in CastleXMLUtils, and will free the stream as necessary. It's implementation (copied from CastleXMLUtils sources) is below:

~~~~
procedure URLReadXML(out Doc: TXMLDocument; const URL: String);
var
  Stream: TStream;
begin
  Doc := nil; // clean "out" param at start, just like ReadXMLFile
  Stream := Download(URL, []);
  try
    ReadXMLFile(Doc, Stream);
  finally FreeAndNil(Stream) end;
end;
~~~~

tsr8:
Tiled files loader is pretty complete now. I suggest to merge the pull request because it is quite hard to resolve new conflicts with the main branch.
Visual component and example project need to be developed more (in new pull requests).

michaliskambi:
Great! I'll review everything ASAP. Give me a couple of days, I made for myself a little backlog in November (including forgetting to announce some recent cool stuff in the engine:)

michaliskambi:
Hi! Sorry for the delay, I still didn't have time to review your work... I'm working through my TODO list, and should hopefully handle this soon. In the meantime, feel free to push more stuff to your branch.

The good news is that most of my TODO list is about doing various cool things in the Castle Game Engine, for the upcoming 6.0 release:) So, although I'm late with my tasks, but Castle Game Engine keeps getting better and better:) Thanks for understanding!

michaliskambi:
Merged! Sorry again for the large delay.

Great and large work, thank you! It was a joy to view a couple of test maps:)

I will be making annoucements on Facebook / Google+ about a couple of new engine features in the next 2 weeks. I'll make a post about your Tiled work of course:) One more feature for a gigantic 6.0 release, that should hopefully happen in 2-3 weeks!:)

I made some post-merge changes in the commits:

- https://github.com/castle-engine/castle-engine/commit/38e9a6247cc562a6e9e01b84c58768789e60280f ,
- https://github.com/castle-engine/castle-engine/commit/79274a6e61c93dc4ba2c1af22cc484433df5a4f2 ,
- https://github.com/castle-engine/castle-engine/commit/ec57dfed0cd03f251097c5984253b03cc487554a ,
- https://github.com/castle-engine/castle-engine/commit/6b662678bda5f3c119437e053e1085d732a2d812

Various notes:

- I see you have a habit of naming local variables with underscore, like "_xxx". While I perfectly understant the usefulness of this notation, it is inconsistent with the rest of the engine code, so it looked a little odd. We don't use such notation, instead we name local variables like "Xxx" (CamelCase, without underscores), just like most other identifiers.

- Compile with "notes" (-vn) to capture and remove unused variables. Notes sometimes show some useful things, so we want to keep them "clean":)

- Do not use tabs, only spaces:)

- I fixed some memory leaks. To easily capture memory leaks in development, I advice compiling code with "-gv" in development. This will give you a report "Heap dump by heaptrc unit" at the end of program's execution. I even keep these lines in my ~/.fpc.cfg :

~~~~
#IFDEF DEBUG
-gh
-gl
#ENDIF
~~~~

- Use Parameters from CastleParameters unit, instead of directly using ParamStr and ParamCount. This allows other things to handle partial parameters and remove them, e.g. Window.ParseParameters can handle parameters like http://castle-engine.sourceforge.net/opengl_options.php .

- Consider changing most types in CastleTiledMap to be "classes", not "records"? And then use TFPGObjectList for lists, not TGenericStructList.

  Reason: Plugging memory leaks is easier when you have a guaranteed constructor/destructor calls. Current combination of records and classes caused some memory leaks.

  In general, "records" (and old-style "objects") are cool if you really need compactness in the memory. But in many cases, using modern "classes" is more comfortable, and the speed/memory difference is not noticeable.

  Right now, I did not change much -- most structures are stil "records". Well, I changed some of them to be old-style "objects", which is pretty much like a record with methods.

I saw you also did some changes to sprite:) Most of them I generally like (including new names for HorizontalSpacing and VerticalSpacing). But:

- I actually need VerticalSpacing to be a spacing at the *top*, not *bottom*, of the frame. Example spritesheet needing this: http://opengameart.org/content/cowboy . So I added a property VerticalSpacingBottom to make this configurable. By default it is false, but you can change it to true.

  (The reasoning for VerticalSpacingBottom=false by default is *not* that it matches my usecase. It's because in the engine, the "left and bottom" are usually matched together, vs "right and top". E.g. in 2D, "left and bottom" is (0, 0). So, if HorizontalSpacing is at "right", than VerticalSpacing should by default be "top".)

  I understand that for Tiled sprites, we want to set VerticalSpacingBottom=true, so I did it. So nothing should be broken:)

- I removed "PSprite = ^TSprite;". Having a pointer to a class instance is usually not necessary, and I see it's not used now.

- I removed the "//todo: new overloaded constructor with params like: framewidth, frameheight, margin, spacing" -- I don't think we should have such constructor. Too many overloaded constructors make the calls confusing, as it's not immediately obvious what each parameter means. All the properties can be set after constructing.

Again, thank you for the incredible work!

michaliskambi:
I did one more important change in commit https://github.com/castle-engine/castle-engine/commit/d7fb54252f9e97ef968e56e96e387416559db83d . I decided to keep both the non-UI class and the UI component inside the same unit, CastleTiledMap, which is also now moved to ui/opengl/ .

The reasons:
- The CastleControls unit was never supposed to be a bag for all UI controls in the CGE:) It became a bag for "small and/or often used UI controls", but let's at least minimize the things inside. So it's good to move tiled map outside of it.
- The "tiled" maps are independent from the rest of the UI stuff, so they fit nicely in a completely own unit. Same reasoning applied e.g. to CastleOnScreenMenu.
- The downside is that now both the OpenGL-dependent and non-OpenGL-dependent parts are inside a single unit. But I think that's acceptable (and it's probably not worth splitting into two units for now).

I'm still convinced that the separation of classes (TCastleTiledMapControl vs TTiledMap) is very important and useful (allows to visualize the map in any way, or use it without the visualization). But separating them into two units does not seem so important.

I also committed https://github.com/castle-engine/castle-engine/commit/5fcb5f2d1973846598267a4589969f07cda723e2 to keep enum names xxYyy, following the typical Pascal conventions:)

That's probably the end of my tweaking for some time. I'm very happy with the API now! Please, please make a hexagonal map implementation:)

P.S. Running 

./tiled_demo data/sewer_automap/rule_008.tmx

fails with 

> Exception: Exception "ERangeError" : 
> Range check error
> 
>   $000000000093B059  RENDERORTHOGONAL,  line 88 of castletiledmap_control.inc
>   $000000000093BB92  RENDER,  line 211 of castletiledmap_control.inc
> ...

tsr8:
> P.S. Running
> 
> ./tiled_demo data/sewer_automap/rule_008.tmx
> 
> fails with
> 
>     Exception: Exception "ERangeError" :
>     Range check error
> 
>     $000000000093B059 RENDERORTHOGONAL, line 88 of castletiledmap_control.inc
>     $000000000093BB92 RENDER, line 211 of castletiledmap_control.inc
>     ...

Sorry, I can't reproduce this. It works fine on my system.

michaliskambi:
> Sorry, I can't reproduce this. It works fine on my system

I forgot to mention: you need to compile in DEBUG mode to see range check errors. Like this:

~~~~
$ castle-engine compile --mode=debug
Compiling project "tiled_demo" for OS / CPU "linux / x86_64" in mode "debug".
...
$ ./tiled_demo data/sewer_automap/rule_008.tmx
Log for "tiled_demo". Started on 2017-01-22 at 10:32:11.
...
LoadTlesetsImages: columns:8
LoadTlesetsImages: columns:3
filepath: data/sewer_automap/rule_008.tmx
Exception: Exception "ERangeError" : 
Range check error

  $000000000093B239  RENDERORTHOGONAL,  line 88 of ../../src/ui/opengl/castletiledmap_control.inc
  $000000000093BD72  RENDER,  line 211 of ../../src/ui/opengl/castletiledmap_control.inc
  $000000000055D53E  RENDERCONTROLCORE,  line 139 of ../../src/ui/opengl/castleglcontainer.pas
  $000000000055D859  RENDEREVERYTHING,  line 171 of ../../src/ui/opengl/castleglcontainer.pas
  $000000000055D73D  EVENTRENDER,  line 210 of ../../src/ui/opengl/castleglcontainer.pas
  $000000000044001C  DORENDER,  line 3396 of ../../src/window/castlewindow.pas
  $0000000000444D51  UPDATEANDRENDEREVERYTHING,  line 4745 of ../../src/window/castlewindow.pas
  $0000000000444DED  UPDATEANDRENDEREVERYTHING,  line 4758 of ../../src/window/castlewindow.pas
  $000000000043DAD9  PROCESSUPDATES,  line 1575 of ../../src/window/gtk/castlewindow_gtk.inc
  $000000000043DE04  PROCESSMESSAGE,  line 1694 of ../../src/window/gtk/castlewindow_gtk.inc
  $0000000000445A5E  DORUN,  line 4954 of ../../src/window/castlewindow.pas
  $00000000004AC1E9
  $00000000004415CD  OPENANDRUN,  line 3717 of ../../src/window/castlewindow.pas
  $000000000040B97B  main,  line 6 of tiled_demo_standalone.lpr
~~~~

By default, castle-engine compiles in release mode, which is faster but doesn't check some things. Which means that you can look at an uninitialized memory area, and sometimes it will return zero, sometimes crash...

tsr8:
Can't reproduce still.

Do you have ERangeError all the time or randomly?
Does it happen on another tmx files too?

michaliskambi:
Can you clean the engine before compilation ("make clean" in top engine directory), to make sure that the engine units are recompiled in debug mode (with range checking)? I should have mentioned it before, sorry.

The error happens reliably, 100% of the time, on affected TMX files. It happens with most (but not all) TMX files inside the ./tiled_demo data/sewer_automap/ directory. I checked that:

~~~~
$  cd .../castle-engine
$ make clean
...
$ cd examples/tiled/
$ castle-engine compile --mode=debug
Compiling project "tiled_demo" for OS / CPU "linux / x86_64" in mode "debug".
FPC version: 3.0.0
...
$ find -iname '*.tmx' -exec ./tiled_demo '{}' ';'
...
filepath: ./data/sewer_automap/sewers.tmx
Exception: Exception "ERangeError" :
...
~~~~

Range check error occured on these files:

- ./data/sewer_automap/sewers.tmx
- ./data/sewer_automap/rule_001.tmx
- ./data/sewer_automap/rule_003.tmx
- ./data/sewer_automap/rule_004.tmx
- ./data/sewer_automap/rule_008.tmx
- ./data/sewer_automap/rule_009.tmx

All other TMX files inside data/ are OK, e.g. sewer_automap/rule_002.tmx is OK.

All of the errors are on the same line, "line 88 of ../../src/ui/opengl/castletiledmap_control.inc", as far as I see.

I was checking with tiled_demo_standalone.lpr modified to set "Application.StopOnException := true;", to crash after first exception, made it easier to test all TMX files in one go. Other than that, I'm using the latest engine version from GitHub, unmodified.

I test on Linux, 64-bit (x86_64). Tested with FPC 3.1.1 (recent FPC trunk), FPC 3.0.0 and FPC 2.6.4. If you still cannot reproduce, tell me your OS / architecture, I'll see if I can test on a similar system:) Although range check errors should occur regardless of OS / architecture.

PR #15, opened 2016-03-25
More usage of castleconf.inc

pascaldragon:
When checking out some of the FPC bugs you worked around for I also noticed that compiling the engine from the commandline (without using fpmake or Lazarus) requires the mode and the string type to be set. Because of this I decided to add the necessary switches to castleconf.inc and add this include to every unit that doesn't have this yet. This way the units are always compiled the same no matter what default settings are used.

### DIFF
--- a/src/3d/castle3d.pas
+++ b/src/3d/castle3d.pas
@@ -16,6 +16,8 @@
 { Base 3D objects (T3D, T3DList, T3DTransform, T3DOrient, T3DMoving). }
 unit Castle3D;
 
+{$I castleconf.inc}
+
 interface
 
 uses SysUtils, Classes, Math, CastleVectors, CastleFrustum,
--- a/src/3d/castleboxes.pas
+++ b/src/3d/castleboxes.pas
@@ -16,6 +16,8 @@
 { Axis-aligned 3D boxes (TBox3D). }
 unit CastleBoxes;
 
+{$I castleconf.inc}
+
 interface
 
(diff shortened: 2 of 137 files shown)


michaliskambi:
I deliberately didn't add castleconf.inc everywhere. I also deliberately didn't add there some directives like $mode or $H. This is mentioned at the beginning of src/base/castleconf.inc.

Instead my idea was that you should always compile the engine using one of the supported methods:

- if you like bare-bone fpc calls, then add @castle-fpc.cfg on the command-line, which has the power to specify mode/string switches and more.
- or use castle-engine build tool,
- or use Lazarus (possibly through lazbuild),
- or use fpmake (that underneath passes @castle-fpc.cfg to FPC).

My motivation for this was that:

1. You cannot specify everything using $xxx options. E.g. optimization options, which can change the speed quite a lot. So it seemed to me that for a long-term engine usage, you should use one of the "supported compilation methods" anyway.

2. Adding castle-conf.inc everywhere is tiresome, and error-prone. The existence of this will need to be checked by Jenkins scripts, otherwise I just *know* that at some point, I'll accidentally introduce a unit and forget to add there "{$I castleconf.inc}":)

The main advantage of using castleconf.inc is that it "just works", and you don't need to know about all these "supported compilation methods" I listed above. You just do "fpc myunit.pas", add -Fu and -Fi as necessary, and it works. Admittedly, this may be a big advantage, if it allows more people to jump into using the engine easily.

Therefore, applied:) Thank you!

PR #16, opened 2016-03-25
Big endian fixes

pascaldragon:
These changes fix some problems when using the engine on big endian systems.
First of there was a bug in TSimplePeekCharStream that affected the X3D reader and secondly at least the 3DS loader misbehaved. For this I've added a TStream helper that deals with reading (and writing) values in the correct endianess and this helper should be used in the future for reading (and writing) binary data. I've already corrected the 3DS loader and I'll adjust more as I get along with testing everything on my big endian machines (one PowerPC and one M68k), though I don't know how long that will take, so merging this already would be good.

### DIFF
--- a/fpmake.pp
+++ b/fpmake.pp
@@ -131,6 +131,7 @@ begin
     P.Targets.AddUnit('castleprogress.pas');
     P.Targets.AddUnit('castlerectangles.pas');
     P.Targets.AddUnit('castlestringutils.pas');
+    P.Targets.AddUnit('castlestreamutils.pas');
     P.Targets.AddUnit('castletimeutils.pas');
     P.Targets.AddUnit('castleunicode.pas');
     P.Targets.AddUnit('castleutils.pas');
--- a/packages/castle_base.lpk
+++ b/packages/castle_base.lpk
@@ -36,7 +36,7 @@
     <License Value="GNU GPL >= 2
 (or LGPL >= 2, if compiled with CASTLE_ENGINE_LGPL)"/>
     <Version Major="5" Minor="2"/>
-    <Files Count="349">
+    <Files Count="350">
       <Item1>
         <Filename Value="../src/3d/castle3d.pas"/>
         <UnitName Value="Castle3D"/>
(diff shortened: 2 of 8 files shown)


michaliskambi:
In general --- a very nice work, I very much want to apply it. This will make existing code reading binary data *much* nicer. And kudos for catching that bug in TSimplePeekCharStream.PeekChar , this was a nasty case where I sneakily depended on endianess. I stared in amazement at my own mistake there:)

There are two issues that bother me (both have the same solution though):

1. Inside CastleStreamUtils methods, you do a single Read(), and assume that reading fails if you cannot read the whole data in one go.

   This will work in most cases, and it will always work for memory streams, but it's not guaranteed to work in general (even on TFileStream). Consider a TCP socket, Read() may read only as much as is available, and e.g. return only 10 leading bytes from the larger structure (like TVector4Double that's 32 bytes long). Your code will treat it as a failure, while in fact it should just try reading more, to get the remaining data.

   The simplest way to fix would be to use ReadBuffer instead of Read, that simply always tries "as hard as it can" to read the necessary number of bytes. It only gives up if the stream ended.

   See http://bugs.freepascal.org/view.php?id=24645 and links.

2. You use the ReadLE (and similar methods) without ever checking the result, as far as I see. This is not OK, the engine deliberately uses ReadBuffer for reading in most places, and depends that unexpected file end will just result in an exception. In your code, the unexpected file end will be silently ignored, and some data will remain undefined (unset by the ReadXX() call).

   The solution to this seems the same as for previous point, luckily:) Simply use ReadBuffer inside.

So, I would propose to redo all ReadXX and WriteXX methods in TStreamHelper to be procedures, using ReadBuffer underneath. E.g.

~~~~
procedure TStreamHelper.ReadLE(out Value: Word);
begin
  ReadBuffer(Value, SizeOf(Value));
  Value := LEtoN(Value);
end;
~~~~

What do you think?

Thank you!

pascaldragon:
Sounds reasonable, especially this way one doesn't need to add error checking code everywhere :)
Thus I made the chances you suggested (I also adjusted the WriteLE()/WriteBE() methods similary for consistency)

michaliskambi:
Excellent! Thank you very much, merged!:)

PR #17, opened 2016-03-28
Implemented dmMultiply blending mode

eugeneloza:
dmMultiply blending mode for TRGBAlphaImage. This is simple implementation of 'multiply' filter like in GIMP which I needed to add some diversity to my game UI leaving it scalable. It seems that it should work with any destination/source image type, but requires some testing to see if everything is working as expected. I've added the dmMultiply everywhere where it was applicable (and if I was sure what I am doing :)), but I've tested it only for TRGBAlphaImage.

Hmmm... it seems that I create a pull request for the whole branch, not for different changes... I've also updated/extended/written some descriptions for TBackgroundNode, TSwitchNode and TInputMotion. So they are in this pull too...


michaliskambi:
First of all, sorry for such long reply:) 

* I see a small (possible) issue with dmMultiply implementation. Can you check (and eventually fix it)?

* As for the descriptions --- I have various small suggestions. There is a lot of stuff that is a matter of taste, undoubtedly. I just got accustomed to how I wrote my descriptions. So discussion is most welcome of course:) 

As for the dmMultiply:

1. I'm not sure is the code correct with large values. You multiply Byte by Byte, which is done in the Byte (0..255) range. Consider this expression:

  ~~~~
  PDest^[0]*PSource^[0] div 255
  ~~~~

  If PDest^[0] = 123 and PSource^[0] = 255, it should calculate 123. But the partial value, (123*255), does not fit in the Byte range. So I suspect that right now, it will cause an overflow. When compiling the engine with overflow checking, you may get an error. When compiled without overflow checking, the result will be just truncated to the least-significant 8 bytes, so it may be incorrect.

  When you use such small integer types, you should be aware that the calculations are also done in this range, the type is not automatically "upgraded" to contain all possible values.

  So I would check to make sure it really works, e.g. use specific test images or write a simple program to check that

  ~~~~
  B1 * B2 div 255
  ~~~~

  always works Ok. If I set B2=255 and iterate over all possible B1 values (0..255), is the result correct (should be equal B1)?

  I suspect it is not always correct, and you should write

  ~~~~
  Word(PDest^[0])*PSource^[0] div 255
  ~~~~

  This will multiply values as Words (16-bit unsigned values), which can calculate and hold correct result.

2. Probably more places need to be filled for dmMultiply, to make it work on grayscale images too:)

3. We have automatic tests of some Draw usage in the unit tests/testimagesdraw.pas . Writing these tests is painful, but it pays off, guaranteeing that the result is always what we want:) You may consider adding some tests there, testing that some simple inputs get what they should. Testing the "overflow" issue mentioned above could also be done as one of these tests.

As for the descriptions, some conventions:

1. Do not start the description with the name of the identifier, and generally keep the 1st description sentence terse and precise. The reason is that PasDoc uses the first sentence as an automatic abstract, and it looks best when it's short and up to the point. So instead of

  ~~~~
  { TInputMotion represents a current mouse or a finger touch state.
  ~~~~

  it's better to say just

  ~~~~
  { Represents a current mouse or a finger touch state.
  ~~~~

  Even better is to remove the verb (it's a type --- saying it "represents something" is a little redundant) and just say this:

  ~~~~
  { Current mouse or a finger touch state.
  ~~~~

2. Place 1 space at the beginning and end of { } comment, so prefer

  ~~~~
  { In case mouse buttons were pressed, they can be found in this variable. }
  ~~~~

  instead of

  ~~~~
  {In case mouse buttons were pressed, they can be found in this variable.}
  ~~~~

  Also, saying "they can be found in this variable" feels redundant. This is a description of a variable, so it's implicit that whatever we write, applies to this variable. I would suggest just this:

  ~~~~
  { Which mouse buttons are pressed at the time of the event. }
  ~~~~

3. Usually try to wrap the lines, if possible, so that they are no longer than 80 characters. 

4. Indent the comments (all lines) same as the identifiers. E.g. your TBackgroundNode comment should be indented.

  Summarizing this, and adding more notes, for TBackgroundNode, I would suggest this:

  ~~~~
    { A skybox of the scene. Visible only if this scene is SceneManager's
      main scene. The skybox is defined by adding items (by @code(Items.Add))
      for each @code(FdXxxUrl) field.

      In @italic(Y-up orientation) they will correspond to back, bottom, front,
      left, right and top sides of the skybox.

      In case of @italic(Z-up and Y-forward) they will be:
      @unorderedList(
        @itemSpacing compact
        @item Back -> top rotated 180 deg.
        @item Bottom -> back rotated 180 deg.
        @item Front -> bottom
        @item Left -> left rotated 90 deg. CCW
        @item Right -> right  rotated 90 CW
        @item Top -> front
      )

      The X3D specification has detailed documentation about the Background node on
      http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/enveffects.html#Background .
      The Background orientation is shown here:
      http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/enveffects.html#f-BackgroundNode .
    }
    TBackgroundNode = class(TAbstractBackgroundNode)
  ~~~~

  As you see, I added indentation, and used some @code and @italic and @unorderedList to make it pretty:)

  I also added links to X3D documentation for this node. X3D specification has extensive documentation for all fields and nodes, linked from our own pages like http://castle-engine.sourceforge.net/x3d_implementation_environmentaleffects.php .

5. As for

  ~~~~
  {This property shouldn't be set manually. See TSwitchNodeHelper.WhichChoice.}
  ~~~~

  I would avoid documenting it here. This comment applies to *all* simple @code(FdXxx) properties of X3D nodes (integer, float, double, vector...). The preferred method is to get/set them using the respective @code(Xxx) properties defined by helpers. In the future, even the complex types (like multiple-value fields) should get their nicer wrappers.

6. Do not change src/x3d/x3dnodes_node_helpers.inc . It is automatically regenerated whenever a node/field is added/changed.

  If we want to place documentation there, then we'll have to
  
  - Extend the src/x3d/nodes_specification/x3d_nodes_spec_to_pascal/ tool, to read comments from text files like src/x3d/nodes_specification/components/EnvironmentalEffects.txt and remake them into Pascal comments when generating src/x3d/x3dnodes_node_helpers.inc .
  
  - Or, probably easier, use PasDoc feature to place documentation in external files: http://pasdoc.sipsolutions.net/ReadDescriptionFromFile .

eugeneloza:
> First of all, sorry for such long reply:)

No problem :) But I'll try to split my answer into sub-comments for easier navigation... I still learn to use GitHub in a correct way :)

> When compiling the engine with overflow checking, you may get an error.  When compiled without overflow checking, the result will be just truncated to the least-significant 8 bytes, so it may be incorrect.

I always try to compile everything with overflow and range checking. And yes, I was aware of the possible issue... I got no errors&warnings and for me it works flawlessly and does everything as expected (Lazarus 1.6.0 usually automatically converts all integer operations to int64/int32 as far as I know). But I think I should really try something like you did in AddBytesPremultiplied to be absolutely sure.
Actually I was making it this way from the very beginning, but got some strange “Error 21*******” and abandoned it for a simple in-line formula. However, now I understand it was due to turned-off range-checking (I was using the wrong values for PDest/PSource pointers components - the issue I solved later, but forgotten about making MultiplyBytes procedure back).
I'll fix this asap. Maybe even today evening.

eugeneloza:
I've fixed blending modes and will re-make the commit from the scratch.

PR #18, opened 2016-04-07
Joystick demo application

tsr8:
I've created a demo application to show how to use CastleJoysticks module.
There are also commits that fixing stuff on Windows platform.

Current status:
* Windows: works and tested (old multimedia joysticks API; task for future to implement xinput devices).
* Linux: works and tested with two devices working at same time (need upgrade to "evdev" interface).

These upgrade tasks will give us more functionality on new devices, for example forcefeedback.


michaliskambi:
Fantastic:) Thank you very much, merged. I'll borrow some joystick and test it on Linux soon:)

PR #19, opened 2016-04-08
Create .gitattributes

tsr8:
*.inc files are treated by GitHub as C++ files so the file change this to Pascal. It's only for statistics.


michaliskambi:
Merged, thank you! It took me a little time to find where the stats are displayed by GitHub:)

PR #22, opened 2016-06-30
Re-implementation of dmBlendSmart and dmMultiply

eugeneloza:
Hi, Michalis!

I've decided to make the pull request from the scratch because I'm not so confident in my understanding of git and I'm afraid to mess something up :) Yesterday I was struggling 4 hours just to finally "discard my fork" and re-fork it again :)

Finally I've re-implemented and extended dmBlendSmart and dmMultiply modes for all applicable image types. I've also extended dmBlendSmart=dmBlend for no-alpha destination as it produces expected results.
Practically there is no sense in dmBlendSmart for no-alpha source, as it'll just replace the original image part and it's better to have an exception here because in this case the user is doing something that'll result in an unexpected result. However, it still might be extended in this sense to no-alpha / no-alpha blending.
I've also changed dmMultiply logic for drawing alpha images over non-alpha images to account for transparency because basic dmMultiply algorithm provided mathematically correct but visually ugly result.
Plus I've made an additional test utility for testing blending modes and images combination - requires manual blend mode specification in the code. I'm not sure if it's good for anything, but I'll attach it here.
However, I saw there are separate tests for CGE in castle-engine\tests. I think I should also make a test there similar to other tests in testcastleimagesdraw.pas. Looks relatively complex and will take me some time to understand the logic and the purpose of the test.

There seems to be a minor typo in castleimages_draw.inc in procedure TRGBAlphaImage.DrawFromCore > procedure SourceRGB
	else raise EImageDrawError.Create('Blend mode not implemented (TRGBAlphaImage.DrawFromCore)');
should be
	else raise EImageDrawError.CreateFmt(SErrBlendModeNotImplemented, [Source.ClassName, ClassName]);
as in all other sources/destination combinations, right? (I've fixed it in the commit)

Practically all 16 sourceXXXXX procedures and each blend mode are so similar... But I've been struggling for 4 hours and found no way to unify them without dropping efficiency and merging 2 or 3 of those won't give any result worth the effort... However, maybe at least sorting them would be a good idea? E.g. to make strictly that RGBAlpha comes first, then GrayscaleAlpha (as it also uses alpha-logic), then non-transparent RGB and Grayscale - both in DrawFromCore order and sourceXXXXX order. No serious profit, but might provide a little better code readability.
This doesn't seem to be difficult and I can hardly mess up something here, so I can do it, if you're ok with the order. Or maybe other ideas?
[DrawModeTest.zip](https://github.com/castle-engine/castle-engine/files/340839/DrawModeTest.zip)

### DIFF
--- a/src/images/castleimages.pas
+++ b/src/images/castleimages.pas
@@ -246,11 +246,16 @@ destination.alpha := destination.alpha; // never changed by this drawing mode
     dmBlend,
 
     { An advanced blending mode capable of blending 2 images with alpha channel.
-      Based on https://en.wikipedia.org/wiki/Alpha_compositing formula for alpha-blending.
-      This one is much less efficient than dmBlend and should be used only in case
-      several layers of semi-transparent images should overlay one another and it
-      matters to accurately account for both images alpha channel. }
+    Based on https://en.wikipedia.org/wiki/Alpha_compositing formula for alpha-blending.
+    This one is much less efficient than dmBlend and should be used only in case
+    several layers of semi-transparent images should overlay one another and it
+    matters to accurately account for both images alpha channel. Implemented for
+    all @link(TRGBAlphaImage) and @link(TGrayscaleAlphaImage) combinations.
+    }
     dmBlendSmart,
+    { Simple implementation of 'multiply' filter. I.e. bytes of one image are just
+    multiplied by bytes of another image.}
+    dmMultiply,
 
     { Additive drawing mode, where the image contents of source image
       are added to the existing destination image. That is,
--- a/src/images/castleimages_draw.inc
+++ b/src/images/castleimages_draw.inc
@@ -136,6 +136,25 @@ begin
   Result := W;
 end;
 
+function MultiplyBytes(const Dest, Source: Byte): Byte; {$ifdef SUPPORTS_INLINE} inline; {$endif}
+var
+  W: Word;
+begin
+  W := Dest * Word(Source) div 255;
+  if W > 255 then W := 255;
+  Result := W;
+end;
+
+function MultiplyBytesOpaque(const Dest, Source, Opacity: Byte): Byte; {$ifdef SUPPORTS_INLINE} inline; {$endif}
+var
+  W: Word;
+begin
+  W := Dest * (255 - Opacity) div 255 + Dest * Word(Source) div 255 * Opacity div 255;
+  if W > 255 then W := 255;
+  Result := W;
+end;
+
+
 { Drawing when TRGBAlphaImage is a destination ------------------------------- }
 
 procedure TRGBAlphaImage.DrawFromCore(Source: TCastleImage;
@@ -161,6 +180,22 @@ procedure TRGBAlphaImage.DrawFromCore(Source: TCastleImage;
             Inc(PDest);
           end;
         end;
+      dmMultiply:
+        begin
+          for DestY := Y to Y + SourceHeight - 1 do
+          begin
+            PSource := Source.PixelPtr(SourceX, SourceY + DestY - Y);
+            PDest := PixelPtr(X, DestY);
+            for DestX := X to X + SourceWidth - 1 do
+            begin
+              PDest^[0] := MultiplyBytes(PDest^[0],PSource^[0]);
+              PDest^[1] := MultiplyBytes(PDest^[1],PSource^[1]);
+              PDest^[2] := MultiplyBytes(PDest^[2],PSource^[2]);
+              Inc(PSource);
+              Inc(PDest);
+            end;
+          end;
+        end;
       dmAdd:
         for DestY := Y to Y + SourceHeight - 1 do
         begin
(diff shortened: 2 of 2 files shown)


michaliskambi:
Thank you very much, the changes are applied!

I only did formatting fixes (and I fixed the MultiplyBytesOpaque to have one more typecast to Word, to multiply "Dest * (255 - Opacity)" as Words). I did them in a separate commit, https://github.com/castle-engine/castle-engine/commit/6c3c24267aea01e950f41b5ea1b17c97d22be981 . You should be able to merge now the castle-engine:master version into your fork eugeneloza:master and thus have everything up-to-date. You should be able then to push further commits to eugeneloza:master, and make more pull requests:)

> Practically there is no sense in dmBlendSmart for no-alpha source, as it'll just replace the original image part and it's better to have an exception here because in this case the user is doing something that'll result in an unexpected result.

I like to have all operations on RGB images behaving the same way as on RGBA images (where all pixels have alpha = maximum, i.e. fully opaque). This makes things simpler for engine user, as you don't have to worry "Which image class to load to - TRGBImage or TRGBAlphaImage?". You can just use LoadImage(URL), and let it choose the "best" image class.

This also allows flexibility in LoadImage implementation:

1. For some formats, as the "best" it chooses the format implied by the file header information.

2. In case of FpImage-based image loaders (which include some very common formats, like PNG and JPG), it analyses the pixel content to decide on the "best" memory format. If you create an image in GIMP that is pure white, but with RGB mode, add then add alpha channel (but leave it opaque), then the PNG image is stored with the RGBA format, but it will be loaded to a TGrayscaleImage in our engine. 

  At the beginning, I thought of this as a disadvantage, but it wasn't possible to do anything better - FpImage didn't provide enough information for us to know the original PNG file header information.

  Now, I actually think that this is a *good thing*, as it saves GPU memory of textures, and it's a precious thing on Android. (The latter reason is so good, that it may be at some point done for *all* image loaders).

To make this "somewhat invisible" to the engine user, we need to:

- We need to have drawing methods working on all 4x4 combinations of images (when the source and destination is grayscale, grayscale+alpha, RGB, RGBA).

- Things should be designed to treat "no alpha" always the same as "there is alpha channel, and it's opaque". We have to make the equations and code follow this principle, the computations for RGB / grayscale images must yield the same results as for RGBA / grayscale+alpha with alpha=255.

In other words, you can say that "using any format besides RGBA is only an optimization of memory use". The visible effects of drawing a grayscale opaque image should be the same as using RGBA image with red = green = blue, and alpha = 255.

(And yes, I should probably document this fact much more clearly somewhere in the CastleImages unit:)

>  I think I should also make a test there similar to other tests in testcastleimagesdraw.pas. Looks relatively complex and will take me some time to understand the logic and the purpose of the test.

Tests inside testcastleimagesdraw.pas grew into something complex, indeed:) But the idea is simple: just create a method that does *anything*, and raises an exception if something yields unexpected results. It's OK to add new tests to TestCastleImagesDraw, also.

> should be
> else raise EImageDrawError.CreateFmt(SErrBlendModeNotImplemented, [Source.ClassName, ClassName]);
> as in all other sources/destination combinations, right? (I've fixed it in the commit)

Yes, thanks!

> Practically all 16 sourceXXXXX procedures and each blend mode are so similar... But I've been struggling for 4 hours and found no way to unify them without dropping efficiency and merging 2 or 3 of those won't give any result worth the effort... However, maybe at least sorting them would be a good idea?

Yeah, this code is quite long and could use some reordering.

1. The alphabetical order will be some improvement, at least the order will become consistent, right now it's sometimes a "historic" order (e.g. I added grayscale versions later, and that's why TGrayscale\*.DrawFromCore is after TRGB\*.DrawFromCore).

  So, go for it:)

2. Possibly, an even better improvement would be to split this per-mode. For each mode, you would create 4 methods, like

  TGrayscaleImage.DrawFromCoreMultiply
  TGrayscaleAlphaImage.DrawFromCoreMultiply
  TRGBImage.DrawFromCoreMultiply
  TRGBAlphaImage.DrawFromCoreMultiply

  Then place these four methods in a separate file, like castleimages_draw_multiply.inc. Then make the main "DrawFromCore" methods only call the appropriate DrawFromCore<CurrentMode> method.

  This is a significant rearrangement, but also probably useful. When adding new mode, you would mostly just create a new include file, instead of modifying the current large file. You would also need to add trivial code in basic "DrawFromCore" to call your new methods, but this would be much simpler (probably).

  Anyway, that's just a suggestion. Improvements are most welcome:)

michaliskambi:
> Plus I've made an additional test utility for testing blending modes and images combination - requires manual blend mode specification in the code.

That's very useful, and I take it!:)

It is committed to the engine code ( https://github.com/castle-engine/castle-engine/commit/d416f49fe3187d08da3c35b198b47b2dabe19d64 ), in castle-engine/examples/images_videos/drawing_modes_test.lpr .
- I changed a little the formatting (whitespace) to suit my conventions a bit (it seems I use more spaces than you usually:)
- I made sure NotApplicable is freed OK.

michaliskambi:
I played around a bit, and added to your demo labels ( https://github.com/castle-engine/castle-engine/commit/0d3cadf0997271cc6165b504a9cbfd8deb6ce0c2 ) and buttons ( https://github.com/castle-engine/castle-engine/commit/c855919e27aa4a444dc963155d33e59db236cb62 ) :)

Issue #13, opened 2015-12-10
Build tool should utilize multicore processors


tsr8:
Build tool and whole castle engine building should utilize multi-core CPUs.

There is "--threads=value" option in fpmake but it doesn’t work for now.

I found a clue that --threads option works if there are many packages to compile. We have only one.

michaliskambi:
Sorry for taking so long to look into this!

I looked into it now, and I don't think this is really doable (at CGE side)... "fpmake" can use threads by calling multiple times FPC in parallel. For this, you not only need multiple projects, they also cannot depend on each other. This is confirmed by looking at ThreadsAmount documentation in fpc/trunk/packages/fpmkunit/src/fpmkunit.pp :

>    // When ThreadsAmount is specified, #threadsamount# worker-threads are
>    // created. When such a worker-thread is ready all worker-threads are evaluated
>    // to see if there are idle threads (there is always at least one such thread.)
>    // To each idle thread a package is assigned which has to be compiled for the
>    // current target and for which all dependencies are compiled earlier.
>    // When no package is available the thread remains idle until another thread
>    // has finished it's task. Compilation stops when all packages are compiled
>    // or when an error occures.

We don't have that luxury in Castle Game Engine units. See https://github.com/castle-engine/castle-engine/wiki/Units-Map . The unit groups all depend on each other. We could split them (like they are splitted into subdirectories now -- base, audio, x3d...). But the groups would tightly depend on each other, so the parallel compilation would not be possible anyway. You have to compile group "x3d" before you compile group "game", you have to compile group "castlescript" before you compile group "x3d" and so on.

Also, this assumes that FPC doesn't try to mistakenly recompile units when not needed. Which unfortunately sometimes happens, due to some FPC bugs (that also manifest in internal errors), sometimes caused by the usage of "inline" routines in the engine (for speed). These problems, right now, are a small nuisance -- sometimes FPC fails and you should just recompile (possibly cleaning before). If we would activate threads, these errors would be very bad, as multiple FPC instances would try to write the same ppu and o files at the same time.

So, using the option --threads with fpmake is not possible for us now...

Same answer goes for the Lazarus packages compilation, BTW. Since some time, Lazarus packages can be compiled in parallel. But this is also done by just executing FPC multiple times. Splitting cge into more Lazarus packages would not help --- as they would depend on each other.

Issue #10, opened 2015-12-06
2D tile map idea


tsr8:
It would be great to have some TiledMap component. It should be compatible with ["Tiled Map Editor"] (http://www.mapeditor.org/) file format. Which class should be an ancestor to "TCastleTiledMap"?

michaliskambi:
Looks cool! Contributions would be most welcome!

It's a 2D control, so it should descend from TUIControl. The unit CastleControls contains many examples of creating 2D controls, and http://castle-engine.sourceforge.net/tutorial_player_2d_controls.php has some (very) introductory info, although I guess you know them already:)

PR #26, opened 2016-10-09
simplify WriteLogMultiline + add a line break after "Message"

eugeneloza:
Simplified the code by universalizing adding the time prefix.
Added a linebreak in multiline, now it looks like

-------------------- asdf begin
asdf
-------------------- asdf end

instead of

-------------------- asdf begin
asdf-------------------- asdf end

### DIFF
--- a/src/base/castlelog.pas
+++ b/src/base/castlelog.pas
@@ -295,16 +295,11 @@ procedure WriteLogMultiline(const Category: string; const Message: string);
 begin
   if Log then
   begin
-    if LogTimePrefix = ltNone then
-      WriteLogRaw(
-          '-------------------- ' + Category + ' begin' + NL +
-          Message +
-          '-------------------- ' + Category + ' end' + NL)
-    else
-      WriteLogRaw(LogTimePrefixStr + NL + //add date&time line only if needed
-          '-------------------- ' + Category + ' begin' + NL +
-          Message +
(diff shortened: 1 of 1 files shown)


michaliskambi:
The simplification is cool:)

Adding newline at the end is not correct, as the Message should already contain newline at the end. WriteLogMultiline documentation explicitly says so:
> Message may be multiline and must be terminated by final newline.

You should use WritelnLogMultiline to have a newline added. I have reverted this change in later commit.

Alternative solutions to the problem of "accidentally calling WriteLogMultiline with no newline at the end of Message":
- Hm, maybe WriteLogMultiline should add a newline, if no newline is detected (no #10 or #13 string at end of Message, and Message non-empty)? If you would like to do it, I would gladly accept this change. 
- I'm also very open to just remove (or deprecate) the WriteLogMultiline function (and force everyone to use WritelnLogMultiline). The WriteLogMultiline (without "ln") is possibly more confusing than helpful.
- Going further, maybe we should even remove (deprecate) the WritelnLogMultiline, and make standard WritelnLog behave like that, if any newline char (#10, #13) is detected inside Message? IOW, just make formatting more automatic, instead of depending that user calls the appropriate version ("Multiline" or "not Multiline"). I have probably made the API of CastleLog less comfortable than it could be:)

PR #27, opened 2016-10-13
Detect NL at WriteLogMultiline, see #26

eugeneloza:
Detect final NL in WriteLogMultiline and if there is none, add one
Tested at Linux and Windows and everything seems to work smoothly. Even if I add spaces after NL symbol - they are removed by "trim" command. I don't check if there are "two or more" NL symbols in the end, but, probably, it's not required.
I think it's not a best idea to detect NL symbol in WriteLnLog, because it'll slow the procedure down, and I don't think user will get any unexpected result if he'll add NL inside a log message. So I think it's best to leave it as it is now. But if you think it's the way to go, I can do that, no problem.


michaliskambi:
Instead of
> copy(MessageNL,length(MessageNL)-length(NL),length(NL)) <> NL

we have ready function IsSuffix is CastleStringUtils:) But I think in this case the check is not required at all, since you do

> MessageNL := trim(Message);

first, then you *know* that there's no whitespace (including no newlines) at the end of MessageNL, never:)

So, if you always do Trim, you may as well just do

~~~~
WriteLogRaw(
         '-------------------- ' + Category + ' begin' + NL +
        Trim(Message) + NL +
         '-------------------- ' + Category + ' end' + NL)
~~~~

This is probably simpler:)

eugeneloza:
> first, then you know that there's no whitespace (including no newlines) at the end of MessageNL, never:)

As far as I see the code it looks for `WhiteSpace = [#0..' '];` If I don't miss anything ' ' is #32 or #33, right? Then, yes, absolutely agreet. Trim is absolutely enough. As LN is #10 in Linux and #13#10 in Windows. Sorry... that was my bad :)

eugeneloza:
I'll fix the commit as soon as I'll switch to Linux

michaliskambi:
Indeed, space is #32 (`begin Writeln(Ord(' ')); end.`) and all the possible newline characters are included in the range:)

eugeneloza:
Looks like I've fixed it.

michaliskambi:
Thank you!

michaliskambi:
Although I would prefer to not deprecate the WritelnLogMultiline. Instead, I would deprecate the WriteLogMultiline (without "ln"). Using the "WritelnXxx" is more consistent with existing WritelnLog and WritelnWarning procedures, used very often in the engine.

So I deprecated the WriteLogMultiline.

I also deprecated the non-multiline equivalent, WriteLog. It was used in a couple of places in the engine... always incorrectly (without newline at the end). This somehow confirms that having both versions "WritelnXxx" and "WriteXxx" that do different things is more confusing and error-prone than helpful:) I definitely overcomplicated this API, I should not have published both WriteLog and WritelnLog.

So now the rule is simple: you should always use "WritelnXxx", and they should work intuitively. The "WriteXxx" are all deprecated.

PR #33, opened 2016-11-07
Major fixes to CastleRandom and minor fixes to CastleJoysticks

eugeneloza:
weird: 2 ancient commits popped up in pull request:

Commits on Dec 31, 2011 
 	@michaliskambi 	michaliskambi 	Tagging the 3.0.0 version of 'castle_game_engine'.
Commits on Oct 12, 2016
	@eugeneloza 	eugeneloza 	Merge branch '3.0.0'

but they don't seem to change any files.


src/base/castlerandom.pas
```diff
@@ -47,11 +47,12 @@   TCastleRandom = class(TObject)
     function RandomBoolean: boolean;
     { Randomly provides "-1", "0" or "1" with equal chances. }
     function RandomSign: longint;
-    { Returns a random number in 0 .. High(LongWord) range. }
+    { Returns a random number in 1 .. High(LongWord) range. }
```
michaliskambi: If Seed always <> 0, then 
- RandomInt64 never returns 0
- RandomBoolean is a bit unfair: 2^31 -1 changes for false, 2^31 chances for true :) :)

Both of these are extremely minor issues of course, not worth correcting. But the true output range (1...xxxx) should be mentioned at the documentation of RandomInt64, just like it is now mentioned for Random32bit.

src/base/castlerandom.pas
```diff
@@ -67,22 +68,50 @@ procedure TCastleRandom.Initialize(RandomSeed: LongWord);
...
+   it's 1000 times slower than CastleRandom,
+   but provides a perfect seed initialization. }
+  //filemode := 0;
+  AssignFile(dev_rnd, '/dev/urandom');
+  reset(dev_rnd);
+  read(dev_rnd,result);
+  CloseFile(dev_rnd);
+end;
+{$ENDIF}
+
```
michaliskambi: I'm unsure about having a special implementation for Linux.

It feels like we're reinventing here something that we shouldn't. We should use the same approach to Randomize as the FPC sources. And if there is a problem in FPC implementation of Randomize, then we should fix the FPC too (especially since using CastleRandom is optional; most applications will use a simple Random, and it's Ok if they don't need the speedup of CastleRandom).

Also, Linux is not the Unix, so in general it's good to avoid {$ifdef LINUX}, and use {$ifdef UNIX}. Unless you really have something specific to the Linux kernel. Your implementation is probably OK for FreeBSD, Darwin (Mac OS X), Android too. (But it should be tested.)

Also, choosing an OS-specific solution often increases the risk of introducing some bugs and not catching them. E.g. I usually use Linux, so the other branch will remain untested by me:) That's not some definite problem, after all we have a lot of OS-specific code in the engine anyway. But when a cross-platform solution is possible, I would prefer it a little.

FPC has been around for a long time, and it's standard library is thoroughly tested on various OSes --- they did the necessary decision-making *and* testing for us:) So I would prefer here to avoid any $ifdefs, and have a simple cross-platform implementation that follows the same approach as FPC Randomize.

src/base/castlerandom.pas
eugeneloza: Oooops. I've forgotten mentioning about randomInt64... and there's a problem we have here. its not 1...high(QWORD) but high(DWORD)...high(QWORD)! because we actually compose it by seed shl 32 + seed, and seed may never be zero in xorshift. 
I've thought about this problem before, but forgotten about it later. I'll see what I can do here.

src/base/castlerandom.pas
eugeneloza: I've made a new random initialization algorithm yesterday evening. There are several problems I was solving of SysUtils.Randomize. It's horribly not-thread safe and our current implementation also wasn't (today's morning tests showed that initializing 16 randoms in 16 threads gives equal seeds for some of them). I hope I've solved this problem (by not a very safe xor-ring QWORD(@c64) which is unique for each thread even if gettickcount64 returns equal results, but I need to test it thoroughly.

eugeneloza: I've fixed randomization on Windows machines. See https://github.com/eugeneloza/castle-engine/commit/9cc21674b8345c0a153f8e40aab9eeb452fdec88 and https://github.com/eugeneloza/castle-engine/commit/c80b92f04245e6d7e13bb52f8763d90ae499cc6c

Not as beautiful as /dev/urandom thou.
 I've tested it thoroughly and it provides really unpredictable random sequence
 by using a 64-bit xorshift algorithm. It also solves problem of "less than 15ms
 randomization".
Tests show it's working as expected https://github.com/eugeneloza/castle-engine/commit/2658d669fce86c18b727365d244f0e0448f0b0b3

P.S. I could never think SysUtils would have such an ugly random seed
 initialization algorithm :). They do so much effort to make the
 random sequence nice, but they don't do anything to have really different
 random sequences... And some users even argue that "don't run randomize more
 than once in your program!", etc.

> And if there is a problem in FPC implementation of Randomize, then we should fix the FPC too

Yes, maybe it'd be ok to submit this procedure as a patch to SysUtils because it's much more random and looks absolutely thread-safe.

> I'm unsure about having a special implementation for Linux.

I've introduced and commented out {$DEFINE USE_DEV_URANDOM}. So initialization works according to equal algorithm on any system at the moment. As far as I see, dev/urandom is not implemented on all unix-like systems, e.g. OpenBSD doesn't have it. However, it looks like all systems have /dev/random implemented (which is slower in Linux). Moreover, some systems might require additional previleges from user to read /dev/random.
So, basically, I'm not sure if it's a good idea to initialize it anywhere except for Linux machines.
The new 64-bit algorithm I've made looks to be very nice as it grabs and integrates 3 semi-independent semi-random variables as CastleTimeUtils.gettickcount64, SysUtils.now and memory allocation address of a local variable. Of course, /dev/urandom is better, but we're not pursing cryptographic safety, but just need to make sure 2 random sequences cannot be equal otherwise than by a random coincidence with a chance of 1/max(longword).

src/base/castlerandom.pas
eugeneloza: fixed RandomInt54 by https://github.com/eugeneloza/castle-engine/commit/6da37266d9c1f328dd5e2de869ea1ac7612b0782

eugeneloza: > RandomInt64 never returns 0

with the last fix it can return 0. Because we return `seed64bit mod N` which can obviously be 0 even in case `N` is `High(QWord)`.
However: `Random32bit` indeed can't be zero. To provide for zero randoms we have to upgrade xorshift algorithm with variable-cycling, e.g. like this one: https://en.wikipedia.org/wiki/Xorshift#Example_implementation
But on the other hand it reduces the speed.
So, basically I'd trade off RandomBoolean 5e-10 (0.00000005%) accuracy loss for speed :) I don't think anything like that might be ever significant. And in case that matters (e.g. like real-money casino), it's highly NOT advised to use any pseudorandom algorithms and revert to /dev/random (not even /dev/urandom) for truly random numbers :)

eugeneloza: Another technical (and might be much more serious) problem of single-variable xorshift is that it has only ONE random sequence 2^32-1 integers long. Random seed just takes the place to start this random sequence. So, e.g. we can get seed[i] and seed[i+1] for 2 our random generators and the sequences they generate will be absolutely equal except for the first element in the first sequence, while seed[i] and seed[i+1] have nothing in common.
A simple example of the problem might be generating two 100x100 simple maps which have ~1000/2^32 chance to contain obviously identical pattern - much higher than it should be normally, but still too low to care about :).
An additional by-side of this fact is that 2 randoms in a sequence cannot be equal unless a full 2^32-1 cycle has passed. This problem is also solved in variable-cycling algorithms. E.g. `TCastleRandom.Random32bit = TCastleRandom.Random32bit` will never be true no matter how many times it is asked (unless accidently seed=0).
If we ask for integer randoms we use `mod` and therefore unless integer is max(DWord) we can get both zero values and identical values.

michaliskambi:
Fantastic:) Thanks for all the work, and for detailed documentation! Applied:)

PR #39, opened 2017-01-12
Utility to convert spritesheets into classic X3D format.

Kagamma:
I have created a simple utility allows to convert spritesheets into ready-to-use classic X3D format.

Support file format:
- Starling (.xml), Fully supported.
- Cocos2D (.plist). Covered most of important stuff, rare features like rotate, polygon sprites are not supported (yet!) but it can be added easily :).

Notes:
- By default anchor will be placed at the center of the sprite if the tool didn't found it in spritesheet.
- Frames must be named after the following format: XXX_YYY. XXX stands for frame name, which will be used as animation name. YYY is frame number. Example: slime_01.png, slime_02.png, ect.
- The tool is test on Windows only, but it should be able to compile on other OSes, as no platform-dependent code is used.

Demo clip from a previous development version but still applied to this version: https://www.youtube.com/watch?v=2JDiwQttgSo

### DIFF
--- a/tools/ss2x3dv/CastleEngineManifest.xml
+++ b/tools/ss2x3dv/CastleEngineManifest.xml
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<project name="ss2x3dv" standalone_source="ss2x3dv.lpr">
+</project>
--- a/tools/ss2x3dv/cocos2dparser.inc
+++ b/tools/ss2x3dv/cocos2dparser.inc
@@ -1,0 +1,119 @@
+procedure Cocos2DParser;
+  procedure ReadQuad(const ASrc: string; out V1, V2, V3, V4: integer);
+  var
+    R: TRegExpr;
+  begin
+    R := TRegExpr.Create;
+    try
+      R.Expression := '{{(\d+),(\d+)},{(\d+),(\d+)}}';
+      if R.Exec(ASrc) and (R.SubExprMatchCount = 4) then
+      begin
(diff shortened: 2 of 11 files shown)


michaliskambi:
Merged! As mentioned on https://sourceforge.net/p/castle-engine/discussion/general/thread/d2a52fb3/ : fantastic job!!!, this will be very useful!

I'll make a post about it on Facebook and Google+ later:)

michaliskambi:
I did some changes:

1. I renamed in the repository "ss2x3dv" to "sprite-sheet-to-x3d". Longer, but I think this is a little better name.

   (I know, it's inconsistent with my own tools called "image2pascal" and "texturefont2pascal", but I don't like my own names anymore, and will rename them one day.)

2. I changed the usage of FPimage and friends into our CastleImages unit and related types. This way the utility supports more image formats:) Internally, our CastleImages uses FPimage for some (most?) formats already.

3. I removed the trailing whitespace.

4. In the future, I may change the writer a bit, to generate X3D using the X3D nodes, not by constructing X3D as a text file. This will allow to save to (XML) .x3d files as easily as (classic) .x3dv files. I recently did a similar "upgrade" to the examples/3d_rendering_processing/tools/castle_anim_frames_to_interpolators.lpr utility.  Or, if you would want to do it, contributions are most welcome of course:)

Can you provide some sample input files (in Starling and Cocos2D) formats? We could commit them e.g. to tools/sprite-sheet-to-x3d/samples/ subdirectory, and use them for testing:)

Again, many thanks! This is a great work!

PR #43, opened 2017-01-25
Enhanced sprite-sheet-to-x3d

Kagamma:
I rewrote the writer to use engine's X3D nodes. I also made a few modifications regarding in how the tool detects anchor and added some spritesheet samples in sprite-sheet-to-x3d/samples subdirectory.


michaliskambi:
Excellent! The rework to use X3D nodes, and the samples/ --- that's precisely what I imagined!

I made two small corrections later:

- You had a memory leak when generating the output. (I extended the manual now to document how to capture these, on http://michalis.ii.uni.wroc.pl/cge-www-preview/manual_optimization.php#section_heaptrc )

  The leak was simple -- the data of TFPGMap is not freed automatically. In FPC >= 3.1.1, there's a trivial class TFPGMapObject that could be used as a drop-in replacement for TFPGMap in this case. But for FPC 2.6.4. and 3.0.0, we free the data ourselves, as in my simple change https://github.com/castle-engine/castle-engine/commit/87b68b3ec122f56220710e2fe0143b14ccd4f8e7 .

- As for DumpExceptionCallStack: we already have utilities for this, in CastleClassUtils -- DumpStackToString and DumpExceptionBackTraceToString. I used them in https://github.com/castle-engine/castle-engine/commit/1dfb9fca4085f95bc063eec0fe8ec0db763bc339 :)

Thank you!

michaliskambi:
Hi,

Be aware that I've done some changes to how sprite-sheet-to-x3d works today.

Most important is that the TimeSensor.cycleInterval is no longer hardcoded as "2". Instead it's calculated looking at the number of animation frames and the FramesPerSecond variable. FramesPerSecond is by default 4, but it can be changed by the --fps=<xxx> command-line option. I hope that this can make the animation duration better by default.

I also did some minor code changes
- To handle --help, --version command-line options (useful for help2man and Debian packaging)
- And I fixed the need for repeating first animation frame ("List.Add(List[0])" in your code), this was actually not needed, one instead has to add one articial key 1.0 at end. The newly generated animations should look the same as the old ones, without the need for "List.Add(List[0])".

E.g. generating the wolf now outputs:

~~~~
$ ./sprite-sheet-to-x3d samples/wolf.plist
Generating animation attack, frames: 4, duration: 1.00
Generating animation run, frames: 6, duration: 1.50
Generating animation sit, frames: 4, duration: 1.00
Generating animation walk, frames: 5, duration: 1.25
~~~~

PR #45, opened 2017-01-27
Support for two extra mouse buttons

tsr8:
I've added support for mouses with more than three buttons. Tested only in GTK window backend.
TMouseButton definition is compatible now with the one in the Controls unit.


michaliskambi:
Thank you!:)

Issue #51, opened 2017-03-06
Create a testcase that local characters work after fixing: Warning: Implicit string type conversion


eugeneloza:
I've got a lot of
```
Warning: (4104) Implicit string type conversion from "AnsiString" to "WideString" (30 times)
Warning: (4105) Implicit string type conversion with potential data loss from "WideString" to "AnsiString" (53 times)
```
when compiling the Engine in Lazarus. Still no problems ever occurred.
However, when I exported my Decoherence data to XML I've eventually remained without Cyrillic characters due to wrong `WideString` to `AnsiString` conversion :)
Making `AnsiString := UTF8encode(WideString)` and `TXMLDocument.CreateTextNode(UTF8decode(AnsiString))` solved the problem correctly (Cyrillic again displaying). Actually the last conversion was redundant - everything worked correctly without it, but, anyway, at least it removed the warning :)
Maybe, it's a correct way to solve the issue (if it needs to be solved)?
It's also possible just to remove the warnings by `{$WARN 4105 OFF}` and `{$WARN 4104 OFF}` if it's intended.

michaliskambi:
This is due to the new encoding-aware strings in FPC >= 3.0.0, and due to the fact that FPC XML units (unlike everything else in FPC RTL) use WideStrings instead of AnsiStrings.

See
- http://wiki.freepascal.org/FPC_Unicode_support
- http://wiki.lazarus.freepascal.org/LCL_Unicode_Support (parts of it outdated, it seems, written before FPC 3.x)
- http://wiki.freepascal.org/Better_Unicode_Support_in_Lazarus (this seems more up-to-date)

I simply did not have the time to research yet how to *correctly* do everything, and avoid these warnings.  Our situation is complicated by the fact that 

- We want to work both with and without LCL. So we cannot blindly follow what LCL does, and we cannot ignore what LCL does either. And there are some "gotchas", see e.g. http://wiki.freepascal.org/Better_Unicode_Support_in_Lazarus#Why_not_use_UTF8String_in_Lazarus.3F .
- We want to work with FPC >= 3.x and FPC 2.6.4, for now.
[...]

PR #61, opened 2017-03-24
Fix implicit string conversion (part 2 :))

eugeneloza:
see https://github.com/castle-engine/castle-engine/issues/51


michaliskambi:
Thank you!:)

PR #65, opened 2017-04-25
Fix Rift -  warning on linux

eugeneloza:
define {$apptype GUI} only for Windows


eugeneloza:
If you don't mind I'll merge such tiny commits myself... :)

michaliskambi:
> If you don't mind I'll merge such tiny commits myself... :)

Absolutely yes, please do! Thank you!:)

PR #69, opened 2017-07-25
Fix $apptype warning on linux in image-to-pascal

eugeneloza:


eugeneloza:
This is a minor fix, so I'll merge it myself.

PR #70, opened 2017-07-27
Fix $apptype warning on Linux

eugeneloza:
Hi, Michalis!
As $Apptype directive is Windows-specific, there's a warning issued on Linux. I've seemed to fixed everything here, thou trivial, the amount of corrections is large (42 files), so it'd be good if you could look it through before merging.
E.g. tools/build-tool/data/standalone/program_template.lpr might correspond to windows-only apps and therefore the check would be unnecessary.
Pay attention that I've also modified files in generics.collections/examples (I'm not sure why are they located in src folder, thou).

### DIFF
--- a/examples/2d_dragon_spine_game/code/castle_spine.lpr
+++ b/examples/2d_dragon_spine_game/code/castle_spine.lpr
@@ -16,7 +16,7 @@
 { "Castle Spine" standalone game binary. }
 program castle_spine;
 
-{$apptype GUI}
+{$ifdef Windows}{$apptype GUI}{$endif}
 
 { This adds icons and version info for Windows,
   automatically created by "castle-engine compile".
--- a/examples/2d_standard_ui/edit_test/edit_test.lpr
+++ b/examples/2d_standard_ui/edit_test/edit_test.lpr
@@ -13,7 +13,7 @@
   ----------------------------------------------------------------------------
 }
 
-{$apptype GUI}
+{$ifdef Windows}{$apptype GUI}{$endif}
 
 { This adds icons and version info for Windows,
   automatically created by "castle-engine compile".
--- a/examples/2d_standard_ui/show_various_ui_controls/show_various_ui_controls.lpr
+++ b/examples/2d_standard_ui/show_various_ui_controls/show_various_ui_controls.lpr
@@ -17,7 +17,7 @@
   This is the program code for the standalone version. }
 uses CastleWindow, Game;
 
-{$apptype GUI}
+{$ifdef Windows}{$apptype GUI}{$endif}
 
 { This adds icons and version info for Windows,
   automatically created by "castle-engine compile".
--- a/examples/2d_standard_ui/timer_test/timer_test.lpr
+++ b/examples/2d_standard_ui/timer_test/timer_test.lpr
@@ -16,7 +16,7 @@
 { Demo of TCastleTimer and TCastleFlashEffect. }
 program timer_test;
 
-{$apptype GUI}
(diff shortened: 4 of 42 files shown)


michaliskambi:
Thanks, applied, everything looks perfect :)

I usually didn't see these warnings, as I compile through the build tool or using `castle-fpc.cfg`, that disable this warning (by passing `-vm2045` to the compiler). But the warning is still visible if you compile from Lazarus. We could disable this warning in every Lazarus project configuration (`xxx.lpi`) of every program... but it seems cleaner to just fix it in code, like you did :)

The `tools/build-tool/data/standalone/program_template.lpr` is used for all desktop platforms (Windows, Linux) so it deserved a fix too, so that's correct

The `generics.collections/examples` can be ignored or corrected -- it doesn't matter much. They are not treated like part of engine examples, I do not check them or compile. They are where they are because the folder `castle-engine/src/compatibility/generics.collections` tries to resemble as much as possible the contents of the https://github.com/dathox/generics.collections , which in turn resembles the contents of the `rtl-generics` directory in FPC sources. That's why `castle-engine/src/compatibility/generics.collections` has inside `src`, `examples`, `fpmake.pp`, `Makefile`, `Makefile.fpc` -- because these are the conventions for packages inside FPC sources. So it's knowingly different from the normal Castle Game Engine conventions. We're only interested in the generics.collections/src/ files, maybe I should just remove everything else, to avoid confusion...

michaliskambi:
I started to remove the `src/compatibility/generics.collections/examples`, but I resigned from it, at least for now. These are nice examples of using Generics.Collections features, so they may be useful for engine developers (like you and me!) as we use Generics.Collections throughout the CGE code now. So, even though these are not "official CGE examples" (not in top-level examples/ dir, not compiled by "make examples"...), let's keep them for now?

eugeneloza:
> let's keep them for now?

Yes, but maybe it's better move them to "examples" (with the commentary above)? Or are they "hard fixed" in this location by relative links?
If I weren't scanning all the engine folder for `$apptype` I would hardly have found them when I needed :)

michaliskambi:
> Yes, but maybe it's better move them to "examples" (with the commentary above)? 

If I move the `src/generics.collections/examples` to top-level `examples/`, then I will feel compelled to add them to "make examples" test, which is checked by Jenkins, and to maintain them (do whatever bugfixing may be necessary, e.g. like this PR) --- which is something I want to avoid... These are only temporary files after all, they are already committed to the FPC SVN repository, and the FPC team maintains them. Since FPC 3.1.1, this code (sources and examples) is part of FPC. We only keep them in CGE to compile with FPC 3.0.x.

Hm, I know, I see the giant hole in my own reasoning here :) The above paragraph implies that I don't want to actually keep these files in the Castle Game Engine repository. If I don't want to spend time maintaining them (since others are already doing this work, in FPC repository), then it's not OK to keep them in CGE repository, and only confuse potential developers who find them.

So I should be brave, and really remove them from CGE repository :) Of course we can link to their sources in https://svn.freepascal.org/cgi-bin/viewvc.cgi/trunk/packages/rtl-generics/ and https://svn.freepascal.org/svn/fpc/trunk/packages/rtl-generics/ (you can do "svn checkout ..." on the latter URL to easily download whole directory). So anyone interested in Generics.Collections examples can get them.

I'll do it shortly :)

michaliskambi:
It is done. I added comments and links to https://github.com/castle-engine/castle-engine/blob/master/src/compatibility/generics.collections/README.md . I think this is better, less code in CGE -> less work for us to maintain it :)

eugeneloza:
Yes, I perfectly agree.

Issue #64, opened 2017-04-09
Use CastleRandom in the Engine (formerly: TGenericStructList shuffle?)


eugeneloza:
Do we need to have a shuffle procedure in a TGenericStructList and similar? I'm speaking of http://forum.lazarus.freepascal.org/index.php/topic,34882.msg229682.html#msg229682 . Never needed anything like that myself yet. However, looks like it's a nice feature worth adding and looks like a simple thing to do (maybe, even make helpers for normal generic lists). If "yes" then I'm on it :)

michaliskambi:
Hm, I would not add it for now. If one of us will need it in an actual game, then I'm all after adding it. But not earlier:)

Reasoning: There are probably hundreds of such "nice to have" methods for a generic list, and implementing them all (without an actual use case)... well, it seems like we could use our time better, to implement something immediately useful:)

That said, it will not make maintenance harder or anything, it's a simple method, so I'm not really against it. So, I leave the decision up to you.

michaliskambi:
I'll leave it open, up to you:)

eugeneloza:
I nearly missed one just a week ago, still solved the problem with a different method. Plus, I've been awfully lazy last week (due to external factors, like illness :)), so I want to do something small but useful :) And I've seen several people asking for and using this approach.
P.S. We don't have any kind of random enabled globally for the Engine? So, basically there are three ways we can do it:
1. Using System.Random. Not optimal, e.g. I don't randomize it due to using CastleRandom.
2. Using an internal unit-wise random. Doesn't interfere with other randoms. But slows down initialization and consumes RAM.
3. Accept a `@random` function as a parameter. Looks like the best variant, because one can use his own initialized TCastleRandom generator to get repeatable random sequences. However, it's not "universal" as we'll accept only internal Castle Random, with, probably, falling back system.random (i.e. default/overload) if `nil` is provided.

michaliskambi:
The existing engine randomness routines (e.g. in CastleUtils) use System.Random right now. It is automatically randomized at the initialization of CastleUtils, which is at the bottom of (almost?) every other CGE unit.

So, this is the one you should use (at least by default).

I think that AD 2 and AD 3 are a bit over-complications of a simple task... unless, like in my previous comment, it's something you actually have a use for. No other randomization routines allow configuring their randomness source for now.

[...]

PR #71, opened 2017-07-28
Significant updates to CastleRandom

eugeneloza:
Addresses https://github.com/castle-engine/castle-engine/issues/64
I've also separated the algorithm from being bound to AnsiString data type as it can handle any data type and made a separate wrapper for AnsiString.


eugeneloza:
Added CastleRandom test

eugeneloza:
I've replaced the native Random with the CastleRandom everywhere I could find any reference. However, this was a very delicate procedure, so it should be double-checked. See https://github.com/castle-engine/castle-engine/pull/71/commits/25f08512f52c8fb126982696b722e8c40dded84f for more details.

eugeneloza:
Checked everything again. Looks like I've lost nothing. FPS game example also compiles and runs fine.

michaliskambi:
> I've replaced the native Random with the CastleRandom everywhere I could find any reference. However, this was a very delicate procedure, so it should be double-checked. See 25f0851 for more details.

Sorry, but I'm not sure do I want this change...

As I mentioned when merging the initial version of CastleRandom, the standard Random is working fine for my games. It was never problematic in terms of speed or randomness. I see the need for CastleRandom if you make a procedural generation using a lot of random class, but by default the standard Random is fine, and I would prefer to keep using it throughout the engine...

eugeneloza:
No problem, then we'll just need to cancel the last commit. However, I'm not exactly sure how to do it :)

src/base/castleutils.pas
```diff
@@ -42,8 +42,6 @@
   Initialization of this unit does some generally-useful things:
 
   @unorderedList(
-    @item(Calls Randomize (so that you never forget about it).)
-
     @item(Sets DecimalSeparator to '.'.
```
michaliskambi: As mentioned in the PR comment, I would prefer to revert this change, and keep using standard `Random` and `Randomize` throughout the engine. Unless there's a good reason... but so far I don't see one -- the standard `Random` worked good enough (fast enough, and is random enough) for my purposes so far. The `CastleRandom` is great to have as an alternative, but for now I think I prefer to not make it a default...

tests/test_castle_game_engine.lpr
```diff
@@ -71,7 +71,9 @@
   TestCastleXMLUtils,
   TestCastleCurves,
   TestCastleTimeUtils,
-  TestCastleControls
+  TestCastleControls,
+
+  TestCastleRandom
```
michaliskambi: No extra line TestCastleRandom

tests/testcastlerandom.pas
```diff
@@ -0,0 +1,75 @@
...
+  "Castle Game Engine" is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  ----------------------------------------------------------------------------
+}
+
+{ Test CastleRandom unit. }
+unit testcastlerandom;
+
```
michaliskambi: Declare using uppercase, `TestCastleRandom`. The Lazarus will take care to save it using lowercase automatically, don't worry about it -- it will be saved as `testcastlerandom.pas`.

tests/testcastlerandom.pas
```diff
@@ -0,0 +1,75 @@
...
+end;
+
+procedure TTestCastleRandom.TestRandom;
+const n_tests = 10000000;
+var RND: TCastleRandom;
+    i: integer;
+    sum: double;
+begin
+  RND := TCastleRandom.Create;
+
```
michaliskambi: Use `Rnd`, not `RND` -- always CamelCase :)

Same for `i` -> `I`

Same for `sum` -> `Sum`

Same for `n_tests` -> `NTests`

michaliskambi:
> No problem, then we'll just need to cancel the last commit. However, I'm not exactly sure how to do it :)

- You could remove or amend this commit and "force push" your branch. But that's GIT magic hackery :) 

- The simpler and safer way is to just add another commit, that reverses the changes of previous commit. The easy way to do it is to use "git revert" command (if you know SVN, don't be confused -- "git revert" is something very different from "svn revert"). If you use TortoiseGit or other GUI, there may be some helper for this (I personally use magit in Emacs, so I cannot help much:)

The additional advantage of the 2nd approach is that it leaves your commit in the history. I may well say some day "OK, we found a usecase that requires that the engine actually uses CastleRandom internally, for better speed or randomness, so let's migrate all randomizations in the engine to CastleRandom", in which case your commit can be taken from history and used.

eugeneloza:
Understood :) But first I have to understand how to cancel the last commit... I'm absolutely sure I've read about it... but I've already forgotten how do I make an automatic "reverse" commit... so that everything would be perfectly back in places.
Hmm... I have another idea - to "mark" all the places where "System.Random" is used in case we'll want to eventually replace it some day (now there are too many words "random" throughout the Engine). 
I've came up with another cool idea - we may make an alias for random - that'd be System.Random for now but will make replacement it in just one click. However, it looks like unnecessary overcomplication :)

michaliskambi:
> Hmm... I have another idea - to "mark" all the places where "System.Random" is used in case we'll want to eventually replace it some day (now there are too many words "random" throughout the Engine).

Indeed, see my above advice about "git revert" --- this way your commit replacing Random -> Rnd will stay in the history, and we can get back to it:)

eugeneloza:
> see my above advice about "git revert" 

Ooops, I've missed this message, sorry.
I'd highly vote against reverting this commit in any future because of one important reason: we'll never know if any more "random" call has been introduced in engine, and missing out just one might result in a disaster. Uninitialized random would give identical sequences each time and that would correspond to a single feature, e.g. temporary files naming - which would force them overwritten uncontrollably, such bugs would be extremely hard to track, so, it's better just to forget it for good, maybe putting something like `//{random}` nearby for easier automatic searching which would still require manual double-checking.
Now... let me check about git revert :)

michaliskambi:
> Uninitialized random would give identical sequences each time and that would correspond to a single feature, e.g. temporary files naming

For this reason, in that potential future when we replace Random -> Rnd, I would propose anyway to:
- still grep for "Random" to catch new occurrences.
- keep the call to "Randomize" at the initialization of CastleUtils anyway, for any games using Castle Game Engine and standard Random,
- Also, for temporary files, you should always use GetTempFileNameCheck or GetTempFileNamePrefix from CastleFilesUtils that check for this :)

eugeneloza:
Git revert worked like a charm! Good, now for the minor stuff :)

eugeneloza:
Looks like everything fixed.

michaliskambi:
Thank you! Everything is perfect now, I see you fixed to CamelCase even more than I marked -- thanks, really, this makes code consistent!

Browsing the patch, I have yet one more suggestion (sorry for not catching in earlier): replace (some) uses of `AssertTrue` to `AssertEquals` (it takes 2 arguments, the 1st should be the hardcoded constant). This is a minor change, in case of assertion failure the `AssertEquals` will make a better error message than `AssertTrue`. So e.g. instead of

```
AssertTrue(StringToHash('') = 0);
```

(this will make a message like "assertion failed") use

```
AssertEquals(0, StringToHash(''));
```

(this will make a message like "assertion failed, expected 0, but got XXX").

It's not always possible to change `AssertTrue` to `AssertEquals`, but it's possible if you have simple comparison `AssertTrue(TestedFunction(...) = KnownCorrectResult)`, then the `AssertEquals(KnownCorrectResult, TestedFunction(...))` is better.

eugeneloza:
Good. I've been lost in those Assert functions :) Coudn't find any normal documentation in FPC. I'll do it right away.

eugeneloza:
Done, tested and working.

michaliskambi:
Thank you! (Sorry for some many minor comments.) Things are perfect now :)

eugeneloza:
Looks like it finalizes https://github.com/castle-engine/castle-engine/issues/64 :)
And thanks for a lot of cool new information for me :) That was very useful.

PR #76, opened 2017-08-21
fix for split string only numbers in obj parser

tigrazone:
I have obj file with
	Ka 0 0 0c
in mtl file and obj parser was crashed when parse string '0c' as number

### DIFF
--- a/src/base/castlestringutils.pas
+++ b/src/base/castlestringutils.pas
@@ -155,6 +155,7 @@ const
   AllChars = [Low(AnsiChar) .. High(AnsiChar)];
   DefaultWordBorders = AllChars - ['a'..'z', 'A'..'Z', '0'..'9', '_'];
   WhiteSpaces = [' ', #9, #10, #13];
+  OnlyNums = AllChars - ['0'..'9', '.', '-','+','e','E'];
   SimpleAsciiCharacters = [#32 .. #126];
 
 function RandomString: string;
--- a/src/base/castlevectors_float.inc
+++ b/src/base/castlevectors_float.inc
@@ -286,8 +286,8 @@ var
   SPosition: Integer;
 begin
   SPosition := 1;
-  Result.Data[0] := StrToFloat(NextToken(S, SPosition));
-  Result.Data[1] := StrToFloat(NextToken(S, SPosition));
+  Result.Data[0] := StrToFloat(NextToken(S, SPosition, OnlyNums));
+  Result.Data[1] := StrToFloat(NextToken(S, SPosition, OnlyNums));
   if NextToken(S, SPosition) <> '' then
     raise EConvertError.Create('Expected end of data when reading vector from string');
 end;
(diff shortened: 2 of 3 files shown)


michaliskambi:
First of all, thanks for the fix!

But, I admit I would prefer to fix it differently. The Vector3FromStr function and friends are used throughout the engine, and it would be a little unsafe if they all "silently ignore" an invalid value in the input. Especially since

- their documentation always explicitly said that "components must be separated by whitespace".  
- and it's a bit surprising, since the letter 'c' would be silently omitted while the letter 'e' would be interpreted as a part of a number.
- and we do it only to account for an invalid OBJ files.

Two alternative suggestions:

1. Use a different function in OBJ reader, i.e. replace line `Result := Vector3FromStr(Line)` in `x3dloadinternalobj.pas` with something like `Result := Vector3FromStrPermissive(Line)` and then implement `Vector3FromStrPermissive` just like in your patch (copy-paste existing Vector3FromStr implementation and augment with `OnlyNums` constant). This way the Vector3FromStrPermissive and OnlyNums can be internal inside the X3DLoadInternalOBJ unit, without affecting the rest of the engine.

2. Alternatively: maybe this "c" has some special meaning? Thought I did not find any mention about it on https://en.wikipedia.org/wiki/Wavefront_.obj_file . If you can attach the problematic OBJ file, I could take a look. In which case we can detect the 'c' and strip it specially, like this:

```
if IsSuffix('c', SomeString) then
begin
  DoSomethingSpecialForC;
  Vector := Vector3FromStr(SuffixRemove('c', SomeString));
end;
```

tigrazone:
2 no special meaning. it's error and numbers and dot and +- and e supported.
  OnlyNums = AllChars - ['0'..'9', '.', '-','+','e','E'];

tigrazone:
[wrong-obj-sportscar.zip](https://github.com/castle-engine/castle-engine/files/1240327/wrong-obj-sportscar.zip)
this file have an error and other obj loaders(on c++ for example, tinyobjloader) reads without errors ignoring 'c'

michaliskambi:
> this file have an error and other obj loaders(on c++ for example, tinyobjloader) reads without errors ignoring 'c'

I see. In this case I propose solution 1. -- add `Vector3FromStrPermissive` as described, and leave existing `Vector3FromStr` untouched. Do you want to implement it? :)

tigrazone:
done

michaliskambi:
I'm terribly sorry for taking so long to answer this! The notification slipped through my inbox, I marked it as "read" but forgot to actually act and check your changes.

Yes, the new version looks OK :)

However, looking at your changes https://github.com/castle-engine/castle-engine/pull/76/files , I have some small requests:

- You left some testing code in `x3dloadinternalobj.pas` checking and printing the loading time. Things related to `ttt:single` and `Time` variables. Please remove it. (BTW, you can measure time using CGE Timer or ProcessTimer utlities, see https://michalis.ii.uni.wroc.pl/cge-www-preview/apidoc/html/CastleTimeUtils.html#Timer for simple sample code.)

- The Vector3FromStrPermissive and OnlyNums are needed only locally. So please move Vector3FromStrPermissive implementation to be inside of x3dloadinternalobj.pas , and please move the OnlyNums constant to be inside the Vector3FromStrPermissive function. This means that Vector3FromStrPermissive is only available in `x3dloadinternalobj.pas`, and that's better (for now -- of course we can move it back to castlevectors_float.inc if it will be useful for other cases too).

Thank you! And again, sorry for the delay. It will not happen again :)

michaliskambi:
I applied a bit modifed version of your patch (besides including my suggestions from previous comment, it also makes a warning in case of such invalid color string).

tigrazone:
ok i'm happy to be useful

PR #77, opened 2017-08-24
Prevent resetting mouse cursor to the same value on SetMousePos

eugeneloza:
I've tested this one only with **castlewindow_gtk.inc**, thou it compiles fine and it shouldn't work in a wrong way anywhere. Resetting mouse cursor doesn't put `onMotion` into an endless loop, so it's fine. I think there is no need to adjust the description, as the issue wasn't actually "fixed", but the code is more fail-safe now.
I've also made the same thing for `CastleControl` in a separate commit, however, I'm not exactly sure about `Mouse.CursorPos` behavior.


eugeneloza:
Thanks a lot! I'm still learning all those style stuff :) but eventually it proves very useful for my project too :)
I'll try to fix it tomorrow, as today I'm knee deep in rewriting interface for Decoherence... Looks like I'll finish all the main tasks till tonight...

eugeneloza:
Sorry for the delay... I've thought I'll be much faster with rewriting my interface... Eventually I've printed out the "coding conventions" to learn them properly :) So, I believe, I'll finish this task today/tomorrow.

eugeneloza:
Done. Looks like I didn't miss anything.
P.S. There is no suggestions on spaces, commas, etc. in `coding_conventions.txt`. Should I add some, or you have them already in some other file (including proper Lazarus settings to automatically handle those)?

michaliskambi:
Thank you! Applied :)

> P.S. There are no suggestions on spaces, commas, etc. in coding_conventions.txt. Should I add some (including proper Lazarus settings to automatically handle those), or you have them already in some other file?

The coding_conventions.txt links to https://edn.embarcadero.com/article/10280 that has a section "4.2 Blank Spaces". We simply follow it. In short, use 1 space around binary operators, like := + - *, e.g. `X := Y + Z * 2`. Declare variables/parameters with 1 space after ":" or "," and no space before, e.g. `function Foo(A, B: Integer): Integer` or `Foo(123, 456)`. This is all shown in https://edn.embarcadero.com/article/10280 , so I don't want to overload our coding_conventions.txt .

This is mostly matching default Lazarus editor conventions, except the spaces around operators, as far as I remember Lazarus by default doesn't add spaces around := so it makes `X:=Y`, you need to select something in Lazarus options (sorry --- I truly don't remember what it was called) to make it put `X := Y`.

eugeneloza:
> This is mostly matching default Lazarus editor conventions

No, it doesn't, I always have to set it up manually every time I reinstall a new version of Lazarus :)
`Tools > Options > Code Tools > Space`
And a similar issue in `Line Splitting`
I don't remember exactly which spaces are missing (I already have it set up), but, as you've mentioned, it's at least "no space before and after `Symbol`", thou afair there were some more (maybe no space after `Comma` and `Semicolon`?) I'm going to install a 1.8.0RC4 soon with FPC 4.0.4 so I'll be able to tell more correctly then.
I'm thinking of making a small "memo" for myself, because I always forget which option I have to change and where they are located. And it's not so intuitive to grasp which option is in IDE options, and which is in Project Options.

> https://edn.embarcadero.com/article/10280 

Yes, thanks for the link. I remember you've given it to me before, but I couldn't find.

PR #80, opened 2018-01-04
Castle paint

eugeneloza:
"Can’t automatically merge" - my favorite :)
So, it looks like I've finally finished this one, we have a working (filled/hollow) ellipse, (filled/hollow) rectangle, line and flood fill.
Everything looks relatively safe and robust, but I'm a bit worried, as I didn't make any thorough testing - just made sure it works in some "marginal cases". The procedures also might misbehave in case they're provided with very incorrect input (like negative radius, width, etc.). I'm not sure if it's reasonable to include them into 6.4, maybe it'd be a good idea to check everything in "real-life" to find hidden bugs if any and make the whole "construction" a bit more fail-proof?
Well, anyway, you can check through the code and give your suggestions, as it looks like very close to stable final variant.


michaliskambi:
After a quick review of the changes, everything looks cool. I'll have a more detailed look tomorrow (today I only "skimmed" over the new `castleimages_paint.inc` that contains a majority of the new code), and will apply.

I definitely want this in CGE 6.4, because I see that you _did_ test it, and it's a useful addition. And even if some bugs crept in --- the changes are quite separated from the rest of CGE, so we can be relatively sure that nothing existing is broken. You can of course always make more tests and submit changes/fixes:)

As for not checking the inputs: that's OK, many other CastleImages routines also don't check (all) the inputs:) The developer is expected to supply sensible values, usually, unless we document in the interface that passing nonsense values is checked.

As for merging -- it seems that the GIT did not properly realize that some of the commits are already applied, "cherry-picked". It may be caused by cherry-picking earlier, the commits extracted and applied by cherry-picking do not have the same GIT hash, so they may confuse GIT. Also, we did modify `castle_base.lpk` a number of times lately, this also probably causes new conflicts. In any case, that's nothing I will not be able to manage, so no worries here:)

Thank you!

eugeneloza:
Yes, I tried to solve to the conflict yesterday evening, and it looked possible this time :). There's a problem that castle_base.lpk got "sorted" in a different way, so castleimages_paint.inc "breaks the order" of units. That was relatively simple to fix, just by deleting it and adding to a new castle_base.lpk. However, the only option I get eventually is "merge commit", the meaning of which is unclear to me - whether this will merge a commit to my repository or will merge the whole pull request (which I'd like to avoid for now). So I decided to keep it this way just to be on the safe side :)

michaliskambi:
Everything applied and working perfectly:) Thank you!!

PR #85, opened 2018-01-29
dmOverwrite in DrawFromCore

eugeneloza:
@michaliskambi you might want to double-check the changes. While straight-forward, I seem to be extremely inattentive today, so I might do something really silly :D


michaliskambi:
Thank you!

PR #88, opened 2018-03-17
Overloaded TCastleImage.DrawFrom3x3

eugeneloza:
Create an overloaded `TCastleImage.DrawFrom3x3` accepting raw variables as `TGLImage.Draw3x3` does. @michaliskambi WDYT?


michaliskambi:
Hm, I admit I deliberately preferred `TCastleImage.DrawFrom3x3` to only have a single overload version, with "concise" parameters (rect as TRectangle, corners as TVector4Integer). There types have helper functions to easily construct them, so one can always call

```
Image.DrawFrom3x3(Rectangle(...), SourceImage, Vector4(...), DrawMode, Interpolation);
```

.. thus providing the left, bottom etc. parameters within the same statement.

I think that this avoids the need for more overloads of `TCastleImage.DrawFrom3x3`. And the same argument works for all other functions that take rectangles / vectors as parameters too, so overall we have simpler API, without creating too many overloads.

I specifically want to *avoid* creating many overloads I invented for `TGLImageCore.Draw3x3`. I was too undecided when designing that API. And as far as I remember, the rectangle types (TRectangle, TFloatRectangle) were not existing when I first introduced `TGLImageCore.Draw3x3`.

With `TCastleImage.DrawFrom3x3`, I wanted to have a cleaner slate :) So I deliberately added only one version.

What do you think? If my arguments sound convincing, then I would not apply this change to CGE. But if you have good arguments to add this overload, I may easily change my mind :)

eugeneloza:
Yeah, that's why I've been asking :)
The idea behind this `overload` was simple: I have to "pack" variables into `TRect` and `TVector4Integer`, I pass them to the `Draw3x3` and then it **unpacks** them again to plain variables. I.e. we have 2 unneeded conversions operations (and needlessly larger "uses" section user-side).
In my case it actually doesn't matter, as I preform the scaling only once per interface item, not every frame, maybe once or twice per second in a normal situation. But it looked a bit inefficient, non-optional.
Of course, speaking of that, using `TCastleImage` for drawing every frame is a waste, so it doesn't really have any serious significance and I don't insist on adding this one, I'm perfectly fine with current version.

michaliskambi:
The inefficiency of packing/unpacking should not matter here, compared to the speed of actual drawing :)

`DrawFrom3x3` calls 
- 9 times `DrawFrom`, which is an intensive loop over many pixels.
- It also typically calls 5 times `Resize`, additional `DrawFrom` and temp image allocation. 

There's no way that you will ever notice the speed of this tiny packing/unpacking :) 

Moreover, 1. you loose speed on packing/unpacking, but you gain speed as you pass fewer parameters (whole TRectangle may be passed as one pointer). And 2. if the caller already had ready TRectangle structure, then just using it is more efficient (in your version, you would unpack TRectangle). So it's not even obvious do you lose or gain speed. In any case, this is not something you will ever notice :)

So here, the "simplicity" of API is definitely more important, as the speed of this should not be ever noticeable. This feeling _"looks a bit inefficient"_ should not be trusted, that is the reason why we have this mantra "never optimize before profiling" :) Here it would optimize something that is not noticeable at all (packing/unpacking), and it would not improve the actual bottleneck inside.

michaliskambi:
I will close it then.

eugeneloza:
Ok! Agreed.
(Uhm... how to revert my fork now? Oh, I'll figure it out eventually. :))

michaliskambi:
1. If others may use your fork, you should probably follow the procedure _Undo published commits with new commits_ on https://stackoverflow.com/a/4114122 . This makes another commit that reverses the effect of the previous one. (if you will make another pull request from this branch, I will see you have these two commits that "cancel each other out" -- and it's OK, not a problem).

2. Otherwise (if you're the only person using your own fork), then you may follow the procedure _Hard delete unpublished commits_ in the same post, https://stackoverflow.com/a/4114122 . This will change the "master" to point to different (e.g. previous commit), completely discarding information about that commit. You have to then "force push" your branch (`git push --force`) to also change the state of remote repository (that is, the one hosted on GitHub server). See also this answer: https://stackoverflow.com/a/5816761 . The `--force` is required since we rewrite the history, which should be avoided if others may use the same repository.

eugeneloza:
Oh, I've just decided not to spam here, yes, I've done a revert commit already.

PR #89, opened 2018-03-23
Fully working network library.

BenediktMagnus:
After I had solved my problems with the Indys and written an Android service, I didn't feel comfortable with my internal solution and thought: "Wouldn't it be much nicer if this was, well encapsulated, a part of the Castle Game Engine itself?" And because the answer was a clear "Yes!" in my head, after a couple of days I had something to share.
The result is a fully working network library integrated into the engine, based on the Indys and a custom implementation for Android. 

It's tested for Linux and Android, but should also work with no problems on the other desktop systems. I cannot say if it's working on iOS.
Someone should have a deeper look into the Java code. I didn't really write anything in Java before, so, even it is working without any problems, there could be some mistakes left.
The Pascal code is fine (except for some less important tweaks like synchronisation of server writes) and can easily be extended for things like encryption or a heartbeat function. Maybe you want to have a look at the internal classes for naming convention. I wasn't sure when to use the Castle prefix and when not.

Attached is a zip with two example files, one for the server and one for the client. As you can see, the usage is very simple. If there are any questions or annotations, don't hesitate to write.

[Examples.zip](https://github.com/castle-engine/castle-engine/files/1843653/Examples.zip)


BenediktMagnus:
Thank you! That's great to hear.

About the names: I had in mind that it should be possible to integrate more protocols in future (like UDP or maybe even FTP and HTTP, all with optional encryption) so the names were broad intentionally. For the classes it makes indeed sense to tag them with the protocol (as in this way one can add more classes for other protocols) but the unit name should stay "CastleNetwork" in my opinion.
Maybe change the service name instead? I don't know what would make more sense as I don't know how a UDP, FTP or HTTP server would be implemented in Java. If it is only a change of the underlaying socket, combining them in one service should be the best to do.

I have changed the breacket thing.

And a full example will follow!

michaliskambi:
Ah, OK, I understand the idea to add more protocols. How about calling the unit `CastleClientServer` then?

As the name `CastleNetwork` sounds too broad, there are various things than I could imagine would deserve to be called "network" :) E.g. downloading stuff over the network (like our `CastleDownload`). Or integration with a larger networking library, like https://github.com/BeRo1985/rnl . Or a unit to replicate classes over the network, like Unity3d networking that replicates GameObjects.

As for adding UDP: I don't think it's that simple, as UDP loses packets. So you need to write code on top of UDP to secure from it, and behave sensibly even when the packet is lost. What exactly to do (just resend? or wait for next packet of information that supersedes the previous one?) depends on the use-case. Unless you would leave this task at the responsibility of users.

BenediktMagnus:
`CastleClientServer` is a great idea! And you are right with `CastleNetwork` being to broad at this point. So `CastleClientServer` seems to be the best solution. I have renamed it.
I have also consistently renamed the service to `ServiceClientServer`. So new protocols can be added to both without having to change the API.

About UDP: Yes, it's much harder than TCP. And in most cases TCP is the easiest and most reliable protocol, that's why I used it in the library.
If you wanted to replace TCP with UDP, you true had to write your own handling of packet lost. But this can, because the way TCP is solving that is very generic, for very specialised cases be more efficient than TCP. Mostly one wants to use UDP if packet lost isn't that common (for example in a local network), for broadcasting (to find other game instances in the local network) or when packet lost doesn't matter, for example when transporting audio or high frequenced position data.
In most of these cases the user of the engine would need to handle more than with the current TCP implementation (when it shall bring a benefit), that's for sure.
But all of that is a thing of the future. By now, having that in mind is a good thing to prevent having to change the basics (the API) later again.

src/files/castleclientserver.pas
```diff
@@ -0,0 +1,737 @@
...
+
+  "Castle Game Engine" is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  ----------------------------------------------------------------------------
+}
+
+unit castleclientserver;
+
```
michaliskambi: Write it CamelCase, `unit CastleClientServer;`

(Note: The filename should stay lowercase, `castleclientserver.pas`, to be reliably found by FPC on case-sensitive filesystems. But the unit name is better CamelCase -- looks nicer, and makes code completion work :)

src/files/castleclientserver.pas
```diff
@@ -0,0 +1,737 @@
...
+  "Castle Game Engine" is free software; see the file COPYING.txt,
+  included in this distribution, for details about the copyright.
+
+  "Castle Game Engine" is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  ----------------------------------------------------------------------------
+}
+
```
michaliskambi: Please place a short documentation before the unit name.

We use this convention with all Castle Game Engine units, so that they look nice when viewed on a generated units list on https://castle-engine.io/apidoc/html/AllUnits.html . The 1st sentence (up to the 1st dot) is special (we use PasDoc with auto-abstract and auto-linking).

I propose this:

```
{ Client and server for network communication using TCP (TCastleTCPClient, TCastleTCPServer). 
  Use this in your games for easy client/server communication. 
  On Android,it requires adding a service client_server (see https://github.com/castle-engine/castle-engine/wiki/Android-Project-Services-Integrated-with-Castle-Game-Engine). 
  On other platforms, it requires having Indy (http://www.indyproject.org/ ) available 
  (note that Indy is not distributed together with Castle Game Engine, 
  you need to download it yourself, and point the compiler to find Indy sources). }
unit CastleClientServer;
```

tools/build-tool/data/android/integrated-services/clientserver/app/src/main/java/net/sourceforge/castleengine/ServiceClientServer.java
```diff
@@ -0,0 +1,270 @@
...
+    public String getName()
+    {
+        return "serviceclientserver";
+    }
+
+    @Override
+    public boolean messageReceived(String[] parts)
+    {
+        if ((parts.length < 4) || (parts.length > 6) || !parts[0].equals("serviceclientserver")) //serviceclientserver key action param1 (param2) (param3)
+            return false;
```
michaliskambi: The message string should be `client-server`, so

```
...parts[0].equals("client-server")...
```

With hyphens, and without initial `service` --- for consistency with other messages, we use hyphens there for multi-word strings. All the code sending/receiving messages should be adjusted to it too, naturally.

tools/build-tool/data/android/integrated-services/clientserver/app/src/main/java/net/sourceforge/castleengine/ServiceClientServer.java
```diff
@@ -0,0 +1,270 @@
...
+
+    public ServiceClientServer(MainActivity activity)
+    {
+        super(activity);
+    }
+
+    public String getName()
+    {
+        return "serviceclientserver";
+    }
```
michaliskambi: The service name should be `client-server`. With hyphens, and without initial `service` --- for consistency with other services, we use hyphens there, see e.g. https://github.com/castle-engine/castle-engine/blob/master/tools/build-tool/data/android/integrated-services/google_play_games/app/src/main/java/net/sourceforge/castleengine/ServiceGooglePlayGames.java#L60 inside `ServiceGooglePlayGames`.

(P.S. I know, this is not consistent with subdirectory names, where we use underscores instead of hyphens. I may change this one day :) But it should be changed in all services. For now, other service use hyphens in `getName`.)

tools/build-tool/data/android/integrated-services/clientserver/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
michaliskambi: Place the service inside a subdirectory with underscores, `data/android/integrated-services/clientserver/` instead of `data/android/integrated-services/client_server/`. We use underscores to name services.

tools/build-tool/data/android/integrated-services/clientserver/app/src/main/java/net/sourceforge/castleengine/ServiceClientServer.java
```diff
@@ -0,0 +1,270 @@
...
+                            messageMap.remove(key);
+                            activeMap.remove(key);                       
+                        }
+                        else
+                            CreateListenerThread(new Socket(host, port), key + "client", socketMessageMap, socketActiveMap);
+                    }
+                    catch (IOException e)
+                    {
+                        Log.d("ServiceClientServer", e.toString());
+                    }
```
michaliskambi: Do not use `Log.xxx` in our services. Instead, use only `logError`, `logWarning` or `logInfo` (they are defined in ancestor `ServiceAbstract`), with `CATEGORY` as the 1st parameter, like

```
logError(CATEGORY, e.toString());
```

Underneath, our `logError` calls Android's `Log.e` method. But it also:
- Makes sure that the messages can be nicely filtered when using `castle-engine run` or `adb logcat -s appName:V`. So it automatically prefixes all log messages with application name.
- And it secures from Android log tag being limited to 23 chars.

BenediktMagnus:
I have implemented your suggestions. (Don't know why Github doesn't recognise it completely.)

The example is also included. Two separate projects, one for the server, one for the client, under "tcp_connection".

I hope all CGE requirements are now complied. If you still find something, let me know. :)

michaliskambi:
Thank you! Everything is perfect now:) Merged!

PR #90, opened 2018-03-26
Add GetInteger with default value overload to CastleXMLConfig.

BenediktMagnus:
Simple thing: Why didn't "GetInteger" had an overload with a default value in the CastleXMLConfig? So, now it has.

### DIFF
--- a/src/files/castlexmlconfig.pas
+++ b/src/files/castlexmlconfig.pas
@@ -68,9 +68,12 @@ type
     constructor Create(AOwner: TComponent); override;
     destructor Destroy; override;
 
-    { Get a @italic(required) integer attribute, raise exception if missing or invalid.
-      @raises(EMissingAttribute If the attribute is missing or empty.)
+    { Integer values reading/writing to config file.
+      @raises(EMissingAttribute Raised by GetInteger(string) (overloaded
+      version without the ADefaultValue parameter) if the attribute is missing.)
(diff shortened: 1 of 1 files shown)


michaliskambi:
The TCastleConfig descends from TXMLConfig. It is documented e.g. here: http://wiki.freepascal.org/xmlconf

`<sidenote>`
The documentation on http://wiki.freepascal.org/xmlconf actually talks about TXMLConfig in XMLConf unit, using WideString. While we use TXMLConfig from XmlCfg, using normal AnsiString. Actually we use our fork of XmlCfg in CastleXMLCfgInternal. In any case, the methods documented there are the same, we just have `string` instead of `WideString`.
`</sidenode>`

TXMLConfig already has methods GetValue / SetValue / SetDeleteValue that cover getting/setting Integer, boolean and string. In particular they have a getter that has a default value:

```
function GetValue(const APath: WideString; ADefault: Integer): Integer;
```

That is why our methods to deal with Integer, boolean and string are limited. Compared e.g. to our methods to deal with Float or Int64.

Anyway: if it wasn't obvious to you, then it will probably confuse also others. So, I agree -- we can add GetInteger with default value, to be consistent with `GetInteger` without default.

BenediktMagnus:
Ohh, okay, I didn't get that.
But thanks for merging! Having the half with GetValue and the other half with GetType is very confusing.

PR #91, opened 2018-03-26
Simple Android keyboard function.

BenediktMagnus:
Simple function for opening and closing the software keyboard on Android. Besides that, when focusing an CastleEdit the keyboard opens, when it loses the focus, it closes.

One thing has to be mentioned here: Because Android toggles the keyboard visibility, a try to close it when it is still closed will open it instead. So when the end user closes the keyboard manually with the back button, the keyboard will open as soon as the edit loses its focus. This is an unwanted behaviour I've often seen in Android apps. Workarounds could be to intercept the back button or to poll for a view height change. Both doesn't seem to be trivial in an Android service, so I suggest to leave it this way for now and let the user of the engine (respectively the user of the game app) handle it. It's still much better than having absolutely no keyboard.


src/ui/opengl/castlecontrols.pas
```diff
@@ -104,7 +104,7 @@ implementation
 
 uses SysUtils, Math, CastleControlsImages, CastleTextureFont_DjvSans_20,
   CastleTextureFont_DejaVuSans_10,
-  CastleApplicationProperties;
+  CastleApplicationProperties{$ifdef ANDROID}, CastleMessaging{$endif};
```
michaliskambi: Do not surround `CastleMessaging` use with `{$ifdef ANDROID}`. Use it on all platforms. It will be ignored on platforms that don't support it, which is totally fine. This way adding iOS implementation for this will be trivial -- it will just handle the same message.

src/ui/opengl/castlecontrols_edit.inc
```diff
@@ -361,4 +362,16 @@ begin
     Result := inherited CapturesEventsAtPosition(Position);
 end;
 
+procedure TCastleEdit.SetFocused(const Value: boolean);
+begin
+  inherited;
+
+  {$ifdef ANDROID}
```
michaliskambi: Same as above -- simply do `Messaging.Send` always, not only on Android, so no need for `{$ifdef ANDROID}`. This way it will also be useful on iOS.

src/ui/opengl/castlecontrols_edit.inc
```diff
@@ -361,4 +362,16 @@ begin
...
+  if Value then
+    Messaging.Send(['change-keyboard-state', 'on'])
+  else
```
michaliskambi: Use TMessaging.BoolToStr to send a boolean to a message. So this can be just

```
Messaging.Send((['change-keyboard-state', TMessaging.BoolToStr(Value)]);
```

There is an equivalent method in Java to decode the boolean:)

tools/build-tool/data/android/integrated/app/src/main/java/net/sourceforge/castleengine/ServiceMiscellaneous.java
```diff
@@ -91,6 +105,10 @@ public boolean messageReceived(String[] parts)
             Toast.makeText(getActivity().getApplicationContext(), parts[1], Toast.LENGTH_SHORT).show();
             return true;
         } else
+        if (parts.length == 2 && parts[0].equals("change-keyboard-state")) {
+            changeKeyboardState(parts[1]);
+            return true;
```
michaliskambi: Use `stringToBoolean` to convert `parts[1]` to boolean, and pass it as a boolean to `changeKeyboardState`. So 

```
changeKeyboardState(stringToBoolean(parts[1]));
```

This is consistent with our other services, e.g. https://github.com/castle-engine/castle-engine/blob/master/tools/build-tool/data/android/integrated-services/google_play_games/app/src/main/java/net/sourceforge/castleengine/ServiceGooglePlayGames.java#L631

michaliskambi:
As for toggling: hm, maybe you could use instead methods like `showSoftInput`, `hideSoftInputFromWindow` ?

michaliskambi:
Addendum to previous comment (sorry for so many comments, I'm slowly waking up :) :

To better explain AD 2 (AutoOnScreenKeyboard): Of course TCastleEdit is not functional on Android without on-screen keyboard. But maybe a particular game can have it's own reliable strategy to control on-screen keyboard visibility. Our SetFocused is not guaranteed to be reliable for this always, it was not codes with such use in mind.

BenediktMagnus:
Okay, I have included your requested changes.
There are still some things to do before it can be useful.

BenediktMagnus:
About `showSoftInput` and `hideSoftInputFromWindow`: They need a view, this is a little bit more complicated, that's why I didn't choose it at first, but maybe it's the better way in the long run.
The more I thing about it, the more seems this pull request a bit too early. But at least will it allow others to contribute in this topic.

michaliskambi:
1. As for `View`: you can get it in a service by

* `getActivity().getWindow().getDecorView()`. Or you can get 
* `getActivity().findViewById(android.R.id.content)`

You can grep the `castle-engine/tools/build-tool/data/android/` to see where they are used.

Although I have not checked is the `View` returned by one of these methods suitable for `showSoftInput` and `hideSoftInputFromWindow`. We are inside a `NativeActivity` (our `MainActivity` descends from Android standard `NativeActivity`), so things may be a little different than in a typical Android UI application.

2. I noticed one more tiny formatting issue: in `changeKeyboardState` in Java, please indent always by 4 spaces. Never tabs, always 4 spaces in Java (and 2 spaces in Pascal :) ).

3. As for:

> the more seems this pull request a bit too early. But at least will it allow others to contribute in this topic.

As this topic revolves around a couple of TODOs in CGE, it's normal that we achieve only something "partial" first. As long as we track what we need to do to make everything perfect (my TODOs around this are mentioned in comment https://github.com/castle-engine/castle-engine/pull/91#issuecomment-376837173 ), then I think it's all cool :) So, it's not a problem for me that not everything is finished at once.

Please fix AD 2 and let me know do you plan to experiment with `showSoftInput` and `hideSoftInputFromWindow` now, following my suggestions how to get `View` instance. If not, we can apply this PR, and leave the TODOs for later, no worries.

BenediktMagnus:
I have changed the tabs to spaces. (It seems that I used, for whatever reason, gedit instead of VS Code for the last change, so the tabs didn't get replaced automatically.)

I have also build in the `TCastleEdit.AutoOnScreenKeyboard` property.

The TUIContainer solution and the tests with `showSoftInput` and `hideSoftInputFromWindow` is a thing for later. First I have to solve some other things for my current project (which will surely result in some other pull requests). So it is your choice if you apply this pull request now or leave it open until then.

michaliskambi:
Cool. Merged, thank you!

I will use it to test keyboard handling on Android, I'll hopefully find a bit of time to handle it soon.

PR #92, opened 2018-04-08
Localisation system.

BenediktMagnus:
Handy localisation system.

The usage is similar to the sound engine.

The localisation strings are stored in a XML file in this format:
`<?xml version="1.0" encoding="utf-8"?>`
`<strings>`
`  <string key="ExampleKey1" value="Example value 1" />`
`  <string key="ExampleKey2" value="Example value 2" />`
`</strings>`
as `data/local/en.xml` for example.

You can load a language file by this:
`Localisation.LoadLanguage(ApplicationData('local/en.xml'));`

Or with the current system language:
`Localisation.LoadLanguage(ApplicationData('local/' + Localisation.SystemLanguage + '.xml'));`

Controls with text can be inherited by the new TUIControlLocalised (happend for button, label and edit), a child of TUIControlFont, for automated localisation:
`MyButton.LocalisationID := 'ExampleKey1';` displays the localised value for the key. If you change the display language by loading another language via `Localisation.LoadLanguage`, it will update the localisation of all controls with a set localisation ID automatically.

You can furthermore react to a language change by subscribing for the update localisation event:
`procedure UpdateTextOnSite of object;`
`Localisation.OnUpdateLocalisation.Add(@UpdateTextOnSite);`

And last but not least, localisation of the application name on Android works by adding this in the Android section of the CastleEngineManifest.xml:
`<localisation>`
`<appname lang="en" value="My English App name" />`
`<appname lang="de" value="Mein deutscher Appname" />`
`</localisation>`

I don't think an example project would be needed, a section in the documentation should be enough. But if you think it was needed, I'd make one nevertheless.

As always, if you have any additional questions or change requests, don't hesitate to ask.


michaliskambi:
(Uh, I'm sorry for writing so long answer...)

I spent today experimenting, and I have to say that I prefer to use GetText, not a custom translation format.

I made a demo using CGE and GetText ( https://github.com/castle-engine/castle-engine/tree/master/examples/localization ), and it works nicely. I list detailed reasons in favor of GetText below.

I am sorry, this means that I would like to reject a large part of your pull request. In particular, I would prefer to *not* apply reading your own XML file format for translations (large part of your castlelocalisation.pas), and all CastleControls changes. (I still want to take the rest of your changes -- obtaining the language code on Android, and CastleEngineManifest.xml modifications for Android localizable caption.)

I know it sucks to have a large part of your work rejected, but I hope you will see my reasons. I did a research (and test) today whether it is reasonable to use GetText, and it seems it is. The reasons:

- GetText (in particular, it's PO format) is a standard (outside of Pascal community too), used by a large number of open-source projects. E.g. it it used also by Lazarus. So people and tools are already familiar with it.

- ‎GetText already has a translation format. And tools/GUI editors for it (https://poedit.net/ ). Since it's popular, translators are probably already familiar with it. It also has mechanisms for dealing with plural forms (although I'm not sure whether it's usable with FPC GetText).

- ‎GetText can be applied easier, to everything throughout the engine, without the need for TUIControlLocalised or other modifications to CastleControls. We just put our string constants in resourcestrings. So we don't need to adjust existing CGE code to "react" to the translation system, instead we have a translation system that just works for everything -- I think that this is a huge gain. In fact some engine units already used `resourcestring`s for messages, and they are already translatable.

- Lazarus has documentation how to use it for Lazarus applications ( http://wiki.lazarus.freepascal.org/Step-by-step_instructions_for_creating_multi-language_applications , http://wiki.lazarus.freepascal.org/Translations_/_i18n_/_localizations_for_programs ). And they have code to integrate it smoothly with lfm editing. I would like to be able to use it as much as possible in the future CGE visual editor (one of my big 2018 plans). So it's better to not invent our own translation system, at least in the scope of CGE.

- I made a demo of using GetText + CGE in `examples/localization/`: it shows how the GetText approach works. I wanted to make sure that what I'm proposing ("just use GetText") is actually workable :) And I think it is. We will want to localize our game "The Unholy Society", and for me --- this is the way to go. Most of our strings in "The Unholy Society" are user-interface (using CGE UI from TUIControl) and custom dialogue text format ( https://docs.google.com/document/d/1kA7oLsMyhIxCCqNYNRnmhw3IxZhDXPq21BoPP3Y_uKA/edit?usp=sharing ). I think GetText will work nicely with this.

- I also looked at online tools. https://translate.google.com/toolkit/list#translations/active allows to collaborate on translations, and use automatic Google translation mechanisms, for free. And they allow uploading `po` files. So, using a standard format for translations has a big benefit here.

- Note that if you really don't want to use GetText, then it is possible to implement custom translation system (like your XML) without changing the CastleControls. So, while I prefer to not include your system in the "core" Castle Game Engine, it's certainly possible to use it in your own games.

    You can iterate over Window.Controls, for each TUIControl also iterating over it's children, and do something special for TCastleLabel/TCastleButton/TCastleEdit and so on. You can read the initial "translation id" from their Caption/Text, and save it for later in a map control->translation id. I realize that it's not as clean and comfortable as just adding "translation id" into the TUIControlLocalized hierarchy, but it's possible. And it does keep the engine code simpler, independent of the translation system. So, while I would encourage to use GetText system, it is still possible to "roll your own".

I know that GetText approach has disadvantage: 

- To switch a language in the middle of the game, you need to call `TranslateResourceStrings` and assign again the resourcestrings to UI controls. While it's possible to code your game to do this, it's probably easier to just let user restart the application.

    But this is really a disadvantage of "translate resourcestrings" approach. It's not a problem of PO file format. So you can make a general utility "translate this hierarchy of TUIControl" using my idea "iterate over Window.Controls..." above. You can use `TMOFile.Translate` to explicitly translate strings using GetText.
    
    Coding this may be more difficult than the current system, but 1. I think that using standard GetText format is worth the extra work, 2. the fact that translation system will not require modifying CastleControls unit is an advantage for me. The UI is already complicated enough, I prefer the translation system to be independent from it.

Note that of course I still want your code to:

- Get language from Android. A `Localisation` singleton that exposes the current language code, as string, is definitely something I want. (But make it `Localization`, see below :) And it could be useful by both GetText approach, or your custom XML approach in your games.

- Localize the application caption. As you did, build tool must be aware of this so it should indeed be inside `CastleEngineManifest.xml`, to put the localization inside resources.

    One comment I have about this is that this is not really specific to Android (e.g. generating Linux .desktop files would also benefit from this), even if we use it for Android only at the beginning. So I would prefer to have it in top-most section of `CastleEngineManifest.xml`, that is: outside the `<android>` element. And call it `caption`, since that's what the attribute is called in `CastleEngineManifest.xml`. I propose a syntax in CastleEngineManifest.xml like this:

    ```
    <?xml version="1.0" encoding="utf-8"?>
    <project caption="My English App name" ...>
      <localization>
        <caption language="en" value="My English App name" />
        <caption language="de" value="Mein deutscher Appname" />
      </localization>
      ...
    </project>
    ```

- Minor note: Use "localization", not "localisation" everywhere (class names, unit names). "Localization" seems a *bit* more popular ( http://wiki.lazarus.freepascal.org/Step-by-step_instructions_for_creating_multi-language_applications , https://en.wikipedia.org/wiki/Video_game_localization , https://en.wikipedia.org/wiki/Internationalization_and_localization versus https://en.wikipedia.org/wiki/Language_localisation ). Googling "localisation or localization" seems to confirm that for US English, "localization" is a *bit* more preferred.

- Do not use own system-dependent code in `TCastleLocalisation.SystemLanguage`. Instead use https://www.freepascal.org/docs-html/fcl/gettext/getlanguageids.html . This should wrap the system-specific logic, at least for some systems. I suspect that it will be OK for at least Windows and Linux.

BenediktMagnus:
First: Thank you very much for your detailed answer. I really appreciate it!

Second: Even if it's not the best thing in the world to have code rejected, it is something one have to consider when writing it, especially if he submits something on his own without previous discussion. Furthermore, it could be worse than this, because I will be able use the code for my own work anyway.


Now the answer to your decision:

Even if I accept your point of view, I still disagree (mostly). As I mentioned shortly in Discord, I have tested around multiple times with resource string localisation and it isn't easy. Using XML and a localisation wrapper is much more comfortable. The barrier of entry is much lower and things like maintaining and extending is much simpler. You can see yourself how much effort was needed for your example project; the same implemented with CastleLocalisation would mean much less work.

You say GetText was a standard, that's absolutely true, but XML is one, too, especially in newer systems (like .NET and Android). And the wrapper I wrote could easily be extended to new file formats like XLIFF, JSON, CSV, Properties or binary ones for commercial solutions (there are a lot of them out there). You could even use the wrapper class for easy usage of resourcestrings, especially making the change of the language on the fly much cleaner. (But then, reading po/mo in the wrapper would make more sense.)
So, what mainly popped out in my head reading your answer and writing this statement: Why should only one solution be possible? Why accepting the downsides of one or the other when we could combine them and have many more benefits or give, at least, the user a choice. I am very sure there will be people accepting GetText more and others who like the clean wrapper with any format.
What do you think about this?


For the sake of completeness, here are answers to the other points:

`TUIControlLocalised`: I thought this would be an elegent solution: Let the visual element handle themself their localisation, encapsulated and transparent for the user. He only has to use `LocalisationID` and everything else happens automatically. How makes this UI more complicated?
By the way: I first had the idea to iterate over a controls list, or to iterate through children, but these ideas were all, as you describe it yourself, much less elegant and the overhead was bigger.

Putting `<localisation>` to the project element will be done.

"Localization" instead of "localisation": If you want this, okay, I will change it. I personally prefer the British "localisation" because it's cleaner and there is no reason for the z, even if it is (due to the American influnce) less common in programming.

`GetLanguageIDs`: Oh, great! Thank you for this! I didn't know about the FreePascal wrapper and couldn't find one when I searched for it.

michaliskambi:
1. As for the "complication" that I mentioned with `TUIControlLocalised` approach: What I meant by this is that each control with text has to descend from `TUIControlLocalised` and implement `UpdateLocalisedText` to benefit from the system. So the code change is not limited to introducing `TUIControlLocalised` class, we also need to modify other classes. 3 classes are in your commit (button, edit, label), and it should be extended to TCastleProgressBar, TStateDialog, TCastleNotifications in CGE, and if a particular game implements custom TUIControl using `Font.Print` -- it will need to do this too.

    I readily admit that it is not a large "code complication". It's a "small change", implementing something like `TCastleButton.UpdateLocalisedCaption` is trivial. But it needs to be repeated for a couple of classes.

    Still, it's more than "no change at all", like with GetText approach (or my proposed "You can iterate over Window.Controls..." approach). When possible, I prefer a solution that does not require any change at all to the engine core :)

2. I do not see the barrier of entry for GetText really higher than for custom system using XML. I was actually pleased with how easy it was to make my example `examples/localization/`.

    The way I see it, the only additional complication is that you need to convert PO->MO, instead of directly loading PO files. But this is easily automated by a script that can be called before each compilation. The rest of the work seems comparable to me: instead of editing XML in your system, we edit PO files in GetText. Instead of assigning translation id in your system, we assign a resourcestring.

    And the benefits seem important for me: PO editor for translators, Google service for automatic translation of PO files, knowledge of translators familiar with GetText, and being consistent with what Lazarus does.

3. I indeed do not want to limit CGE games to use only GetText. That would be unfortunate.

    In fact, I was happy that the integration of GetText with CGE in my example did not require *any* modification to the CGE code.

    I believe that my proposed approach with "You can iterate over Window.Controls..." is a workable way to implement alternative translation systems. And it can be done independently of the engine. And it is indeed independent from the translation format (it can work with PO, custom XML, any other format).

    Note that for "translation id" in this case you can use control's `Name` (inherited from `TComponent`). In my earlier comment, I suggested """You can read the initial translation id from their Caption/Text, and save it for later in a map control->translation id.""" --- this is not necessary if you just use `Name` as a key.

    Actually, we could add a simple global callback `OnUIControlRename`. This would be called each time TUIControl descendant changes `Name`, by trivially overriding `TComponent.SetName` ( http://lazarus-ccr.sourceforge.net/docs/rtl/classes/tcomponent.setname.html ) inside TUIControl. A custom translation system could listen on this callback to immediately adjust component's Caption/Text based on `Name`. To support the language change later, the translation system could remember all components it received through `OnUIControlRename` in a TComponentList ( https://www.freepascal.org/docs-html/fcl/contnrs/tcomponentlist.html ), which is automatically updated when a component is destroyed. When language change occurs, it just updates all components currently present on the list. Bonus points: if you ever need to, you can use the same approach for any `TComponent` descendant of your own, not just for CGE TUIControl.

    I think this would be better than my initial idea "You can iterate over Window.Controls...". It makes implementing custom translation system easy, following the above paragraph, and clean (no need to iterate over a tree of controls at some carefully-chosen time). And I have no problem with a minimal change to the engine it introduces: just make `OnUIControlRename` working, which is local change in TUIControl only :) What do you think?

michaliskambi:
(
I removed the mention of TCastleProgressBar, TStateDialog, TCastleNotifications from my comment above -- that was not relevant.

They render text, but the text is usually assigned from other code, so translating them by something like `UpdateLocalisedCaption` would not OK. In their case, `resourcestring`s in case of my GetText approach or just explicit translation by something like `Localisation.Get(...)` in case of your system is necessary.
)

michaliskambi:
We don't have a common class with `Caption`, because various text-containinig controls have various needs. There are controls that allow to express their text as `TStrings` and can be multiline (TCastleLabel, future TCastleMemo), there are controls that cannot be multiline (TCastleButton, TCastleEdit), there are specials like TStateDialog. There are controls like future TCastleRadioGroup that has text as Caption:string + Items:TStrings -- translating them means translating both Caption and Items.

Our API is similar to Lazarus LCL or Delphi VCL controls here -- various controls have various needs to express text, not always a single "one string per control".

To explain better my suggestion about `OnUIControlRename`:

Add this to CastleUIControls:

```
type
  TUIControl = class
  protected
    procedure SetName(const Value: string); override;
    ...
  end;

var
  { Notification when TUIControl.Name changes, for custom translation systems. }
  OnUIControlRename: TNotifyEvent;

implementation

procedure TUIControl.SetName(const Value: string);
begin
  if Name <> Value then
  begin
    inherited := Value; // this already checks is name valid and unique in owner
    if Assigned(OnUIControlRename) then
      OnUIControlRename(Self);
  end;
end;
```

And you can now create a localization unit comfortably, independent of the engine:

```
type
  TMyLocalization = class
  private
    FKnownComponents: TComponentList;
    procedure ControlRename(Sender: TObject);
    function Translate(const ControlId: string): string;
    procedure TranslateComponent(const Component: TComponent);
    procedure SetLanguage(const NewLanguage: string);
  public
    constructor Create;
    destructor Destroy;
    property Language: string read FLanguage write SetLanguage;
  end;

implementation

constructor TMyLocalization.Create;
begin
  FKnownComponents := TComponentList.Create;
  OnUIControlRename := @ControlRename;
end;

destructor TMyLocalization.Destroy;
begin
  FreeAndNil(FKnownComponents);
end;

procedure TMyLocalization.TranslateComponent(const Component: TComponent);
begin
  if Component is TCastleLabel then
    TCastleLabel(Component).Caption := Translate(Component.Name)
  else
  if Component is TCastleButton then
    TCastleButton(Component).Caption := Translate(Component.Name)
  else
    ...
  ;
end;

procedure TMyLocalization.ControlRename(Sender: TObject);
var
  Component: TComponent;
begin
  Component := Sender as TComponent;
  TranslateComponent(Component);
  { Don't worry about removing Component from the list,
    the freed components are automatically removed. }
  if FKnownComponents.IndexOf(Component) = -1 then
    FKnownComponents.Add(Component);
end;

procedure TMyLocalization.SetLanguage(const NewLanguage: string);
var
  I: Integer;
begin
  if Language <> NewLanguage then
  begin
    Language := NewLanguage;
    for I := 0 to FKnownComponents.Count - 1 do
      TranslateComponent(FKnownComponents[I]);
  end;      
end;
```

And in your application, just assign `Name` to each control, and use it as translation id in your translation format.

This way the localization system can be implemented as an independent unit. You need to carry the burden of that ugly sequence `if Control is TCastleLabel then`, but in exchange the core engine remains independent from the translation system.

michaliskambi:
Alternatively, your proposed `Localisation.Add(TUIControl, LocalisationID)` seems even simpler -- this avoids the need to even add OnUIControlRename to the engine.

michaliskambi:
> A question: How is it that you aggree to the fact that the textual elements all have to handle their text themself, because they have to handle it differently, but think that the translation of these texts, in fact a oneliner plus extra bits, shouldn't be handled by them if needed?

I do not want to modify the CastleControls, because there is no need for this modification when using the existing approach of GetText (PO file format, and translation using resourcestrings). It's as simple as that.

I simply consider the GetText system good. GetText is a general system (translates everything with the same approach -- whether it's a text in X3D node, or text in TUIControl, or text inside a dialog message). It is used by Lazarus, and Lazarus applications. Numerous other projects in other languages use this system too -- I know PHP frameworks are integrated with it, and most Linux software is using it for localization, like GNOME or KDE or GNU command-line utils. Translators know it.

And Lazarus shows that this translation system can work comfortably, when coupled with a visual editor and lfm. As I want to add visual editor for CGE, using the same streaming format as lfm, so I know that "GetText approach" will cooperate nicely with it.

So I come from the point of view where I simply want to use GetText with CGE applications. Both PO format, and the resourcestring approach. And in general, I want to do what Lazarus does, because they have already tested this approach on a large scale.

So, while I do want to allow using other translation systems with CGE (custom translation format and/or auto-translating TUIControls), I prefer to keep them independent from CGE code. I prefer to *not* maintain methods like `UpdateLocalisedText` in UI controls in CGE, when I do not see the need for them. Just like Lazarus has many visual controls, and large applications are created using it (including Lazarus itself), but they do not need such translation methods in their UI controls.

> On the other side, I can't see any downsides in having to change CGE core besides the fact itself that CGE core has to be changed.

It has to be changed, and maintained. The change potentially affects all current and future UI controls dealing with text. It's not a large change, but it is something that needs to be maintained, across various UI control classes.

That's why I prefer to put this logic inside the translation system (even if it means an ugly sequence of "ifs" like "if Control is TCastleLabel"), and outside of the engine core.

I know that it seems uncomfortable from your point of view, if you want to implement a translation system with a different approach. But I hope you can see my point of view :) For me, GetText (along with resourcestrings) is a good system, and we have a large confirmation that it's good for others too (Lazarus).

BenediktMagnus:
Okay, thank you. I still disagree, but at least I can now understand why you think so.

To be clear: If the localisation unit works without any changes to the CGE core units, will you integrate it with this pull request or will it never become a part of CGE either way?

michaliskambi:
> To be clear: If the localisation unit works without any changes to the CGE core units, will you integrate it with this pull request or will it never become a part of CGE either way?

I'm fine with having your system in CGE as an alternative localization system. I understand that you find your approach better, and I certainly may be wrong about whether GetText is better or not. Let's enable both approaches in CGE.

The one thing I ask is to separate the "common" localization code, that is useful for all approaches (GetText approach, your custom approach) from the code related to only your approach. That is, I think I would split your `CastleLocalisation` into 2 units:

1. `CastleLocalization` that exposes the "user preferred language" (from https://www.freepascal.org/docs-html/fcl/gettext/getlanguageids.html on Windows and Linux, from the JNI on Android). Maybe it should be just a trivial function `function PreferredLanguage: string`. The goal is to be able to plug it easily at this line in GetText demo : https://github.com/castle-engine/castle-engine/blob/master/examples/localization/game.pas#L189 , *and* to make it available also for your custom system. 

2. And a separate unit with your translation approach. I'm not sure about the name, `CastleLocalizationXML` ? `CastleCustomLocalization`? I think the latter is better? And there you would add a class singleton implementing your system, with 

    - a writeable property like `Language` that is only initialized with `PreferredLanguage`, 
    - routines to load the XML translation format, 
    - and that allows to translate TUIControls automatically.

I believe that this makes both our approaches happy ? :)

BenediktMagnus:
> I'm fine with having your system in CGE as an alternative localization system. I understand that you find your approach better, and I certainly may be wrong about whether GetText is better or not. Let's enable both approaches in CGE.

Great! Then maybe I have an idea with class helpers that could fulfil all conditions.


Splitting the unit in two will be necessary, that's right. I had to refactor the unit anayway to allow the usage of multiple file formats. How about `CastleSystemLanguage` (or `CastleLanguage`) and `CastleLocalization`?


> I believe that this makes both our approaches happy ? :)

I believe that, too. :)


> P.S. You're also welcome to submit a PR to add documentation about your system.

Uh, I fear I am very bad at documentation (mostly I try to write self-explaining code). But I already thought this could be useful (if not appropriate) after CastleClientServer was finished.

michaliskambi:
Unit names `CastleSystemLanguage` (or `CastleLanguage`) are OK for me too.

But I would still prefer to add some part to `CastleLocalization` unit name, like `CastleCustomLocalization`, to make it clear that it's a "custom" system, and when using the GetText approach -- this unit is not applicable.

BenediktMagnus:
So, localisation without touching the engine core is now working via an interface (and TComponent as base class) and an AddOrSet method in the localisation class. Support for the engine core controls (label, button, menu button and edit) is given by child classes ("localizable ones") that can be used instead of the base controls. (Isn't perfect, but I am waaay more comfortable with this than with an `if Control is` approach.) Is this okay for you?
To Dos left: Refactoring of LoadLanguage for the support of multiple file formats (and custom ones), splitting the unit into two (CastleSystemLanguage and CastleCustomLocalization) as well as some other minor things.
As soon as I am ready with that, I will describe all of it much further. You can, of course, have a look at it, but very much of it will certainly change.

src/base/castlelocalisation.pas
```diff
@@ -0,0 +1,223 @@
...
+var
+  IsNewEntry: Boolean;
+begin
+  if ALocalisationID = '' then
+    Exit;
+
+  IsNewEntry := true;
+  try
+    FLocalisationIDList.Add(@ALocalisationComponent.OnUpdateLocalisation, ALocalisationID);
+  except
```
michaliskambi: If possible, do not react to all possible exceptions.

1. better to catch only the exception classes that mean "the item was already existing". Looking at Generics.Collections sources, it seems you want to capture `EListError`.

2. Alternative is to do it like:

```
IsNewEntry := FLocalisationIDList.ContainsKey(@ALocalisationComponent.OnUpdateLocalisation);
FLocalisationIDList.AddOrSetValue(@ALocalisationComponent.OnUpdateLocalisation, ALocalisationID);
```

It's a *bit* slower (you're doing a `ContainsKey` check that will be essentially repeated inside `AddOrSetValue`), but it looks simple.

Both options 1. and 2. are OK by me:)

michaliskambi:
Yes, the approach with interface (ICastleLocalisation) is very fine by me. You're right, it's much better than a sequence of "ifs".

Remember to spell it with "z" (ICastleLocalization) :)

BenediktMagnus:
Ah, thank you! It's `EListError`. I had searched for the exact error and wasn't sure if I had found the right one.

_facepalm_ You are right, my fault. I looked ad `ContainsValue` (instead of `ContainsKey`) and was very confused why it was an O(n) operation... that's the reason for the exception workaround... embarrasing. I will change it to the `ContainsKey` version.

And yes, I will change every occurance of `Localisation` to `Localization` eventually. It's only a full-text replace in the end (no matter if I change it now or later).

BenediktMagnus:
Okay, I _should_ be ready so far. If I have forgotten anything, please let me know.

So, a lot has changed.

1. Objects that shall be automatically localised have to implement the `ICastleLocalization`. For now it implies to be a descendent of `TComponent`. Maybe in the future there could be an alternative interface for custom objects not inherited from `TComponent/TUIControl`.
In `CastleLocalization_CastleCore.inc` there are classes that implement this `ICastleLocalization` and inherits from the standard Castle controls (label, button, menu button and edit). They localise themselfes automatically if used.

2. An object that shall be localised is added by `TCastleLocalization.AddOrSet` with a localisation ID. The same method is used to update the ID. Freeing the objects removes them automatically from the list.

3. System language (especially the Android part) is in it's own unit now (`CastleSystemLanguage`) for a possible use case without `CastleLocalization`. It allows to read the system language in two ways: As simple language code (`en`, `de`, `pl`) or with local info (`en_US`, `en_GB`, `pl_PL`). `TCastleLocalization` encapsulates these two functions, so one doesn't have to include two units.

4. File loader: In `CastleLocalization_FileLoader.inc` there are different loading procedures for different file formats, for now `xml`, `json`, `csv` and `mo` (yes, even `mo` ^^). New file formats can easily be added by simply writing a procedure that loads from a stream and writes to a dictionary. Besides adding these procedures in the engine itself, custom loader can be added from outside the engine via `Localization.FileLoader.Add(extension, procedure)`. With this it is very easy to use existing or custom localisation formats with the engine localisation system, or to change the format later.
I am not very happy with it being in an inc file. It should get it's own unit, so the interface part wouldn't be exposed to users of `CastleLocalization`, but before this is possible, the circular reference has to be solved. I think I will do this later, but still in this pull request.

5. `TGenericCollection`: Used in the JSON file reader, there is a generic collection. Actually, it should be included in the Generics.Collections unit. Ironically, if you look at it's name, it isn't. So putting it in an inc file is only a temporary solution. Do you know where to make a pull request for an integration into the FreePascal Generics.Collections unit? Is https://github.com/maciej-izak/generics.collections still the responsible repository or would I have to go to FreePascal itself? Or would another place in the engine fit better?

Okay, that's it. What do you say? Does this fulfil all your requirements?

Please let the pull request open, even if you are okay with it's current state, because of the above mentioned things I still want to change.

michaliskambi:
I looked briefly now, and everything looks good!

Notes:

1. Kudos for supporting MO files, using a descendant of TMOFile. This is very nice.

2. Note that Delphi doesn't support CORBA interfaces (a pity...). And I want to eventually make CGE compatible with Delphi too. Anyway, it should not really matter, descending from TComponent makes them non-reference-counted anyway (although in a hacky way, with COM interfaces in Delphi). So, this is just FYI, it probably doesn't matter to you :) I will just at some point replace `{$interfaces corba}` with `{$ifdef FPC} {$interfaces corba} {$endif}`.

3. As I mentioned in https://github.com/castle-engine/castle-engine/pull/92#issuecomment-379855539 , I would prefer the unit name `CastleCustomLocalization`, not just `CastleLocalization`. I somehow want to make it clear that it's a "custom" system, and when using the standard GetText approach -- this unit is not needed.

    Also, please document it at the top of the unit (at `unit CastleLocalization;`), like:

    ```
    Note that this is not the only way to localize your Castle Game Engine games.
    You can as well use resourcestrings and standard FPC GetText unit directly,
    as shown in the example code in "examples/localization".
    ```

    I don't want to be too stubborn about this, and I'm unsure does the additional word "Custom" in unit name clearly express what I want anyway ("that it's only 1 possibility to get localization"). So, I guess, feel free to ignore this, and keep `CastleLocalization`... or propose a better name, that somehow concisely conveys what I have in mind :)

    But in any case, please document that it's just one possibility for localization, like my text above.

4. The hardcoded constant "5" below is suspicious :)

    >  if Length(Result) > 5 then //Because on Unix, for example, the encoding is put behind the local info.
    >    Result := Copy(Result, 1, 5);

    I don't know the format of the value expected here, but maybe some additional check (e.g. that Result[6] is a semicolon) would improve this? Or maybe you want to split `Result`, e.g. using `NextTokenOnce(Result, 1, [';'])` ?

As for FPC:

1. They use SVN instead of GIT, and they do not accept pull requests (which caused some heated discussions on FPC mailing lists...). The FPC bugtracker is on https://bugs.freepascal.org/ . You should submit patches generated by `svn diff` vs SVN trunk. Although other options (e.g. pointing to a GitHub commit, see below) are also acceptable. The instructions how to get FPC source code from SVN  are on https://www.freepascal.org/develop.var .

2. There is an unofficial mirror of FPC code on GitHub: https://github.com/graemeg/freepascal . You can use this, if you like working with GIT and GitHub more than SVN. But in the end, you should generate a patch and submit it to https://bugs.freepascal.org/ anyway. Or you can link the FPC bugreport to your changes on GitHub, from what I understand it is OK for FPC developers too (it's not like they hate GitHub, they just chose to not use it for FPC repository).

    In any case, do not submit pull requests versus https://github.com/graemeg/freepascal , as this repository is not maintained by FPC developers.

3. I think that the pull requests versus https://github.com/maciej-izak/generics.collections are OK, although this is somewhat "backdoor" way to insert your modifications to FPC :) The "official" version of `Generics.Collections` is kept inside FPC source code (in SVN repository mentioned in 1., and mirrored in 2.). The https://github.com/maciej-izak/generics.collections is a repository of Maciej Izak, the original developer of Generics.Collections implementation of FPC, and FPC developer. He's a cool person and he may be happy to apply your modifications to his repository, and then push them to FPC repository. But I cannot say for sure -- it would be best to ask him. He may prefer that you submit the modifications to the FPC bugtracker, so that every FPC developer can review them.

michaliskambi:
Hm, actually, one more note about `Generics.Collections`:

Note that this unit provides an API compatible mostly with Delphi, http://docwiki.embarcadero.com/Libraries/Tokyo/en/System.Generics.Collections . FPC developers probably don't mind introducing extensions above Delphi... But it will be a problem for Castle Game Engine. When compiling Castle Game Engine with Delphi, we will use Delphi `Generics.Collections`.

So, I would prefer to keep an utility like TGenericCollection inside Castle Game Engine. Even if it seems generally-useful. Unless you can add it to both FPC *and* Delphi standard library, we have to keep our own implementation anyway.

If you would prefer to move it outside of your CastleLocalization unit, you can add it to CastleClassUtils, which serves as our "general bag for class-related utilities".

michaliskambi:
Note that I have edited my earlier comment, and I deleted my rambling about how I _"prefer `CastleCustomLocalization`, but actually `CastleLocalization` is also OK"_.

If you think that `CastleLocalization` is a good name -- then it is :) As long as it's clearly documented that one does *not* need to use `CastleLocalization` for localization, that's it's just one option -- I'm satisfied :)

BenediktMagnus:
Thank you!

1. Thank you. The integration of MO files into the localisation system combines the best from both GetText and `CastleLocalization`.

2. Oh, right. I added `{$ifdef FPC}`, so it's done.

3. I have added the note. :)

4. The estimated result should look like `en_GB` or `en_US`, two letters for the langauge code, the underline and two more letters for the local info. On some systems there can be more information behind these two. On Debian (and descendents) for example, it returns the name of the language pack, which has for the UTF8 version ".UTF-8" as suffix. So cutting everything behind the the fifth char is safe for whatever the system may put behind the local info.
I changed it to Delete and extended the comment, hopefully this is more understanable now.

FPC: Thanks for the explanation! Because of the Delphi thing (if I open an issue there with the code, it will need years until it is available... and then only for new Delphi versions, so...), it's surely the best to put it into `CastleClassUtils` (exactly the unit this class needs!) and maybe write Maciej Izak about it.

About the name: I am very happy you are okay with `CastleLocalization`! Surely we can document the optional usage in the unit and the documentation.
Because I have thought a lot about a good name after you suggested `CastleCustomLocalization`. I thought that "Custom" doesn't fit so well because it connotes one had to customise it by himself, and in fact the opposite is the case: `CastleLocalization` does everything for the user, without the need for further customisation or adjustments (as with GetText). I had a lot of possible names in mind, like `CastleExtendableLocalisation`, `CastleSimpleLocalisation`, `CastleDataLocalisation`, `CastleStringLocalisation`, `CastleInternalLocalisation`, `CastleAlternativeLocalisation`, `CastleCustomisedLocalisation`... but they all weren't good and didn't get the point. `CastleLocalization` is, in it's simplicity, the best description for the unit. So, I am thankful that you are open for it and I will make it clear in the documentation (that will come) that the unit is fully optional and needn't be used for GetText.

BenediktMagnus:
Okay, I have done everything that was left to do. If I forgot anything, please let me know.

Next to the smaller things, these are the major ones:

1. Example for the basic usage of CastleLocalization is placed under `localization/custom`. It shows how to use localisation with the basic components like `TCastleLabel` and `TCastleButton`, how to exemplary localise images and how to use all the standard file formats.

2. There is now a much more lightweight JSON format with direct key value pairs instead of the old "XML like" object based format. It is still available as `.jsonobj` file extension because it could be useful under some circumstances. (If one wanted to use the object based format with `.json` as file extension, he only needed to set it with `Localization.FileLoader['.json'] := Localization.FileLoader['.jsonobj'];` before loading the language file.)

3. Next to `ICastleLocalization` one can now use `ICastleLocalizationCustom` for custom localisable classes without the need to inherited from TComponent. This allows the creation of lightweight classes that handle adding and removing from the localisation system by themselves.

Additionally, I thought about adding comfort functions like `Create(AOwner: TComponent; ALocalizedText: String)` and `property LocalizationID: String` to the localisable classes of the four standard CGE components, but the redundant, badly maintainable and inconsistent code isn't worth the small comfort gain.

If you have no left annotations, I would be really happy about a merge into master. :)

BenediktMagnus:
Thank you!

I have corrected everything.

- Double semicolon: Hilarious that you found this! ^^

- Project template: I had copied the `tcp_connection` example, which explains the `cthreads` unit. I don't know where I got the base for the `tcp_connection` example from, either from another example or my current project, which could explain the `Window.Free`. I removed both from the `localization` example.

- Locale: Ohh, the ending e in English words... great you saw that.

michaliskambi:
Fantastic! Merged, thank you very much. Thanks for a ton of work making this happen. It will be announced this week :)

BenediktMagnus:
Thank you for merging and announcing! :) It was a lot of fun to make this.

PR #94, opened 2018-04-09
Freetype library.

BenediktMagnus:
Contains the compiled Android library, the Android service and a small patch for TTextureFontData to accept downloadable URLs (including Android assets).


src/fonts/castletexturefontdata.pas
```diff
@@ -313,9 +315,22 @@ constructor TTextureFontData.Create(const URL: string;
...
   FileName := URIToFilenameSafe(URL);
   if FileName = '' then
-    raise Exception.CreateFmt('Cannot read font from URL "%s". Note that right now only local file URLs are supported', [URL]);
+  begin
+    Cache := Download(URL);
+    try
+      FileName := ApplicationConfig('cache_' + ExtractURIName(URL));
+      StreamSaveToFile(Cache, FileName);
+      FileName := URIDeleteProtocol(FileName);
+      IsCachedFile := true;
```
michaliskambi: Use `URIToFilenameSafe` to convert URL -> filename, not `URIDeleteProtocol`. Sometimes `URIDeleteProtocol` also works, but the `URIToFilenameSafe` is more reliable for this.

A minor suggestion: maybe add another local variable, like `CacheURL` here. In the code above, the `FileName` variable sometimes holds an URL, sometimes a filename --- I usually found it's more obvious to use separate variables for URLs, separate for filenames. So, I would suggest to write:

```
CacheURL := ApplicationConfig('cache_' + ExtractURIName(URL));
StreamSaveToFile(Cache, CacheURL);
FileName := URIToFilenameSafe(CacheURL);
```

src/fonts/castletexturefontdata.pas
```diff
@@ -313,9 +315,22 @@ constructor TTextureFontData.Create(const URL: string;
...
+    finally
+      Cache.Free;
+    end;
+  end;
   FontId := FontMgr.RequestFont(FileName);
 
+  if IsCachedFile then
+    DeleteFile(FileName);
+
```
michaliskambi: Use `CheckDeleteFile(FileName, true)` instead `DeleteFile(FileName)`.

The `CheckDeleteFile` is a trivial utility in CastleFilesUtils, it calls `DeleteFile` and checks the result, and makes a warning if it failed. So a failure in deletion will not cause the program crash -- but it will be visible in log file, and may help spot eventual errors.

BenediktMagnus:
I made the changes, except `URIToFilenameSafe` because it returns an empty string for the specific URLs. That's how we detect if it is a non local file:

`FileName := URIToFilenameSafe(URL);`
`if FileName = '' then`
`we do downloading and caching here`

Or did I misunderstand you?

michaliskambi:
> I made the changes, except URIToFilenameSafe because it returns an empty string for the specific URLs.

My suggestion about URIToFilenameSafe was about this line:

```
FileName := URIDeleteProtocol(CacheURL);
```

Changing it to 

```
FileName := URIToFilenameSafe(CacheURL);
```

should work and be better. `CacheURL` is `file://` protocol on Android.

(And if `CacheURL` would not be `file://`, then using `URIDeleteProtocol(CacheURL)` would not return something useful either.)

michaliskambi:
Thank you!

As my only remaining comment ( https://github.com/castle-engine/castle-engine/pull/94#issuecomment-380206120 ) was just a request for 1-line change, I went ahead and applied your PR and did the change myself ( https://github.com/castle-engine/castle-engine/commit/499da6983ffee8805f7dbaad610b5484f54b485d ). Please test whether it works OK -- I did not yet check whether it runs on actual Android device, I only checked compilation now.

BenediktMagnus:
It makes totally sense for CacheURL, you are right.
Great you changed it quickly by yourself.

I will test it soon.

michaliskambi:
I updated the https://github.com/castle-engine/castle-engine/wiki/OpenGL-ES,-Android-and-iOS-TODOs and https://github.com/castle-engine/castle-engine/wiki/Android-Project-Services-Integrated-with-Castle-Game-Engine#freetype to mention the "freetype" service on Android :)

BenediktMagnus:
It works perfectly. :)

PR #95, opened 2018-04-09
Removed CastleUIState and CastleDialogStates from alternate_castle_window.

BenediktMagnus:
In ce1b411b882c5660237051aa66162ad85706998c CastleDialogStates and CastleUIState were moved to ui/opengl into the castle_base.lpk. You forgot to remove if from alternative_castle_window_based_on_lcl.lpk.


michaliskambi:
Thank you!

PR #97, opened 2018-04-14
Add Russian and Ukrainian to Localization example

eugeneloza:


examples/localization/fonts/mk_fonts.sh
```diff
@@ -1,13 +1,13 @@
...
-# Include all the special Polish and German chars to include in a font.
-SAMPLE_TEXT='ŚĆĘĄŹŁŻÓŃśćęąźłżóń'
+# Include all the special Polish, German, Russian and Ukrainian chars to include in a font.
+SAMPLE_TEXT='ŚĆĘĄŹŁŻÓŃśćęąźłżóńйцукенгшщзхъфывапролджэячсмитьбюёЙЦУКЕНГШЩЗХЪФЫВАПРОЛДЖЭЯЧСМИТЬБЮЁїієЇІЄ'
 
 # Note that we create font with a large size (50),
 # this makes it look good in the 3D view when the letters are huge.
 
-texture-font-to-pascal DejaVuSans.ttf \
+./texture-font-to-pascal DejaVuSans.ttf \
```
michaliskambi: This line should be as before, 

```
texture-font-to-pascal ...
```

and not

```
./texture-font-to-pascal ...
```

We should not assume that the texture-font-to-pascal executable is present in the current directory (none of our scripts puts it there). We have to assume it is present on the PATH environment variable.

eugeneloza: Oh, ok. It didn't work for me this way (yeah, I didn't set PATH after reinstalling Debian) :) I'll fix it right away.

michaliskambi:
Thank you! :)

PR #99, opened 2018-04-15
Minor improvements to Localization test

eugeneloza:
To support for German special characters and make it compile cleaner.


michaliskambi:
Thank you! :)

PR #100, opened 2018-04-19
UTF8 keypress

eugeneloza:
See comments in https://github.com/eugeneloza/castle-engine/issues/54#issuecomment-382845047
In short:
Full support of UTF8 in GTK and WinAPI.
Compiles for CastleComponents and XLIB, but only with ASCII keypress support.

### DIFF
--- a/src/images/opengl/castleglimages_sprite.inc
+++ b/src/images/opengl/castleglimages_sprite.inc
@@ -513,7 +513,6 @@ begin
   FImage.Color := AValue;
 end;
 
-
 constructor TSprite.CreateFrameSize(const AImage: TGLImage;
   const AFrames, AColumns, AFrameWidth, AFrameHeight: Cardinal;
   const ATimeLoop: Boolean = True; const APlay: Boolean = False);
--- a/src/ui/castlekeysmouse.pas
+++ b/src/ui/castlekeysmouse.pas
@@ -672,11 +672,12 @@ type
       keyboards by pressing shift + zero, does not have any TKey value.
       So it will generate event with Key = keyNone, but KeyString = '('.
 
-      @bold(Likewise, not all key presses can be represented as char value.)
-      For example "up arrow" (Key = keyUp) doesn't have a char code
-      (it will have KeyString = '').
+      @bold(Likewise, not all key presses can be represented as UTF8 char or
+      simple char.) For example "up arrow" (Key = keyUp) doesn't have a char code
+      (it will have KeyString = '' and KeyCharacter = #0).
 
-      KeyString is an UTF8 symbol and is influenced by some other keys state,
+      KeyString is a string (encoded using UTF-8, like all strings
+      in Castle Game Engine) and is influenced by some other keys state,
       like Shift or Ctrl or CapsLock or some key to input localized characters
       (all dependent on your system settings, we don't deal with it in our engine,
       we merely take what system gives us). For example, you can get "a" or "A"
@@ -771,9 +772,10 @@ type
 
     { Textual description of this event. }
     function ToString: string;
-    { Character corresponding to EventType = itKey
-      Returns #0 if the event was not a keyboard event or keypress event
-      wasn't a simple keyboard event (e.g. a Cyrillic or Arabic character) }
+    { Character corresponding to EventType = itKey.
+      Returns #0 if the event was not a keyboard event or it cannot be
+      represented as a simple 8-bit character (e.g. it is a Cyrillic or Arabic
+      character, or it is a special key like "up arrow"). }
     function KeyCharacter: char;
     { @deprecated Deprecated name for ToString. }
     function Description: string; deprecated;
@@ -788,8 +790,8 @@ type
 
 { Construct TInputPressRelease corresponding to given event.
   @groupBegin }
-function InputKey(const Position: TVector2;
-  const Key: TKey; const KeyString: string; const ModifiersDown: TModifierKeys): TInputPressRelease;
+function InputKey(const Position: TVector2; const Key: TKey;
+  const KeyString: string; const ModifiersDown: TModifierKeys = []): TInputPressRelease;
 function InputMouseButton(const Position: TVector2;
   const MouseButton: TMouseButton; const FingerIndex: TFingerIndex): TInputPressRelease;
 function InputMouseWheel(const Position: TVector2;
@@ -1340,8 +1342,8 @@ begin
   Result := ToString;
 end;
 
-function InputKey(const Position: TVector2;
-  const Key: TKey; const KeyString: string; const ModifiersDown: TModifierKeys): TInputPressRelease;
+function InputKey(const Position: TVector2; const Key: TKey;
+  const KeyString: string; const ModifiersDown: TModifierKeys = []): TInputPressRelease;
 begin
   FillChar(Result, SizeOf(Result), 0);
   Result.Position := Position;
--- a/src/window/windows/castlewindow_winapi.inc
+++ b/src/window/windows/castlewindow_winapi.inc
@@ -381,15 +381,6 @@ function TCastleWindowCustom.WindowProc(uMsg: UINT; wParm: WPARAM; lParm: LPARAM
       MultiByteToWideChar(CP_ACP, MB_PRECOMPOSED, @CharMsg.wParam, SizeOf(CharMsg.wParam), @KeyChar, SizeOf(KeyChar));
       { Now, hopefully convert WideChar to UTF8Char }
       KeyString := UTF8Encode(KeyChar);
-      { This is a bugfix with Ukrainian "i" letter which always messes up encodings :)
-        I just replace it by English "i" here,
-        however, I'm really if this is a good idea,
-        as this will make the words spell wrong (English letter instead of Ukrainian)
-        and will mess up with automatic word processing, e.g. spellchecking.
-        It seems that the character actually gets converted correctly, but
-        Windows functions (e.g. output to console window) don't understand it }
-      {if CharMsg.wParam = 179 then
-        KeyString := 'i';}
     end
     else
       KeyString := '';


michaliskambi:
I'll take a detailed look on the weekend, but everything looks nice on first glance :)

Ignore my earlier comment from https://github.com/eugeneloza/castle-engine/issues/54#issuecomment-382845047 , I wrote it before seeing your `TInputPressRelease.KeyCharacter`. What you did is perfect.

src/ui/castlekeysmouse.pas
```diff
@@ -1300,19 +1304,41 @@ function TInputPressRelease.ToString: string;
...
 end;
 
+function TInputPressRelease.KeyCharacter: char;
+begin
+  {$ifdef MSWINDOWS}
+  { It seems that GTK 1.3 for Windows cannot translate GDK_KEY_Escape and
+    GDK_KEY_Return to standard chars (#13 and #27). So I'm fixing it here. }
+  if Key = K_Escape then
+  Result := CharEscape else
+  if Key = K_Enter then
```
michaliskambi: Indent the lines with `Result` here, like

```
if Key = K_Escape then
  Result := CharEscape else
if Key = K_Enter then
  Result := CharEnter else
```

Better yet, place `else` on separate line:

```
if Key = K_Escape then
  Result := CharEscape
else
if Key = K_Enter then
  Result := CharEnter
else
```


See https://github.com/castle-engine/castle-engine/blob/master/doc/miscellaneous_notes/coding_conventions.txt .

src/ui/castlekeysmouse.pas
```diff
@@ -1300,19 +1304,41 @@ function TInputPressRelease.ToString: string;
...
+  { It seems that GTK 2 doesn't translate backspace and tab to
+    appropriate chars. So I'm fixing it here. }
+  if Key = K_Tab then
+  Result := CharTab else
+  if Key = K_BackSpace then
+  Result := CharBackSpace else
+
+  if Length(Self.KeyString) = 1 then
+   Result := Self.KeyString[1] else
+   Result := #0;
```
michaliskambi: Use just `KeyString`, instead of `Self.KeyString`. Usually qualifying things with `Self.` is not needed, and it does not seem needed here?

michaliskambi:
Looking at my use-cases, I see that my idea for simple `TInputPressRelease.KeyCharacter` has one flaw: we cannot support KeyCharacter equal to Ctrl* constants, like `CtrlA`, `CtrlB`... So a condition `Event.IsKey(CtrlC)` in user code will no longer work.

I propose adding `ModifiersDown: TModifierKeys` to `TInputPressRelease` record, this is known by containers like CastleWindow (just get `ModifiersDown(Pressed)`) so it can be set. The `ModifiersDown` will be useful by itself anyway.

And then the `TInputPressRelease.KeyCharacter` method may have additional lines:

```
if (Result = #0) and 
   (mkCtrl in ModifiersDown) and 
   (Key >= keyA) and 
   (Key <= keyZ) then
  Result := Chr(Ord(Key) - Ord(keyA) + Ord(CtrlA));
```

michaliskambi:
P.S. I forgot to mention it (and I did not announce it yet): The constants `K_Xxx` can be expressed as `keyXxx` since some time. And the latter form is preferred now (as it looks more like normal Pascal enumerated value). At some point, I will even deprecate the `K_Xxx` constants, but for some time they will remain undeprecated, but still we will advice to use `keyXxx`.

src/ui/castlekeysmouse.pas
eugeneloza: Oops. I've got it wrong somehow. I've thought it's done. I'll fix it right away.

src/ui/castlekeysmouse.pas
eugeneloza: Yes, I've made it just to "make clear" (after switching paradigm from GTK procedure :)). I'll fix it.

eugeneloza:
> ModifiersDown: TModifierKeys

I'll try to do this, but this'll take a bit more time (just woke up and still can't put things together clearly)...

> keyXxx

Yes, I've been planning to make some refactoring throughout the Engine (including converting all _k_Xxx_ to _keyXxx_), but make it as  a separate pull request to make changes to UTF8 keystring easier to see as a commit sum-up.
This includes more order in if/then/else (they're not always up to current coding conventions).
But all of those will be cosmetic changes, so I preferred to keep the meaningful changes in a separate pull request.

eugeneloza:
> Looking at my use-cases, I see that my idea for simple TInputPressRelease.KeyCharacter has one flaw: we cannot support KeyCharacter equal to Ctrl* constants, like CtrlA, CtrlB... So a condition Event.IsKey(CtrlC) in user code will no longer work.

I'm not really sure what did you mean by this. I didn't change any tiniest bit of the logic behind KeyString to KeyCharacter conversion. I've tested and the code perfectly catches things like:
```
  if Event.KeyCharacter in [CtrlA..CtrlZ] then
    WriteLn(Event.KeyString);
```
I've left `KeyCharacter` unchanged.

> I propose adding ModifiersDown: TModifierKeys 

Done.

michaliskambi:
>  I've tested and the code perfectly catches things like:
>  if Event.KeyCharacter in [CtrlA..CtrlZ] then

Ah, you're right of course, my bad. For some reason I remembered that we construct the keys like `CtrlA` etc. deep inside GTK/WinAPI backends, and now this mechanism will no longer work. But my memory was wrong -- we don't do anything dirty like that, and these keys should indeed continue to work as before :) Sorry about that.

src/images/opengl/castleglimages_sprite.inc
```diff
@@ -498,6 +508,12 @@ begin
   FImage.Rotation := AValue;
 end;
 
+procedure TSprite.SetColor(AValue: TCastleColor);
+begin
+  FImage.Color := AValue;
+end;
+
+
```
michaliskambi: An extra newline -- please remove :)

src/ui/castlekeysmouse.pas
```diff
@@ -661,27 +661,30 @@ function StrToKey(const S: string; const DefaultKey: TKey): TKey;
...
 
       @bold(Not all key presses can be represented as TKey value.)
       For example, pressing '(' (opening parenthesis), which is done on most
       keyboards by pressing shift + zero, does not have any TKey value.
-      So it will generate event with Key = keyNone, but KeyCharacter = '('.
+      So it will generate event with Key = keyNone, but KeyString = '('.
 
       @bold(Likewise, not all key presses can be represented as char value.)
       For example "up arrow" (Key = keyUp) doesn't have a char code
-      (it will have KeyCharacter = #0).
```
michaliskambi: Instead of

```
       @bold(Likewise, not all key presses can be represented as char value.)
```

maybe better this:

```
       @bold(Likewise, not all key presses can be represented as a string or character.)
```

src/ui/castlekeysmouse.pas
```diff
@@ -661,27 +661,30 @@ function StrToKey(const S: string; const DefaultKey: TKey): TKey;
...
+      (it will have KeyString = '').
 
-      KeyCharacter is influenced by some other keys state,
+      KeyString is an UTF8 symbol and is influenced by some other keys state,
       like Shift or Ctrl or CapsLock or some key to input localized characters
```
michaliskambi: Instead of

```
KeyString is an UTF8 symbol
```

better:

```
KeyString is a string (encoded using UTF-8, like all strings in Castle Game Engine)
```

src/ui/castlekeysmouse.pas
```diff
@@ -781,7 +789,7 @@ function StrToKey(const S: string; const DefaultKey: TKey): TKey;
 { Construct TInputPressRelease corresponding to given event.
   @groupBegin }
 function InputKey(const Position: TVector2;
-  const Key: TKey; const KeyCharacter: Char): TInputPressRelease;
+  const Key: TKey; const KeyString: string; const ModifiersDown: TModifierKeys): TInputPressRelease;
 function InputMouseButton(const Position: TVector2;
```
michaliskambi: Better make the parameter `ModifiersDown` have default value, like

```
function InputKey(const Position: TVector2;
  const Key: TKey; const KeyString: string; const ModifiersDown: TModifierKeys = []): TInputPressRelease;
```

This function is public, so it's better to keep backward compatibility here.

src/ui/castlekeysmouse.pas
```diff
@@ -767,6 +771,10 @@ function StrToKey(const S: string; const DefaultKey: TKey): TKey;
 
     { Textual description of this event. }
     function ToString: string;
+    { Character corresponding to EventType = itKey
+      Returns #0 if the event was not a keyboard event or keypress event
+      wasn't a simple keyboard event (e.g. a Cyrillic or Arabic character) }
+    function KeyCharacter: char;
```
michaliskambi: Make sure to end sentences with a dot. It looks better, and it is also used by https://github.com/pasdoc/pasdoc/wiki/AutoAbstractOption for the 1st sentence.

I propose to change

```
{ Character corresponding to EventType = itKey
  Returns #0 if the event was not a keyboard event or keypress event
  wasn't a simple keyboard event (e.g. a Cyrillic or Arabic character) }
```

to this:

```
{ Character corresponding to EventType = itKey.
  Returns #0 if the event was not a keyboard event or it cannot be
  represented as a simple 8-bit character (e.g. it is a Cyrillic or Arabic character,
  or it is a special key like "up arrow"). }
```

src/window/windows/castlewindow_winapi.inc
```diff
@@ -367,9 +368,31 @@ function TCastleWindowCustom.WindowProc(uMsg: UINT; wParm: WPARAM; lParm: LPARAM
...
+      { This is a bugfix with Ukrainian "i" letter which always messes up encodings :)
+        I just replace it by English "i" here,
+        however, I'm really if this is a good idea,
+        as this will make the words spell wrong (English letter instead of Ukrainian)
+        and will mess up with automatic word processing, e.g. spellchecking.
+        It seems that the character actually gets converted correctly, but
+        Windows functions (e.g. output to console window) don't understand it }
+      {if CharMsg.wParam = 179 then
+        KeyString := 'i';}
+    end
```
michaliskambi: This comment (from _"{ This is a bugfix with Ukrainian "i"..."_) can be removed :) As far as I understand, it works correctly in Castle Game Engine?

(And indeed, the Windows console is another beast. It does not expect strings in UTF-8 encoding. This is unrelated to CGE, you simply cannot directly output UTF-8 string on a Windows console. I saw a thread on Lazarus mailing list about it, but I did not follow... in any case, this is outside of CGE. We want strings in UTF-8, just like Lazarus. And our log wants to output UTF-8, because it may go to file, and writing a text file with UTF-8 encoding is fine and desired.)

src/images/opengl/castleglimages_sprite.inc
eugeneloza: Sorry, I noted this one, but thought about fixing it in some "refactoring" commit later (not to spawn a whole commit for just one line deletion) :) Done.

src/ui/castlekeysmouse.pas
eugeneloza: Done.

src/ui/castlekeysmouse.pas
eugeneloza: Done.

src/ui/castlekeysmouse.pas
eugeneloza: Done.
I left explicit `[]` specifications in `TStateDialogInput.ButtonOKClick`, `TStateDialogInput.ButtonCancelClick` in `castledialogstates.pas`, maybe remove those too?

src/ui/castlekeysmouse.pas
eugeneloza: Done.

src/window/windows/castlewindow_winapi.inc
eugeneloza: Yes, I think we can remove this comment, as Ukrainian "i" gets converted to UTF-8 correctly. It's Windows console problem :) Done.

src/ui/castlekeysmouse.pas
eugeneloza: On the second thought, I believe it's worth removing (in case `InputKey` changes into something different in future empty array is "compatible" with everything so it might get a wrong initializing value). Removed those.

PR #101, opened 2018-05-10
Reliable showing/hiding of Android keyboard

BenediktMagnus:
I was able to change the toggle approach for the Android software keyboard with working `showSoftInput` and `hideSoftInput`, which makes it much less buggy in practical use.
Next to your suggestion to get the view via `getActivity().getWindow().getDecorView()`, it was necessary to use `InputMethodManager.SHOW_FORCED`.
Besides that, I had to extend the `stringToBoolean` function because the FreePascal `BoolToStr` used in CastleEdit wasn't compatible (anymore?). It returns `true` as `-1` and `false` as `0`. (When used in "string mode" it returns `True` and `False`, so this should cover most cases.)

### DIFF
--- a/src/ui/opengl/castlecontrols_edit.inc
+++ b/src/ui/opengl/castlecontrols_edit.inc
@@ -381,7 +381,7 @@ begin
   inherited;
 
   if FAutoOnScreenKeyboard then
-    Messaging.Send(['change-keyboard-state', BoolToStr(Value)])
+    Messaging.Send(['change-keyboard-state', TMessaging.BoolToStr(Value)])
 end;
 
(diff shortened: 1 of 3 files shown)


BenediktMagnus:
Ahh, okay, `TMessaging.BoolToStr` makes sense, my fault. I thought it had something to do with the newest changes in caste-engine/master (because I never had the warnings before), but great that the changes in `stringToBoolean` wasn't necessary.
I reverted the commit and changed `BoolToStr` to `TMessaging.BoolToStr` in CastleEdit.

michaliskambi:
Thank you, merged!

PR #110, opened 2019-04-08
Option to place ScrollView ScrollBar on the left side

eugeneloza:
TCastleScrollViewCustom.ScrollBarLeft

### DIFF
--- a/src/ui/opengl/castlecontrols_scrollview.inc
+++ b/src/ui/opengl/castlecontrols_scrollview.inc
@@ -91,11 +91,14 @@
       in @link(TCastleTheme.ScrollBarWidth Theme.ScrollBarWidth). }
     property ScrollBarWidth: Single read FScrollBarWidth write FScrollBarWidth default 0;
     { Position of the scroll bar.
-      false (default) - scroll bar is located to the right of the ScrollView
-      true - scroll bar is located to the left of the ScrollView
-      Pay attention, that additional space is not made for the scroll bar
-      specify ScrollArea.HorizontalAnchorDelta manually,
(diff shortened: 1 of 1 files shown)


src/ui/opengl/castlecontrols_scrollview.inc
```diff
@@ -88,11 +90,18 @@
       Value of 0 means to use global default
       in @link(TCastleTheme.ScrollBarWidth Theme.ScrollBarWidth). }
     property ScrollBarWidth: Single read FScrollBarWidth write FScrollBarWidth default 0;
+    { Position of the scroll bar.
+      false (default) - scroll bar is located to the right of the ScrollView
```
michaliskambi: I minor note is to make formatting of this nicer in the output documentation :)

Right now the text:

```
false (default) - scroll bar is located to the right of the ScrollView
true - scroll bar is located to the left of the ScrollView
Pay attention, that additional space is not made for the scroll bar
...
```

may be glued into one long line in the output. (Because a single line break is equivalent to space, just like in HTML, see notes about "Whitespace" on https://github.com/pasdoc/pasdoc/wiki/WritingDocumentation ). You could break it into multiple paragraphs, but in this case it's even better to use a list, following https://github.com/pasdoc/pasdoc/wiki/ListTags :

```
@definitionList(
  @itemLabel @false (default)
  @item Scroll bar is located to the right of the ScrollView
  @itemLabe @true
  @item Scroll bar is located to the left of the ScrollView
)
```

eugeneloza:
Thanks! Looks like I understood it :) Committed.

PR #113, opened 2019-05-18
Fix curves example

and3md:
Curves example just crash because there is no data file. 
This PR just add sample XML file and little readme how to create them.


michaliskambi:
Cool, thank you!

PR #114, opened 2019-05-26
Fix Pitch multiplied by SoundInfoGain. 

eugeneloza:


eugeneloza:
On the second thought, I've triple checked everything, and it looks like this fix is correct and trivial, so I'll merge it, if you don't mind.

PR #115, opened 2019-05-26
Make AllocatedSource Pitch affected by SoundInfoPitch.

eugeneloza:


eugeneloza:
Sorry, I've just saw that there is a `SoundInfoPitch` variable, which was meant to be instead of `SoundInfoGain`. Now fixed it completely. Merging, if you don't mind.

michaliskambi:
Excellent, thanks for investigating and fixing this! :) ( I was away since Friday morning, I'm back now. )

PR #116, opened 2019-05-26
Physics - Collision Detection (OnCollisionEnter and GetCollidingTransforms)

and3md:
This PR adds the basis of collision detection from the Kraft engine.

Collisions can be checked in two ways:
1. By getting a list of colliding `TRigidBody` objects (`GetCollidingTransforms`). Works like `get_colliding_bodies()` in Godot. 
2. By using `OnCollisionEnter` event in `TRigidBody` (similar to Unity).

I added very simple example (which will be expanded in the future).

**Update:**
- added `OnCollisionExit`,
- removed maximal linear velocity limit,
- more code clean up.
- implemented TRigidBody.Exists.


michaliskambi:
Thousand thanks for doing this, and addressing my nitpicking, and adding even more features! :) Merged.

I added some tiny improvements (see the 2 commits referenced above).

I'll make a news post about it around the next weekend :)

and3md:
I'm glad I could help :)

Now I thinking about expose more `TKraftRigidBody` methods (before I start working on triggers).

PR #117, opened 2019-05-28
Enable use of FontFamily in CastleSettings.

eugeneloza:


src/ui/opengl/castleinternalsettings.pas
```diff
@@ -261,25 +289,24 @@ procedure SettingsLoad(const Container: TUIContainer; const SettingsUrl: String)
...
-        TCustomizedFont(NewDefaultFont).Load(DefaultFontUrl, AllSizesAtLoad, DefaultFontAntiAliased);
-      end else
-      begin
-        DefaultFontLoadSize := E.AttributeCardinalDef('size_at_load', DefaultFontSize);
-        NewDefaultFont := TTextureFont.Create(Container);
-        TTextureFont(NewDefaultFont).Load(DefaultFontUrl, DefaultFontLoadSize, DefaultFontAntiAliased);
-      end;
-      NewDefaultFont.Size := DefaultFontSize;
+      NewFontFamily := TFontFamily.Create(Container);
+      NewFontFamily.Name := 'FontVariants';
```
michaliskambi: Assigning the `NewFontFamily.Name` here is risky, as user may create his/her own TComponent with such name, and set as `Owner` the `Container`. And all components with the same owner must have unique names.

Sure, there's only a small chance that someone will use exactly this name, but in general we want to leave all `Name` (that can affect user) to be configurable by developer.

Hm, or we can clearly mark the `Name` as something that will never be used by developer. Instead of `FontVariants`, I recommend `CastleInternalDefaultFont`. There's no way someone will use it by accident (without knowingly causing clash :) ).

src/ui/opengl/castleinternalsettings.pas
```diff
@@ -261,25 +289,24 @@ procedure SettingsLoad(const Container: TUIContainer; const SettingsUrl: String)
         E.AttributeSingleDef('reference_height', DefaultUIReferenceHeight);
     end;
 
-    E := SettingsDoc.DocumentElement.Child('default_font', false);
-    if E <> nil then
+    NewFonts[ftRegular] := LoadFontSettings(SettingsDoc.DocumentElement.Child('default_font', false));
```
michaliskambi: This creates a structure in XML file that doesn't look perfect, because it's "flat" where it could be "hierarchical":

```
<default_font_regular>
</default_font_regular>
<default_font_bold>
</default_font_bold>
...
```

I would prefer to change to code to support this:

```
<default_font>
  <regular
      url="castle-data:/MyFontFile.ttf"
      size="20"
      sizes_at_load="10 20 30"
      anti_aliased="true"
    />
  <bold
      url="castle-data:/MyFontFile-Bold.ttf"
      size="20"
      sizes_at_load="10 20 30"
      anti_aliased="true"
    />
  <italic
      url="castle-data:/MyFontFile-Italic.ttf"
      size="20"
      sizes_at_load="10 20 30"
      anti_aliased="true"
    />
  <bold_italic
      url="castle-data:/MyFontFile-BoldItalic.ttf"
      size="20"
      sizes_at_load="10 20 30"
      anti_aliased="true"
    />
</default_font>
```

The way to do it:

In `LoadFontSettings`, check not only `if FontElement <> nil then`, check `if (FontElement <> nil) and FontElement.HasAttribute('url') then` .

Look for fonts by first checking is `DefaultFontElement := SettingsDoc.DocumentElement.Child('default_font', false); if DefaultFontElement <> nil then`, and if true, get the children like `DefaultFontElement.Child('regular', false)`.

src/ui/opengl/castleinternalsettings.pas
```diff
@@ -261,25 +289,24 @@ procedure SettingsLoad(const Container: TUIContainer; const SettingsUrl: String)
...
+    if NewFonts[ftRegular] = nil then
+      NewFonts[ftRegular] := LoadFontSettings(SettingsDoc.DocumentElement.Child('default_font_regular', false));
+    NewFonts[ftBold] := LoadFontSettings(SettingsDoc.DocumentElement.Child('default_font_bold', false));
+    NewFonts[ftItalic] := LoadFontSettings(SettingsDoc.DocumentElement.Child('default_font_italic', false));
+    NewFonts[ftBoldItalic] := LoadFontSettings(SettingsDoc.DocumentElement.Child('default_font_bolditalic', false));
+    if (NewFonts[ftBold] = nil) and (NewFonts[ftItalic] = nil) and
+      (NewFonts[ftBoldItalic] = nil) then
+      NewDefaultFont := NewFonts[ftRegular]
```
michaliskambi: I am unsure about this "auto-detection" (to not use TFontFamily when only "regular" is defined).

I think it's better to just always use `TFontFamily` when user defines even only "regular" font.

So, in the new XML syntax I propose in previous comment, it would mean that 

- If you use the old method, `<default_font url="..."/>` then you don't create TFontFamily.

- If you use the new method, `<default_font><regular url="..." /></default_font>` then you always create a TFontFamily (even if bold, italic, bold_italic were not provided).

This corresponds more directlly to what is possible in code: you can have TFontFamily with only TFontFamily.Regular <> nil.

eugeneloza:
Ok, I see. This is a much more optimal approach :) I'll do it.

eugeneloza:
All seems to be fixed. Also added `TCastleUserInterface.AnchorDelta` property.

eugeneloza:
Thanks! I was doubting that one :)

PR #118, opened 2019-06-04
KraftRigidBody lifetime fix and expose some methods (WIP)

and3md:
As you probably noticed `TRigidBody.GetLinearVelocity()` has some problem. There is no 100% working way to check whether there is existing `FKraftBody` or that is just hanging pointer (so I have not added this check). It's because sometimes KraftEngine could be freed before `TCastleTransforms.DeinitializeTransform` call. Like in comment in `DeinitializeTransform`:

>  { FKraftBody is owned by FKraftEngine.
      If FKraftEngine = nil then FKraftBody was already freed.
      This can happen because TSceneManagerWorld.Destroy frees FKraftEngine,
      but the TCastleTransform instances (with TRigidBody)
      may be detached later, see TTestCastleTransform.TestPhysicsWorldOwner.

Because we currently have a pointer to our `TRigidBody` in UserData. I think we can fix it and have 100% certainty that this pointer is OK. I think it will allow to avoid many mistakes/bugs in the future.

Please check it. If you approve it a lot of check will base on it. Now I back to work on my game (in CGE of course!) so you don't need to hurry :)

### DIFF
--- a/src/3d/castletransform_physics.inc
+++ b/src/3d/castletransform_physics.inc
@@ -136,18 +136,12 @@
 
     procedure SetAngularVelocity(const AVelocity: TVector3);
     function GetAngularVelocity: TVector3;
-
     procedure SetAngularVelocityDamp(const AValue: Single);
-    function GetAngularVelocityDamp: Single;
-
     procedure SetMaximalAngularVelocity(const AValue: Single);
 
     procedure SetLinearVelocity(const LVelocity: TVector3);
     function GetLinearVelocity: TVector3;
-
     procedure SetLinearVelocityDamp(const AValue: Single);
-    function GetLinearVelocityDamp: Single;
-
     procedure SetMaximalLinearVelocity(const AValue: Single);
 
     procedure SetExists(const Value: Boolean);
@@ -170,11 +164,11 @@
     function GetCollidingTransforms: TCastleTransformList;
 
     property AngularVelocity: TVector3 read GetAngularVelocity write SetAngularVelocity;
-    property AngularVelocityDamp: Single read GetAngularVelocityDamp write SetAngularVelocityDamp;
+    property AngularVelocityDamp: Single read FAngularVelocityDamp write SetAngularVelocityDamp;
     property MaximalAngularVelocity: Single read FMaximalAngularVelocity write SetMaximalAngularVelocity;
 
     property LinearVelocity: TVector3 read GetLinearVelocity write SetLinearVelocity;
-    property LinearVelocityDamp: Single read GetLinearVelocityDamp write SetLinearVelocityDamp;
+    property LinearVelocityDamp: Single read FLinearVelocityDamp write SetLinearVelocityDamp;
     property MaximalLinearVelocity: Single read FMaximalLinearVelocity write SetMaximalLinearVelocity;
 
     { Occurs when TRigidBody starts colliding with another TRigidBody.
@@ -563,28 +557,12 @@ begin
   end;
 end;
 
-function TRigidBody.GetAngularVelocityDamp: Single;
-begin
-  if FKraftBody <> nil then
-    Result := FKraftBody.AngularVelocityDamp
-  else
-    Result := FAngularVelocityDamp;
-end;
-
 procedure TRigidBody.SetMaximalAngularVelocity(const AValue: Single);
 begin
   FMaximalAngularVelocity := AValue;
   { Kraft uses max velocity for delta time which is physics update frequency. }
   if FKraftBody <> nil then
     FKraftBody.MaximalAngularVelocity := AValue / FTransform.World.FKraftEngine.WorldFrequency;
-end;
-
-function TRigidBody.GetLinearVelocityDamp: Single;
-begin
-  if FKraftBody <> nil then
-    Result := FKraftBody.LinearVelocityDamp
-  else
-    Result := FLinearVelocityDamp;
 end;
 
 procedure TRigidBody.SetMaximalLinearVelocity(const AValue: Single);
@@ -615,9 +593,7 @@ begin
     Exit;
 
   FLinearVelocity := VectorFromKraft(FKraftBody.LinearVelocity);
-  FLinearVelocityDamp := FKraftBody.LinearVelocityDamp;
   FAngularVelocity := VectorFromKraft(FKraftBody.AngularVelocity);
-  FAngularVelocityDamp := FKraftBody.AngularVelocityDamp;
 end;
 
 procedure TRigidBody.Update(const Transform: TCastleTransform; const SecondsPassed: Single);
--- a/src/3d/castletransform_physics.inc
+++ b/src/3d/castletransform_physics.inc
@@ -135,12 +135,10 @@
     procedure SynchronizeFromKraft;
 
     procedure SetAngularVelocity(const AVelocity: TVector3);
-    function GetAngularVelocity: TVector3;
     procedure SetAngularVelocityDamp(const AValue: Single);
     procedure SetMaximalAngularVelocity(const AValue: Single);
 
     procedure SetLinearVelocity(const LVelocity: TVector3);
-    function GetLinearVelocity: TVector3;
     procedure SetLinearVelocityDamp(const AValue: Single);
     procedure SetMaximalLinearVelocity(const AValue: Single);
 
@@ -163,11 +161,11 @@
     { Transformations that we collide with currently. }
     function GetCollidingTransforms: TCastleTransformList;
 
-    property AngularVelocity: TVector3 read GetAngularVelocity write SetAngularVelocity;
+    property AngularVelocity: TVector3 read FAngularVelocity write SetAngularVelocity;
     property AngularVelocityDamp: Single read FAngularVelocityDamp write SetAngularVelocityDamp;
     property MaximalAngularVelocity: Single read FMaximalAngularVelocity write SetMaximalAngularVelocity;
 
-    property LinearVelocity: TVector3 read GetLinearVelocity write SetLinearVelocity;
+    property LinearVelocity: TVector3 read FLinearVelocity write SetLinearVelocity;
     property LinearVelocityDamp: Single read FLinearVelocityDamp write SetLinearVelocityDamp;
     property MaximalLinearVelocity: Single read FMaximalLinearVelocity write SetMaximalLinearVelocity;
 
@@ -529,7 +527,6 @@ begin
 
   Assert(not ((Transform.World.FKraftEngine = nil) and (FKraftBody <> nil)), 'KraftBody should not live longer than KraftEngine!');
 
-  SynchronizeFromKraft;
   FreeAndNil(FKraftBody);
 
   { Collider.FKraftShape is owned by FKraftBody, it was automatically freed already }
@@ -637,8 +634,10 @@ begin
 
   UpdateCollides(Transform);
   if Dynamic then
-    TransformationFromKraft
-  else
+  begin
+    TransformationFromKraft;
+    SynchronizeFromKraft; // I think its only required for dynamic bodies.
+  end else
   if Animated then
   begin
     // TODO: check "if TransformChanged then" or such, don't do this every frame
@@ -677,14 +676,6 @@ begin
   end;
 end;
 
-function TRigidBody.GetLinearVelocity: TVector3;
-begin
-  if FKraftBody <> nil then
-    Result := VectorFromKraft(FKraftBody.LinearVelocity)
-  else
-    Result := FLinearVelocity;
-end;
-
 procedure TRigidBody.SetAngularVelocity(const AVelocity: TVector3);
 begin
   FAngularVelocity := AVelocity;
@@ -694,14 +685,6 @@ begin
     if not AVelocity.IsPerfectlyZero then
       FKraftBody.SetToAwake;
   end;
-end;
-
-function TRigidBody.GetAngularVelocity: TVector3;
-begin
-  if FKraftBody <> nil then
-    Result := VectorFromKraft(FKraftBody.AngularVelocity)
-  else
-    Result := FAngularVelocity;
 end;
 
 procedure TRigidBody.SetExists(const Value: Boolean);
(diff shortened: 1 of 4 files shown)


src/3d/castletransform_physics.inc
```diff
@@ -818,7 +815,24 @@ begin
 end;
 
 destructor TSceneManagerWorld.Destroy;
+var
+  CastleRigidBody: TRigidBody;
+  KraftRigidBody: TKraftRigidBody;
+  NextKraftRigidBody: TKraftRigidBody;
 begin
+  KraftRigidBody := FKraftEngine.RigidBodyFirst;
```
michaliskambi: Destructor may be called even when some exception occurs in the middle of the constructor. So destructor body must be "robust", i.e. prepared to to handle the case when *anything* can be nil. In this case, it means that you should put the whole block of code here under `if FKraftEngine <> nil then ...`.

src/3d/castletransform_physics.inc
```diff
@@ -501,17 +501,14 @@ end;
 
 procedure TRigidBody.DeinitializeTransform(const Transform: TCastleTransform);
 begin
+  if FRecreateKraftInstance then
```
michaliskambi: I'm uncertain of this line, I will need more time to investigate is it 100% correct. I mean, `DeinitializeTransform` is called from various occasions. 

- E.g. when you call `MyTransform.RigidBody := nil` in the middle of the game (when physics sumulation started).
- Or when you call `SceneManager.Items.Remove(MyTransform)` in the middle of the game.

I want to check (read the code & do the tests) what is the value of `FRecreateKraftInstance` in this case (and if it can be `true`, is it correct that we abort the rest of DeinitializeTransform work?)

and3md: I added this check because in some cases deinitialisation was called twice.  I think `ChangeWorld` or `PhysicsChangeWorldDetach` need more checks to fix that. This need more investigation.

and3md:
I was also thinking about this flag `FRecreateKraftInstance` and now it seems redundant. Because of that I could not decide. Should I check the existence of `FKraftBody` or `FRecreateKraftInstance` in methods like `GetLinearVelocity()`? I think it should be removed. I think that the existence of `FKraftBody` should means the correct initialization of physics in TRigidBody.

At the moment I wanted to know do you accept my approach to check the existence of FKraftBody. Is it worth to continue :)

michaliskambi:
👍  Yes, `FKraftBody <> nil` should be enough to test whether the object is correctly initialized with the physics engine, after all `TKraftRigidBody` cannot be instantiated otherwise. After this PR, `FKraftBody` should be always `nil` when it is not initialized, and that's very good :)

Note that in case of `LinearVelocity`, we will also need a private field `FLinearVelocity` on the CGE side. `TRigidBody.SetLinearVelocity` should set both Kraft's linear velocity (if `FKraftBody <> nil`) and CGE `FLinearVelocity` (always). Then `GetLinearVelocity` should update `FLinearVelocity` from Kraft state (if `FKraftBody <> nil`), as it may change due to simulation (as far as I understand) and then return `FLinearVelocity`.

IOW, my point is that doing

```
RigidBody.LinearVelocity := Vector3(1, 2, 3);
Writeln(RigidBody.LinearVelocity.ToString);
```

should write ` 1 2 3`, regardless if the physics engine is initialized or not (e.g. regardless if `SceneManager.EnablePhysics` is true or false).

I would be happy to remove `FRecreateKraftInstance` and replace it with something cleaner (or nothing!), this indeed needs investigation :)

and3md:
You read in my mind! :)

In next step I wanted propose to remove all `Initial...` methods and add `FLinearVelocity` to `LinearVelocity` property.

The only difference was that I wanted to add a flag, which would determine whether during the deinitialization the value `FLinearVelocity` should  be updated, or at each initialization the value should return to the one given by the programmer. Something like `BackToInitialValue` or `SynchronizeOnDeinitialization`. It could also serve as an optimization if we know that after detach we destroy the rigid body, so there is no need to lose time to synchronize this values. But maybe it's not intuitive/too complicated.

What do you think about it?

michaliskambi:
Hm, I would prefer to have both `InitialLinearVelocity` and `LinearVelocity` as separate properties, at least for now. 

Reasons:

1. If we have these two as separate properties, then the approach to "serialization" (what should be "published", what should remain only "public") is easy: `InitialLinearVelocity` can be serialized at any moment, so it can be "published" one day, and serializing `LinearVelocity` is not generally useful so it should remain just "public".

    IOW, `InitialLinearVelocity` guarantees that it has the value you assign to it, and never changes automatically. And this is important for CGE editor ( https://castle-engine.io/manual_editor.php ) where I  would like to "expose" the ability to configure TRigidBody and colliders one day (TRigidBody can be just a subcomponent with some published properties). And in general, it is useful for any code doing (de)serialization of TCastleTransform hierarchy using e.g. [CastleComponentSerialize](https://castle-engine.io/apidoc-unstable/html/CastleComponentSerialize.html) unit.

    If we would merge these into only one property `LinearVelocity`... then everything is still possible (e.g. when serializing, we could always seriallize only the "initial value", not "current value") but it seems less obvious.

2. This way user can also change the "initial" value always, regardless if the simulation is currently  running or not, since the `InitialLinearVelocity` can always be changed. If we would merge them into one field (only `LinearVelocity` visible for user) then internally we would still have two states (something keeps the "initial velocity" vector, something keeps the "current velocity" vector) but user would not be able to change the "initial value" before e.g. setting `TCastleTransform.RogidBody := nil`. 

    Granted, the above is not a usual use-case, so it's not very important... but it points that having two separate properties (`InitialLinearVelocity` and `LinearVelocity`) is more natural. We  would need to "think" about these two values anyway ("initial value" and "current value") so it makes sense to have explicitly two properties representing them.

P.S. I already have this problem with `TCastleTransform.Translation` and `TCastleTransform.Rotation`. They simply change when you run physics, and there is no `InitialTranslation` or `InitialRotation` (and I don't want it, to keep the API simple). So my decisions *are* inconsistent here, I know this :) But the bottom line is that I will need to solve this once I implement designing TRigidBody in CGE editor. In the meantime, having separate `InitialLinearVelocity` and `LinearVelocity` feels safer, i.e. "the way things work is more obvious".

P.P.S. When I speak about "make `InitialLinearVelocity` published" it is a little shortcut. In fact you cannot publish in Object Pascal a property of type TVector3, but we make it work using TCastleVector3Persistent wrapper. So actually the published property would be called `InitialLinearVelocityPersistent`.

michaliskambi:
After 5 minutes of more thinking: Hm, on the other hand, I realized that *not* merging these properties into one `LinearVelocity` also causes problems:

- if physics is not initialized, what happens to `LinearVelocity` when you assign `InitialLinearVelocity`? Does setting `InitialLinearVelocity` change the `LinearVelocity` immediately?

- if physics is not initialized, what happens if you change `LinearVelocity`? It seems that it's a trap -- whatever you set to `LinearVelocity`, will be overridden by `InitialLinearVelocity` when physics starts.

So, **I have just changed my mind** :) Here is the new decision:

Yes, please merge them into one `LinearVelocity`. Assigning  it before FKraftBody is initialized should still work (and in the effect, set the initial value, following my earlier post with paragraph """Note that in case of LinearVelocity, we will also need a private field FLinearVelocity on the CGE side. ...").

For now I would not bother with `BackToInitialValue`. If necessary, user code has do this reset "manually" for now.

This way, `LinearVelocity` willl work the same as current `TCastleTransform.Translation` and `TCastleTransform.Rotation`. When physics work, these values just change, and that's it. That's a simple and consistent behavior for all these properties.

In the future, when I will enable TRigidBody configuration in CGE editor:

- in the simplest initial version, CGE editor will just keep physics disabled, so no problem.

- later I may allow to enable physics, and then I will decide how to restore `LinearVelocity`  as well as `TCastleTransform.Translation/Rotation` in the same, uniform fashion. Maybe I'll introduce a property like `BackToInitialValue` then. Or maybe I will just save + restore the whole world state (serializing it to/from file) when the physics starts + stop (which would be similar to the "Play" in Unity Editor behavior).

src/3d/castletransform_physics.inc
```diff
@@ -467,17 +467,17 @@ procedure TRigidBody.InitializeTransform(const Transform: TCastleTransform);
 
     UpdateCollides(Transform);
 
-    if (not InitialAngularVelocity.IsPerfectlyZero) or
-       (not InitialLinearVelocity.IsPerfectlyZero) then
+    if (not FAngularVelocity.IsPerfectlyZero) or
+       (not FLinearVelocity.IsPerfectlyZero) then
     begin
       { The behaviour is more natural when we zero the InitialXxxVelocity
```
michaliskambi: Update this comment, it says `InitialXxxVelocity`, should be `XxxVelocity`.

src/3d/castletransform_physics.inc
```diff
@@ -613,14 +603,42 @@ end;
 
 procedure TRigidBody.SetLinearVelocity(const LVelocity: TVector3);
 begin
-  FKraftBody.LinearVelocity := VectorToKraft(ZeroLockedComponents(LVelocity, FLockTranslation));
-  if not LVelocity.IsPerfectlyZero then
-    FKraftBody.SetToAwake;
+  if FKraftBody <> nil then
```
michaliskambi: The setter should *always* set CGE private field `FLinearVelocity` (regardless if FKraftBody <> nil).

Likewise the getter (TRigidBody.GetLinearVelocity below) should update the CGE private field `FLinearVelocity` (only if FKraftBody <> nil), and then always return `FLinearVelocity` (regardless if FKraftBody <> nil).

Same note for TRigidBody.SetAngularVelocity, TRigidBody.GetAngularVelocity below.

Reason:

My approach seems more limiting, but is more intuitive, and it is consistent with how TCastleTransform.Translation/Rotation behave. The goal is that these sequences of code should result in the 100% same state:

```
// doing SetLinearVelocity when FKraftBody is nil
RigidBody.LinearVelocity := Vector3(1, 2, 3);
MyTransform.RigidBody := RigidBody;
```

and

```
MyTransform.RigidBody := RigidBody;
// doing SetLinearVelocity when FKraftBody <> nil
RigidBody.LinearVelocity := Vector3(1, 2, 3);
```

These just swap 2 lines, and it is intuitive that they should achieve 100% the same state. A deserialization system (once we made RigidBody and it's properties published) could also execute these 2 lines in any order.

Right now, in your code, there is a subtle difference between these two sequences: if you set `Transform.RigidBody := nil`, then in one case `RigidBody.LinearVelocity` will switch back to zero (since FLinearVelocity was not modified). We want to avoid it. In general, doing `Transform.RigidBody := nil` should keep the last seen state of `RigidBody.LinearVelocity`.

and3md: > Right now, in your code, there is a subtle difference between these two sequences: if you set Transform.RigidBody := nil, then in one case RigidBody.LinearVelocity will switch back to zero (since FLinearVelocity was not modified). We want to avoid it. In general, doing Transform.RigidBody := nil should keep the last seen state of RigidBody.LinearVelocity.

Yes it will work like that (there will be no difference when you set velocity before assigning rigid body or after assigning like in ). I just not finished it yesterday and pushed changes habitually. Sorry about this. I wanted add this values (linear/angular velocity and many others) synchronization code on `DeinitializeTransform`. I think this should work.

I agree with you that setters should update FLinearVelocity, and I will change that. But I have some doubts about getters. In setters we will set that because of serialization (I must remember about that!). Setting `FLinearVelocity` in getter will not guarantee that value will be correct and additionally can lower performance (for example when in user code someone will get linear velocity on each frame).  I do not know how much, but I preferred not to do something that you do not have to do.

I think better solution will be add function which synchronizes all the physics values like linear velocity. And run it before serialization. The same function will be called in `DeinitializeTransform`.
Then we will have certainty that all values are valid. We can name it `SynchronizePhysics` or something like that.  What do you think about that?

src/3d/castletransform_physics.inc
and3md: Fixed in https://github.com/castle-engine/castle-engine/pull/118/commits/27df87cd0b18a824b111fe56843a71cf11738011 :)

src/3d/castletransform_physics.inc
and3md: I added `SynchronizeFromKraft`.  And after that I'm not sure. Maybe it would be better to set private CGE fields in the getter as you said to do not duplicate code and just call getters in `SynchronizeFromKraft`. On the one hand the code will be in one place and not duplicated but on the other hand I feel that the code will not be so optimal when we do that. Please check current implementation and say what do you think about that.

michaliskambi:
My main concern here is consistency. We should make the properties like `LinearVelocity` behave consistently with how `TCastleTransform.Translation` , `TCastleTransform.Rotation` operate (from user's point of view).

And this means that your `TRigidBody.SynchronizeFromKraft` is a cool idea, and I want it, although used in a little different way:

- `SynchronizeFromKraft` should be moved to a local function inside the `TRigidBody.Update`. Exactly like `TransformationFromKraft` is defined already.

- `SynchronizeFromKraft` should be called from `TRigidBody.Update`

- `SynchronizeFromKraft` should not be called from `DeinitializeTransform` (because `TCastleTransform.Translation` , `TCastleTransform.Rotation` also do not do this, and it seems not necessary).

- And this way the getter `GetLinearVelocity` can be just removed, this property should just return the field (`property LinearVelocity read FLinearVelocity write SetLinearVelocity`).

I believe that this makes the behavior 100% consistent with how `TCastleTransform.Translation` , `TCastleTransform.Rotation` is updated (my primary goal), and the implementation code will be quite simple this way (my secondary goal). We also gain a little performance (which is just a nice bonus :) ) -- reading `LinearVelocity` property multiple times has no additional cost, it is updated only once per `TRigidBody.Update`.

and3md:
Finally, I think the code is OK and can be merged. The physics code is now more clear and useful. I think next things should go on next PR :) Please check and merge if everything is OK.

michaliskambi:
Excellent, thank you very much, merged!

I added `InitialAngularVelocity` and `InitialLinearVelocity` as "deprecated" properties (they have the same getters/setters as their new versions without `Initial` prefix), to have backward-compatibility.

PR #120, opened 2019-06-22
Physics: Fix first castle transform position update.

and3md:
Little fix for first Castle Transform position update by `TransformationFromKraft`. After `TRigidBody.InitializeTransform` Kraft shape transform was not set properly. That makes strange effects before first physics step. See code comment for more info. Tested in my game ;)

### DIFF
--- a/examples/physics/physics_2d_collisions/gameinitialize.pas
+++ b/examples/physics/physics_2d_collisions/gameinitialize.pas
@@ -205,7 +205,7 @@ begin
     'FPS: %s' + NL +
     'Scene Manager Objects: %d' + NL +
     'Linear velocity: %f' + NL +
-    'Use AWSD to change plane velocity, space to pause, R to restart plane ' + NL +
+    'Use AWSD to change plane velocity, space to pause, R to restart plane.' + NL +
     NL+
     'Current Plane Colisions (from TRigidBody.GetCollidingTransforms):' + NL +
     '  %s' + NL +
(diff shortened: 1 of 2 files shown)


michaliskambi:
Makes sense. 

But I would like to reproduce the problem to be sure I can always confirm it's still fixed :) I tried adding

```
SceneManager.TimeScale := 0.0;
```

to ApplicationInitialize in examples/physics/physics_3d_demo/gameinitialize.pas , but I do not see a bug. 

Can you describe what I would need to do (in examples/physics/physics_3d_demo/ or other demo) to see the problem?

(BTW, if I exit the physics_3d_demo then, it reports an error -- `EAssertionFailed: Second deinitialization! (castletransform_physics.inc, line 530)`. )

and3md:
I will check physics_3d_demo and try to make a simple example with this bug.

and3md:
In physics_3d_demo when level is destroyed first `TRigidBody` of a box is destroyed next the box (TCastleTransform) itself. Then the FRigidBody pointer is incorrect. So in TRigidBody.Destroy we need set TCastleTransform.FRigidBody to nil.

and3md:
OK, I added space to stop time and R to restart in collision example. Just press space to pause and then R. The position of new plane will be incorrect (without https://github.com/castle-engine/castle-engine/pull/120/commits/e1f6063ddbfbb14026e8c9598d1ae3e5286e930c). When time is on this is hard to see.

michaliskambi:
Thank you for both fixes! Tested and merged :) I extended the comments a bit -- my memory is short, so I need to document everything :)

PR #121, opened 2019-06-24
TTextNode.MaxDisplayChars

eugeneloza:


src/x3d/auto_generated_node_helpers/x3dnodes_text.inc
```diff
@@ -66,14 +74,38 @@ begin
...
+var
+  NewValue: Integer;
+begin
+  NewValue := Value;
+  if NewValue > FTotalChars then
+    NewValue := FTotalChars; //to avoid redrawing text each time if the NewValue is increasing above TotalChars
+  if FMaxDisplayChars <> NewValue then
+  begin
+    FMaxDisplayChars := NewValue;
+    FdString.Changed; //ask the node to update content
```
michaliskambi: It's a bit hack to call `FdString.Changed` now. Since `FdString` contents did not change, and `FdString.Changed` should only be called then.

But I do not have a simple better proposition for now.

One alternative is to remake `maxDisplayChars` as a proper X3D field of type `SFInt32` (that can also be specified inside X3D file). E.g. like `maxExtent` field of TTextNode. Then the `maxDisplayChars` field upon change could cause anything, in particular `FdMaxDisplayChars.ChangesAlways := [chGeometry];` would make it rebuild the text node. But doing it is admittedly more work, so I just note it as an alternative, no need to change anything -- for now, `FdString.Changed` can stay :)

eugeneloza: I've thought about the issue and it seems like making a X3D field is nice and it proved to be much **simpler**, than the code I've had before. However, it doesn't seem like  `TextNode.FdMaxDisplayChars.Value := somevalue;` calls rebuilding of the geometry, only if I use `TextNode.FdMaxDisplayChars.Send(somevalue);` - then it works as expected.
I've had a second thought on the optimization of MaxDisplayCharacters and maybe it's too minor, so that we can leave it on user-side. So I removed it for now.

michaliskambi: Cool!

Indeed `Node.FdXx.Value:=yyy` doesn't call notification. User code should in 90% cases do `Node.Xxx:=yyy` (auto-generated, calls Send(yyy)) or `Node.FdXxx.Send(yyy)` or `Node.FdXxx.Value:=yyy; Node.FdXxx.Changed;`.

That is why I want to eventually make FdXxx internal (private) some day:)

src/x3d/auto_generated_node_helpers/x3dnodes_text.inc
```diff
@@ -66,12 +66,16 @@ begin
   FdMaxExtent.Send(Value);
 end;
 
-procedure TTextNode.SetString(const Value: array of string);
+procedure TTextNode.SetString(const Value: array of String);
+var
```
michaliskambi: Note that you left modifications to auto-generated file here :)

michaliskambi:
Cool. After adding new X3D field, edit the `tools/internal/x3d-nodes-to-pascal/nodes-specification/Text.txt` file and rerun the x3d-nodes-to-pascal program by 
```
cd tools/internal/x3d-nodes-to-pascal/code
./x3d-nodes-to-pascal_compile.sh
./x3d-nodes-to-pascal_run.sh
```

In code you should prefer then to use `MaxDisplayChars` instead of `FdMaxDisplayChars.Value`.

src/x3d/auto_generated_node_helpers/x3dnodes_text.inc
eugeneloza: Yes, just noted this one too :) Fixed.

PR #124, opened 2019-07-11
ASTC texture compression

and3md:
This PR adds ASTC textures compression to CGE. ASTC compression is available on most mobile devices (starting from 2nd generation of Mali-T600 series, Adreno 400, Apple A8). 

It's very flexible:
- block size can be from 4 (best quality) to 12 (smallest size)
- size don't need to be the power of 2

Using ASTC textures speeded up my game launch form 5-7 sec to about 0.5 sec on Helio X20 with Mali-T880 MP4 graphics.

## KTX
When you call `castle-engine auto-generate-textures` all ASTC textures are generated and saved to KTX files. The engine expects KTX files when try to load ASTC textures.

You need `astcenc` in your `PATH` variable if you want generate ASTC textures.

### Extra ETC1 fix
There was a issue with ETC1 textures loading:
```
07-05 12:27:18.948 25967 25984 I my_fantastic_game: Error loading image from URL "castle-data:/auto_generated/etc1/example_image.png.ktx": Cannot handle this KTX data type: glType 0x0000, glTypeSize 1, glFormat 0x0000, glInternalFormat 0x8D64, glBaseInternalFormat 0x1907
```
So I changed default format for ETC1 from DDS to KTX which knows this compression type.


and3md:
Rebased to master.

michaliskambi:
1. Question: you mention "You need `astcenc` in your `PATH` variable if you want generate ASTC textures." . I see that the code uses PVRTexTool to generate such textures. Does it mean that PVRTexTool calls `astcenc` under the hood?

    Or do you mean that we could have alternative implementation of `AutoGenerateTextures` that calls `astcenc` instead of PVRTexTool for ASTC?

    It's OK in any case, I'm just curious :) And I want to know how to update https://castle-engine.io/creating_data_auto_generated_textures.php .

2. I have a small request for change in code: 

    - Extend the `TTextureCompression` type with additional field `FileExtension: String` . 
    - Then in `TextureCompressionInfo` constant array specify `FileExtension: '.dds'` or `FileExtension: '.ktx'` (KTX for ETC1 and ASTC).
    - Oh, and add a note to `TTextureCompression.DDSFlipped` that "This field is ignored if FileExtension is not .dds".
    - The advantage is that now you will not have to place the list of formats supporting KTX inside TMaterialProperties.TAutoGeneratedTextures.GeneratedTextureURL . Instead, TMaterialProperties.TAutoGeneratedTextures.GeneratedTextureURL can do

```
  if UseCompression then
    Result := Result + TextureCompressionInfo[TextureCompression].FileExtension
  else
    ...
```

I think this will be a bit cleaner. This way the "behavior" of each compression format is nicely contained in the `TextureCompressionInfo` array.

src/x3d/castlematerialproperties.pas
```diff
@@ -584,7 +584,41 @@ function TMaterialProperties.TAutoGeneratedTextures.GeneratedTextureURL(
...
+      tcASTC_10x5_SRGB8_ALPHA8,
+      tcASTC_10x6_SRGB8_ALPHA8,
+      tcASTC_10x8_SRGB8_ALPHA8,
+      tcASTC_10x10_SRGB8_ALPHA8,
+      tcASTC_12x10_SRGB8_ALPHA8,
+      tcASTC_12x12_SRGB8_ALPHA8] then
+      Result := Result + '.ktx'
+    else
+      Result := Result + '.dds';
+  end
```
michaliskambi: This could be replaced by

```
Result := Result + TextureCompressionInfo[TextureCompression].FileExtension
```

per my previous comment :)

and3md:
> Does it  mean that PVRTexTool calls `astcenc` under the hood?

Yes,  `PVRTexTool` calls `astcenc` under the hood.

> I have a small request for change in code:

Make sense, adding extension to `TTextureCompression` is better solution :)

and3md:
I think it's OK now :)

michaliskambi:
Thank you!

I also did small updates to docs, https://github.com/castle-engine/cge-www/commit/46e3c7e56fee7a2ecb1dbed018979bfc92406422 , to the page https://castle-engine.io/creating_data_auto_generated_textures.php .

I'll announce this on CGE news soon.

PR #125, opened 2019-07-16
 Proper resume android activity after pause without context destroy (fixes sound missing after resume)

and3md:
In some cases context is not destroyed, so `ContextOpen` will not be called. Test case: share button click and return to app.


and3md:
Rebased to master.

and3md:
Sorry I accidentally deleted wrong branch.

michaliskambi:
Thank you!

PR #126, opened 2019-07-22
Remove sound buffers when change repository URL.

and3md:
Loading all music on Android is quite slow. In addition, every time you show ad or send the app to the background, it reloads.
To work around the problem I used several XML files, such as `menu.xml`, `level1.xml`, `level2.xml`.
But currently changing the repository file does not delete sound buffers, so reopening OpenAL context reloads sound buffers from previous XML files and from the new one. This PR changes this the sound buffers are freed before the new XML file is loaded.

### DIFF
--- a/src/audio/castlesoundengine.pas
+++ b/src/audio/castlesoundengine.pas
@@ -2787,9 +2787,31 @@ var
   I: TXMLElementIterator;
   BaseUrl: string;
   TimeStart: TCastleProfilerTime;
+  J: Integer;
+  SoundInfo: TSoundInfo;
+  SoundInfoBuffer: TSoundInfoBuffer;
 begin
   if FRepositoryURL = Value then Exit;
   FRepositoryURL := Value;
+
+  { Remove sound buffers }
+  for J := 0 to FSounds.Count - 1 do
+  begin
+    SoundInfo := FSounds[J];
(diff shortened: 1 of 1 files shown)


michaliskambi:
I think I see a bit simpler way to solve this: in SetRepositoryXml, we already call `FSounds.Clear` that frees the `TSoundInfoBuffer` instances. So `TSoundInfoBuffer` destructor should call `FreeBuffer`.

I did it in https://github.com/castle-engine/castle-engine/commit/9fef077f648d8f0568dbe26ddfae1c8b506978f6 .

Please test does it solve your issue :) If yes, then I think my solution is a bit simpler, so I would close this issue without applying your changes.

P.S. I was testing it with a tweaked version of `examples/3d_sound_game/lets_take_a_walk.lpr`, I'm attaching a patch in case it is useful.
[test_repository_xml_reload.patch.txt](https://github.com/castle-engine/castle-engine/files/3423834/test_repository_xml_reload.patch.txt)

and3md:
Why didn't I think of just add `FreeBuffer` to `TRepoSoundEngine` destructor :)

I think It should work. I will test it tomorrow. Thanks!

and3md:
Your solution works great, thanks :)

PR #127, opened 2019-07-31
Streaming music - WIP

and3md:
After #126 I still have problems with loading music. Decoding even one whole file results in a noticeable performance drop on mobile devices. Testers do not have mercy;) So I spent two evenings trying to implement streaming music. To see if this will improve performance. After that I can say that streaming music buffers can fix this issue.

This is first draft so do not check the details, let me know if you like the direction and general API changes. Or propose how it should be done :)

Currently only OGG is supported and OpenAL. To make sound buffer streamed add stream="true"
in XML Repo File. With this changes my game load levels immediately. Without streaming level load need 1-2 second to start. Even with one 700kB ogg music file to load, RAM usage drops by 20MB.


src/audio/openal/castleopenalsoundbackend.pas
```diff
@@ -50,10 +52,35 @@   TOpenALSoundBufferBackend = class(TSoundBufferBackendFromSoundFile)
...
+    function FillBuffer(ALBuffer: TALuint): Integer;
+  public
+    procedure FeedBuffers(ALSource: TALuint);
+    procedure StopFeedingBuffers;
+
+    procedure ContextOpenFromStreamedFile(const StreamedSoundFile: TStreamedSoundFile); override;
+    procedure ContextClose; override;
+  end;
+
+  TOpenALStreamFeedThread = class(TThread)
```
michaliskambi: Note that some platforms do not have working threads now. E.g. Nintendo Switch in the current implementation. Other weird platforms may have this limitation too :)

So the threading code should be surrounded by a clause like `{$ifdef CASTLE_SUPPORTS_THREADING}`. When it is not defined, the code should still work, just fallback to non-threading version, which means:

- Maybe read parts of files in the main thread. That is, do what you're currently doing, but not inside TThread.Execute. Only if it makes sense at all for some buffer sizes (we would need to read really small chunks of files to not cause a slowdown, I guess? And any file access is actually a bad idea in the middle of rendering... So this may be a stupid idea from my side.)

- Or just fallback to a version of playing without streaming.

The point is, game sound should work, even when the platform doesn't support streaming.

src/audio/openal/castleopenalsoundbackend.pas
```diff
@@ -274,16 +471,30 @@ procedure TOpenALSoundSourceBackend.SetMaxGain(const Value: Single);
...
+      FullLoaded := TOpenALSoundBufferBackend(FBuffer);
+      { TSoundBuffer is unsigned, while alSourcei is declared as taking signed integer.
+        But we know we can pass TSoundBuffer to alSourcei, just typecasting it to
+        whatever alSourcei requires. }
+      {$I norqcheckbegin.inc}
+      alSourcei(ALSource, AL_BUFFER, FullLoaded.ALBuffer);
+      {$I norqcheckend.inc}
+    end
+    else if FBuffer is TOpenALStreamBuffersBackend then
+    begin
```
michaliskambi: Make a newline before `if` here, we usually write code in CGE like this:

```
if xxxx then
begin
  ...
end else 
if yyy then
begin
  ...
end;
```

So the `else` is at the end of the line with `end`, not at the beginning of the line with next `begin`.

src/audio/ogg_vorbis/castleinternalvorbisdecoder.pas
```diff
@@ -40,9 +40,20 @@   EVorbisFileError = class(EVorbisLoadError);
 function VorbisDecode(Stream: TStream; out DataFormat: TSoundDataFormat;
   out Frequency: LongWord): TMemoryStream;
 
+procedure OpenVorbisFile(var VorbisFile: TOggVorbis_File; Stream: TStream; out DataFormat: TSoundDataFormat; out Frequency: LongWord);
```
michaliskambi: Can we maybe implement a class like `TOggVorbisStream`, that descends from `TStream` ? It's constructor / `Read` / destructor methods could play the same role as current OpenVorbisFile / ReadVorbisFile / CloseVorbisFile do. Seeking and writing can just raise exceptions.

This would make the code using this even more obvious, as the code would use standard Pascal TStream API.

And then `VorbisDecode` would basically be a function that copies the `TOggVorbisStream` into `TMemoryStream`, which seems sensible, since that's what happens when we "load a file as a whole".

src/audio/openal/castleopenalsoundbackend.pas
and3md: Sometimes I forget in which line what should be, so usually before PR I check all blocks with the open coding convention page. Because I considered this implementation as an early draft, I skipped this step. Now when I know the direction is OK and you want this, I will polish the code :)

and3md:
> P.S. I like the simple API change of `LoadBuffer` (just additional parameter). This makes it easy for code not using sounds.xml to load buffers with streaming. Kudos.

Thanks :) At first I just wanted to add `Streamed: Boolean` but we have `const ExceptionOnError: Boolean = true` and that makes confusion (what true/false means), so I decided to make enumeration (`TSoundBufferType`). This allows to easily add an overloaded LoadBuffer function without this new parameter for backward compatibility.

src/audio/openal/castleopenalsoundbackend.pas
and3md: > Note that some platforms do not have working threads now. E.g. Nintendo Switch in the current implementation. Other weird platforms may have this limitation too :)

No threads, I did not expect that :) Is this a limitation of FeePascal or Nintendo Switch?

michaliskambi: It's a limitation of my port to Nintendo Switch (the Nintendo Switch parts are in closed-source CGE version now, as Nintendo doesn't allow to publish anything about their API to general public).

It is 100% possible to make Pascal `TThread` on Nintendo Switch working, but it would be an additional work that I simply didn't do yet :)

and3md:
I made draft 2 of streaming music feature. Some changes in the concept were needed. But the result is much better. 
Now multiple playback on the same time is supported :) So there is a dictionary of buffer resources for each sound source. Maybe code is more complicated but adding music to your xml repository with stream=true will not allocate/read form disk anything before start using stream buffer (playing). Android likes that :) The last thing I need to figure out is looping. After that I clean the code.

and3md:
OK, now sound looping works! Tomorrow more code cleaning :)

src/audio/castleinternalsoundfile.pas
```diff
@@ -90,7 +90,58 @@   TSoundFile = class
     procedure ConvertTo16bit; virtual;
   end;
 
+  { TStreamedSoundFile }
```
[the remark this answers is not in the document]
and3md: I was thinking about this and it's hard for me to create some reasonable base class, the use of which will not require continuous casting to the sub-type (as in the case of `TAbstractSoundFile.CreateFromFile(...):TAbstractSoundFile`). I think the differences in the method of use are too big for me to connect these two branches. A common ancestor will always require casting after createFromFile() or class responsibility will be unclean (by adding all methods and implement only some of them). Unless I do not see any obvious solution, that you see.

BTW. I know it's quite annoying to have two classes for each sound format like `TSoundOggVorbis` and `TStreamedSoundOggVorbis`. But I think it's easier to maintain two simple classes (with well-defined responsibility) than one with a lot of "if streming then" conditions.

src/audio/openal/castleopenalsoundbackend.pas
and3md: I think that introducing threads and streams of music could greatly improve performance on the Switch (in your games) just like on Android. After switching to music streaming, my test devices  are also much cooler.

michaliskambi: Right now on Switch we use (at Cat-astrophe Games) FMOD backend, https://github.com/castle-engine/castle-engine/wiki/FMOD . The OpenAL is not available on Nintendo Switch. While there is a "native Nintendo Switch backend" implemented in CGE (part of closed-source CGE for Nintendo Switch code), but FMOD is just more functional.

FMOD provides streaming in a simpler way (I did not implement it yet, but I remember seeing it in docs) -- you just pass a flag like "streaming" to some API call, and it handles the actual file loading + streaming underneath. So it deals with eventual threads by itself.

michaliskambi: P.S. This actually makes my earlier comment ("streaming must work without threads") less useful. I said that threads are not implemented on Nintendo Switch, and now I said that on Nintendo Switch we don't use OpenAL :)

So, this has low priority.  For safety I would prefer to have a path for platforms that don't support threads but support OpenAL, but it's not really important now in any practical place.

src/audio/castleinternalsoundfile.pas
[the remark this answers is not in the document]
michaliskambi: I will need to think about it more :)

I understand the case you make. Indeed we want to have a clean code -- and I can see that sometimes (like in this case) it may be cleaner (producing code simpler to understand) to duplicate some stuff, than to try to reuse everything by a common class.

So, I think you're right here, and I'll try to do some thinking about it later :), maybe I'll be able to propose something.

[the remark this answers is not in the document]
and3md: Done in https://github.com/castle-engine/castle-engine/pull/127/commits/80c2989c7853a83cac9d36f54ab8a5588f666f8e

src/audio/openal/castleopenalsoundbackend.pas
and3md: Done in https://github.com/castle-engine/castle-engine/pull/127/commits/80c2989c7853a83cac9d36f54ab8a5588f666f8e

src/audio/ogg_vorbis/castleinternalvorbisdecoder.pas
and3md: This is so internal so... can we move that to another PR? I want now finish my game (ASAP) :)

and3md:
I think it is now ready for review. Works perfect in my game (no lags when load levels) and make CGE more more mobile game friendly. I think it will be helpful for many users.

michaliskambi:
Cool, thank you! I'll review it around this weekend.

( Yes, the switch to `TOggVorbisStream` can be made later, that's low priority indeed. )

eugeneloza:
>  I think it will be helpful for many users.

Indeed, thanks!!! :)

michaliskambi:
I think I have an idea how to make the code a bit more flexible in `castleinternalsoundfile.pas`.  I will want to experiment with it myself ASAP :) I'll see if this makes sense then.

Idea:

1. Implement `RegisterSoundFormat(const MimeType: String; const SoundReader: TSoundReadEvent)`, and `TSoundReadEvent` is a callback that for URL returns a stream. So we make 1 implementation of it for OggVorbis (that returns TOggVorbisStream, doing in TOggVorbisStream.Create/Read/Destroy the same things we do now in OpenVorbisFile/ReadVorbisFile/CloseVorbisFile), and 2nd implementation for WAV (that can return just TMemoryStream, this will not require to change current WAV reader much).

    The `TSoundReadEvent` is *advised* to not return TMemoryStream. Instead it should actually read the file in each `Read` call. This allows streaming of the registered sound format.

    But it is still correct if the `TSoundReadEvent` returns TMemoryStream. Everything will work, just requesting "streaming" for this format is useless (since the whole data will be read at loading anyway).

    Benefits:

    - Everything else is sound-format-independent. We avoid TStreamedSoundOggVorbis class. If we implement other sound formats in the future (like MP3, or a zilion other formats supported by libsndfile), they will automatically work in "streaming" mode too.

    - I wanted to introduce `RegisterSoundFormat` at some point anyway.

    - I wanted to introduce `TOggVorbisStream` anyway.

2. TSoundFile, when reading uses the appropriate reader (calls some `TSoundReadEvent`).

    - A method like `TSoundFile.LoadStream` more-or-less just returns the stream returned by proper registered `TSoundReadEvent` (this way it allows streaming).

    - A method like `TSoundFile.LoadMemoryStream` calls `LoadStream` and converts the stream to TMemoryStream (if it is not TMemoryStream already). It uses ReadGrowingStream to move data from any TStream to TMemoryStream.

    Benefits:

    - I think we can avoid TStreamedSoundFile class this way. Instead, `TSoundFile` can be useful both for streaming and non-streaming needs.

    - We no longer need `VorbisDecode`. Instead, `TSoundFile.LoadMemoryStream` does the equivalent work in format-independent way.

and3md:
Sounds good, I think your solution will be better in long term (will be better with each added format, I like `RegisterSoundFormat` idea :). In my solution I wanted to make changes as similar as possible to the current code and not change current `SoundFile` so much  (I'm still learning CGE code base, don't wanted to break too much things). So I was blind for some solutions...

Especially because this division to two classes, looks very clear for me. When you look at `TSoundFile` you know that it loads full data (there is `Data`, `DataSize`), and in streaming version there is only `Read`, `Rewind`. But unfortunately it comes with a price. So I think your solution will be better. 

But high level changes are OK?
I think division to `TSoundBufferBackendFromSoundFile` and `TSoundBufferBackendFromStreamedFile` should stay. 

They job is most important for me: When you have streamed file, you don't want to load anything at game start time (`contextOpen`) . This will be always waste of resources (and will kill android on activity change). Everything should be done in start playing (to be exact when buffer is added to sound source). `TSoundBufferBackendFromStreamedFile` cares about that.
In other hand when you want load full sound you don't want to make it on first usage but on game/level start (to be exact in `contextOpen`). `TSoundBufferBackendFromSoundFile` is for that. 

I think `TOpenALStreamBuffersBackend` (`TSoundBufferBackendFromStreamedFile` descendant)  is complicated enough to justify such a division. 

BTW, Within one module, I was quite comfortable using private variables between classes. What is your view on this? Should I make more setters/getters?

michaliskambi:
> I think division to TSoundBufferBackendFromSoundFile and TSoundBufferBackendFromStreamedFile should stay.

Sure, your explanation makes sense :) I'll be playing with the code, and for 90% leave them be.

> BTW, Within one module, I was quite comfortable using private variables between classes. What is your view on this? Should I make more setters/getters?

It's cool for me. All classes in the same unit are "friends" in C++ sense (they can use each other's private sense), it's cool to use this feature.

One thing I would advise (I try to do in most new code) is to have different sections for `strict private` and `private`. This way, when reading class declaration, it's obvious what is really only used by this class (`strict private`) and what is used potentially by all classes in this unit (`private`). In most cases, most stuff will go to `strict private`. In my experience, this division helps me easily understand the code.

and3md:
> One thing I would advise (I try to do in most new code) is to have different sections for strict private and private. This way, when reading class declaration, it's obvious what is really only used by this class (strict private) and what is used potentially by all classes in this unit (private). In most cases, most stuff will go to strict private. In my experience, this division helps me easily understand the code.

Ok, I will remember that in the new code :)

michaliskambi:
Merged! Sorry for taking so long, I invented a few things I wanted to add, and pushed them :)

Summary of the changes I did:

- RegisterSoundFormat introduced, and Vorbis decoding expressed as TOggVorbisStream.

- FMOD backend supports streaming too.

- Also (completely unrelated to this PR, but I needed it to test FMOD on Debian testing) FMOD library is now loaded dynamically, in a smarter way (we can detect at runtime if it's not available, and it works with GNU 2.32).

- Reading Duration for streamed sounds works (it requires reading file header, which in case of OggVorbis contains the duration information, without the need to decompress whole file).

- examples/audio/play_sounds/ has comments how to trivially activate streaming of all sounds.

- Fix streaming of short sounds (that do not need 4 buffers), previously they were causing OpenAL errors.

Notes:

- I made care to preserve the logic you mentioned, that merely loading a buffer with slStreaming should not do *anything* (not even reading the file header). The header is read on-demand when accessing various stuff (like Duration, after my changes).

- Note that I did not merge the TSoundFile and TStreamedSoundFile classes, after all (even though I planned this). Working with them, I saw clearly what you already mentioned earlier -- they are similar in some regards, but they also need to do some things differently, and merging them into one didn't seem like something beneficial after all.

    However, I simplified whole implementation of CastleInternalSoundFile :)

Kudos for developing this!

This was a large change, and I saw better the reasons for some things (e.g. why TOpenALStreamBuffersSoundSourceRes.ContextOpen creates TStreamedSoundFile) when trying to fiddle with them :) A few times, I wanted to try something, only to realize that your solution is optimal, and I ended up merely adding comments:) Thank you!

and3md:
I thank you for finding time to introduce such big changes in such a short time. And you improved them a lot. I read all changes. Code looks great. CGE gained another important functionality :) I'm glad that I could help in this :)

PR #128, opened 2019-08-11
New collision detection algorithm which cares about duplicate events - WIP

and3md:
Old collision detection events code was based on `TKraft.OnContactBegin`/`TKraft.OnContactEnd`. But sometimes there was two `OnContactBegin` events and one `OnContactEnd` or vice versa. That makes this event very hard to use in real code (for example double damage problem). This new approach based on `TKraftRigidBody.OnPostStep` cares to make only one `OnCollisionEnter` and only one `OnCollisionExit`. 

Extra feature `OnCollisionStay` - new event that is fired each physics frame when collision still remain. 

In addition, the code is only called for bodies that have `OnCollisionXXXX` event assigned. That should be faster.

### DIFF
--- a/src/3d/castletransform_physics.inc
+++ b/src/3d/castletransform_physics.inc
@@ -210,7 +210,7 @@
       (you can free it later, e.g. during @link(TCastleTransform.OnUpdate)). }
     property OnCollisionExit: TOnCollision read FOnCollisionExit write SetOnCollisionExit;
 
-    { Occurs when TRigidBody still colliding with another TRigidBody.
+    { Occurs when TRigidBody still collides with another TRigidBody.
 
       @italic(Warning:) Do not free the @link(TCastleTransform) instances
       that collide during this event.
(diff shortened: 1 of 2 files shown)


and3md:
I think I found some issues so please wait with merge.

and3md:
This solution works for me, now there are no double `CollisionEnter`/`CollisionExit` events. Perhaps some checks are redundant. I am still learning kraft code so I cannot determine that yet. Either way, it works better than the current implementation. So I think it can be merged. I think I will improve this implementation every time I have more knowledge about kraft :)

src/3d/castletransform_physics.inc
```diff
@@ -193,7 +208,17 @@
       It would free the rigid body instance, which will crash the physics engine for now.
       Instead, you can set @link(TCastleTransform.Exists) to @false
       (you can free it later, e.g. during @link(TCastleTransform.OnUpdate)). }
-    property OnCollisionExit: TOnCollision read FOnCollisionExit write FOnCollisionExit;
+    property OnCollisionExit: TOnCollision read FOnCollisionExit write SetOnCollisionExit;
+
+    { Occurs when TRigidBody still colliding with another TRigidBody.
```
michaliskambi: "still colliding" -> "still collides" (in English you should say "X still collides with Y" or "X is still colliding with Y")

and3md: Sorry, I'm still learning English ;)

PR #130, opened 2019-08-17
Limit logs in Timer function to not affect performance

and3md:
When I test my game, I noticed that sometimes (very rare, about one time on 20 minutes) the physics freezes, and moving rigid bodies appear about 2 centimeters further (on 5'' screen). Then I discovered that when this happens logs receive a lot of `Time: Detected gettimeofday () going backwards on Unix, workarounding. This is known to happen on some Android devices`. 

There can be about 300 lines with this information. So I tried to limit this logs. After that freezes still occur, but they are shorter, and don't affect physics so much (about 0.5 centimetre).

The log looks better also.
This PR limits this log to one occurrence.

### DIFF
--- a/src/base/castletimeutils.pas
+++ b/src/base/castletimeutils.pas
@@ -593,6 +593,7 @@ var
   { Note that using this makes Timer not thread-safe
     (but we neved guaranteed in the interface that it's thread-safe...). }
   LastTimer: TTimerResult;
+  LastTimerLog: TTimerResult;
 {$endif}
 
 function Timer: TTimerResult;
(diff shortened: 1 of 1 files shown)


michaliskambi:
Thanks! Merged.

We have a few more places in CGE where logs need to be limited to avoid flooding, indeed they can kill performance (and make log useless) if they are produced too often. Sometimes I display a log only once (guarded by Boolean), sometimes I limit it to 10 occurrences (like in TTextureFontData.MakeFallbackWarning https://github.com/castle-engine/castle-engine/blob/master/src/fonts/castletexturefontdata.pas#L535 ).

Issue #131, opened 2019-08-18
Android time issue


and3md:
I did a small investigation regarding the problem of time errors on Android. And I found that Android provides three clocks. And most likely we use the wrong one. More information in https://developer.android.com/reference/android/os/SystemClock.html .
I don't know how to fix it in FreePascal at the moment. So I add this as an issue. Maybe @michaliskambi knows how to fix it :)

and3md:
Ok, I found solution, I make PR, after more testing.

michaliskambi:
Closing, further discussion in #132 .

PR #132, opened 2019-08-18
Fix Android time issue #131.

and3md:
As I say in #131 Androdid have three clocks. I found that `clock_gettime(CLOCK_MONOTONIC)` uses clock that we need (https://stackoverflow.com/questions/3832097/how-to-get-the-current-time-in-native-android-code). 

Because Android is linux I use `clock_gettime(CLOCK_MONOTONIC, @tp);` from linux module. This works for me. I played 2 hours my game without any freezes :) Tomorrow will do more test. Please test this too on your projects/devices.

### Kraft
I also had to fix kraft. Do we use any special version? If I do PR in Kraft repo, will I be able to just use the latest version in CGE?

### DIFF
--- a/src/base/castletimeutils.pas
+++ b/src/base/castletimeutils.pas
@@ -26,7 +26,7 @@ interface
 
 uses
   {$ifdef MSWINDOWS} Windows, {$endif}
-  {$ifdef UNIX} BaseUnix, Unix, Dl, {$endif}
+  {$ifdef UNIX} BaseUnix, Unix, Dl, {$endif} {$ifdef ANDROID}linux, {$endif}
   SysUtils, Math, Generics.Collections,
   CastleUtils;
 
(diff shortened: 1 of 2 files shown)


src/base/castletimeutils.pas
```diff
@@ -591,28 +591,28 @@ function Timer: TTimerResult;
 {$ifdef ANDROID}
 var
   { Note that using this makes Timer not thread-safe
-    (but we neved guaranteed in the interface that it's thread-safe...). }
+    (but we neved guaranteed in the interface that it's thread-safe...).
+    This variables can be removed once everyone confirms that the solution is
+    always working.
```
michaliskambi: Improve the comment "solution" -> "solution of using CLOCK_MONOTONIC", to remember in the future what solution does this mention :)

src/base/castletimeutils.pas
```diff
@@ -26,7 +26,7 @@ interface
 
 uses
   {$ifdef MSWINDOWS} Windows, {$endif}
-  {$ifdef UNIX} BaseUnix, Unix, Dl, {$endif}
+  {$ifdef UNIX} BaseUnix, Unix, Dl, {$endif} {$ifdef ANDROID}linux, {$endif}
   SysUtils, Math, Generics.Collections,
```
michaliskambi: `Linux` (from uppercase), we write all unit names CamelCase :)

src/base/castletimeutils.pas
```diff
@@ -591,28 +591,28 @@ function Timer: TTimerResult;
...
   LastTimerLog: TTimerResult;
-{$endif}
 
 function Timer: TTimerResult;
 var
-  tv: TTimeval;
+  tp: TimeSpec;
 begin
-  FpGettimeofday(@tv, nil);
+  { Android have three clocks we need use clock which never leaps forward
```
michaliskambi: "have" -> "has" (3rd person singular in English uses "has").

src/base/castletimeutils.pas
```diff
@@ -591,28 +591,28 @@ function Timer: TTimerResult;
...
-  {$ifdef ANDROID}
-  { We cannot trust some Android systems to return increasing values here
-    (Android devices we found this bug:
-     - "Moto X Play", "XT1562", OS version 5.1.1;
-     - "Xiaomi Redmi Note 4", MediaTek Helio X20, OS version 6.0, MIUI Global 10.2
-    ).
-    Maybe they synchronize the time from the Internet, and do not take care
-    to keep it monotonic (unlike https://lwn.net/Articles/23313/ says?) }
+  { I leave this test to check the solution, but it really isn't needed anymore }
   if Result.Value < LastTimer.Value then
```
michaliskambi: Here also improve comment "solution" -> "solution of using CLOCK_MONOTONIC".

and3md:
Should be OK now. Thanks for the corrections :)

michaliskambi:
Merged both your PRs. Thank you!!!

PR #133, opened 2019-08-21
Build tool: Set minimal FPC version that support 64-bit Android to 3.2.0

and3md:
FPC 3.2.0 supports AArch64 (https://forum.lazarus.freepascal.org/index.php?topic=42400.0).


michaliskambi:
Thanks!

PR #134, opened 2019-08-22
Android updates

and3md:
Upgrades:
- default API level increased 28 to comply with Google requirements
- Google Services, Google Play Services and Admob upgraded to versions from https://developers.google.com/android/guides/setup#split According to our conversation. Tested on android 4.1.1 and on newer devices. I tested only login and achievements on Google Play Services.


michaliskambi:
Thank you!

michaliskambi:
I'll make a news post about it next weekend :)

PR #135, opened 2019-09-01
Latest Kraft version, with #131 fixed and other improvements.

and3md:
Latest version from https://github.com/BeRo1985/kraft master branch. Tested on Android and Linux.


michaliskambi:
Thank you!:)

PR #136, opened 2019-09-03
Removed unneeded (after #132) time measurement test on Android.

and3md:
After two weeks of tests, I think its time to remove these checks :)


michaliskambi:
It''s Sunday evening  and noone reported any problems, so applying :)

PR #137, opened 2019-09-12
Read additional unicode characters for a font in CastleSettings

eugeneloza:
As by default the fonts are created with only ASCII charset, now the user is able to specify something like `characters="©ŚĆĘĄŹŁŻÓŃśćęąźłżóń"` to include those characters in font. Also it is possible to specify `exclude_ascii="true"` not to load ASCII symbols at all, and use only the provided charset, e.g. that looks like a nice idea for numbers-only font.

Manual text for: https://castle-engine.io/manual_castle_settings.php

By default the font loads only ASCII symbols for the specified font. If you want to add other UTF8 characters, you can do it by providing `characters="some_sample_string"`, e.g.  `characters="ŚĆĘĄŹŁŻÓŃśćęąźłżóń"` which would add to normal set of ASCII symbols Polish letters. You can also prevent the font from loading ASCII symbols by specifying `exclude_ascii="true"`, in this case only characters specified by `characters` field will be added. Note that `exclude_ascii` is ignored in case no characters are specified and ASCII characters will be loaded by default.


michaliskambi:
Looks great, thank you! Merging :)

michaliskambi:
I completely reworked the page on https://castle-engine.io/manual_castle_settings.php , and documented there new attributes. Thank you!

PR #139, opened 2019-09-26
Android: Fixed TCastleApplication.ProcessMessage loop suspending

and3md:
This PR fix `TCastleApplication.ProcessMessage` loop suspending when game was sent to background. Without this fix when game is in background the loop still works unnecessarily wasting battery and CPU.

Now the condition only checks for the existence of the window, I removed other conditions from the check. 

## Why I removed `AllowSuspendForInput`?
In android game we need suspend loop always when the app is in background. I think `AllowSuspendForInput` will never be true in "normal" game because of:

```
function TExamineCamera.AllowSuspendForInput: boolean;
begin
  Result := false;
end;
```
The same in `TWalkCamera`.

```
function TCastleOnScreenMenu.AllowSuspendForInput: boolean;
begin
  Result := false;
end;
```

```
function TCastleApplication.AllowSuspendForInput: boolean;
var
  I: Integer;
begin
  Result := not (
    Assigned(OnUpdate) or
    Assigned(OnTimer) or
    (ApplicationProperties.OnUpdate.Count <> 0));
  if not Result then Exit;

  for I := 0 to OpenWindowsCount - 1 do
  begin
    Result := OpenWindows[I].AllowSuspendForInput;
    if not Result then Exit;
  end;
end;
```
## Tests

Tested on my game, drawing_toy, simple_3d_demo, test_bump_mapping.

### DIFF
--- a/src/window/castlewindow_android.inc
+++ b/src/window/castlewindow_android.inc
@@ -627,13 +627,15 @@ var
 begin
   MessageHandlingBegin := Timer;
   repeat
-    { Android app can't be suspended by WaitForMessage and AllowSuspendForInput
-      because we do not react to resize events soon enough then,
-      since we have to actually do a couple of loop passes until resize
+    { When the window is open: We cannot suspend (wait indefinitely for a message) 
+      an open window, even when AllowSuspendForInput or WaitForMessage. 
+      That's because we would not react to resize events soon enough then,
+      since we have to actually do a couple of loop passes until "resize" event
       reaches us. Reproducible on drawing_toy.
 
-      On the other hand, we must always stop the message loop when the
-      window is closed to not waste battery. }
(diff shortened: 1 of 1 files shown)


michaliskambi:
Thanks! 

Indeed, `AllowSuspendForInput` will be true for majority of CGE applications. Anything that needs to animate even when there's no user input (e.g. gravity of `TWalkCamera` must walk regardless if you press some buttons) must disable `AllowSuspendForInput`. That said, `AllowSuspendForInput` should be just ignored when the window is closed. Looks like I broke it when adding `... and (MainWindow=nil)...` rule, I should have just removed `AllowSuspendForInput` check as you did.

Thanks for untangling it:)

and3md:
For the accuracy only, you certainly thought about `false` value. 

Thanks for merging :)

PR #140, opened 2019-10-08
Triggers implementation for TRigidBody

and3md:
Triggers implementation for `TRigidBody`. To make trigger just set `Trigger` property to true. 

Triggers use the same events as other rigid bodies `TRigidBody.OnCollisionEnter`, `TRigidBody.OnCollisionStay`, `TRigidBody.OnCollisionExit`.

I added simple trigger to `physics_2d_collisions` example (I think it will be expanded when we add more features to physics).


michaliskambi:
I'll make a news post about it around the next weekend :)

PR #142, opened 2019-10-19
Access to physics settings in TCastleSceneManager

and3md:
Allows change frequency and velocity integration scheme. Default values like in Kraft. 
Maybe we should check changes in `TSceneManagerWorld.Update` like `gravity` but I'm not sure if this is needed.


and3md:
Physics settings update is now implemented :)

PR #143, opened 2019-10-26
Fixed build-tool compilation.

and3md:
I just needed to compile build-tool, but there was some unit lacks and error on project load:

```
The file "/home/and3md/fpc/castle_game_engine/tools/build-tool/code/toolarchitectures.pas" was not found.
```
Maybe you work on build-tool now, in that case just ignore(close) this PR, otherwise, this PR fixes the problem.

### DIFF
--- a/tools/build-tool/code/castle-engine.lpi
+++ b/tools/build-tool/code/castle-engine.lpi
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CONFIG>
   <ProjectOptions>
-    <Version Value="10"/>
+    <Version Value="11"/>
     <General>
       <Flags>
         <MainUnitHasUsesSectionForAllUnits Value="False"/>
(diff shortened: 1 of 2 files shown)


michaliskambi:
Thanks! I simply didn't test compilation of the build tool using LPI (from Lazarus) since quite some time. I always used the script `castle_game_engine/tools/build-tool/castle-engine_compile.sh` that calls FPC with proper options.

I'll add ASAP to Jenkinsfile to test it. I *want* to support compilation using LPI too, Jenkins  will make sure it is tested :)

michaliskambi:
It turns out that Jenkins already tests that compilation using `lazbuild` of the build tool works. More precisely, `Jenkinsfile` says to call `make examples-laz` which compiles various things, including the build tool, through `lazbuild xxx.lpi`.

The compilation of the build tool worked (before your PR too), using `lazbuild` or GUI `lazarus`, in my tests. I'm now on Windows using Lazarus 2.1.0 rev 61813. Even though the LPI was indeed incorrect (e.g. it pointed to `toolarchitectures.pas` instead of `../../common-code/toolarchitectures.pas` which you fixed). Compilation went OK probably because `../../common-code` was on the unit search path, so FPC could find `toolarchitectures.pas`  even when Lazarus had wrong location of it.

In any case, thanks to your fixes LPI is now really correct :) Thank you!

and3md:
I use Lazarus 2.0.2 (61236 - stable) with fpc 3.2 on this machine and when I opened castle-engine.lpi by Project->Open project... there was message box with project load error. This is Linux machine.

BTW. Some times ago (about 3 months ago) I tried to use Lazarus 2.1 but sometimes projects was damaged (I couldn't determine why, maybe now it's fixed) so I go back to 2.0.x.

PR #146, opened 2019-11-28
Fix gamepad D-Pad behavior on Windows

eugeneloza:


src/ui/castleinternaljoystickswindows.pas
```diff
@@ -245,8 +249,18 @@ procedure TWindowsJoysticksBackend.Poll(const List: TJoystickList;
       FillChar( Joystick.State.Axis[ JOY_POVX ], 8, 0 );
       if ( Joystick.Info.Caps and JOY_HAS_POV > 0 ) and ( state.dwPOV and $FFFF <> $FFFF ) then
       begin
-        Joystick.State.Axis[ JOY_POVX ] := Round( Sin( state.dwPOV and $FFFF div 100 ) );
-        Joystick.State.Axis[ JOY_POVY ] := -Round( Cos( state.dwPOV and $FFFF div 100 ) );
+        _value := Round( Sin( state.dwPOV and $FFFF div 100 / 180 * Pi ) );
```
michaliskambi: - To convert from/to radians, it's best to use DegToRad / RadToDeg from Math unit. This way it looks obvious to the reader :)

- The `Round` at the end seems unnecessary (possibly a leftover of some previous refactor -- I realize it was there before your changes). Since we operate on float values, can we just remove this `Round` and have smooth values? Right now, this `Round` unnecessarily changes the value to -1 / 0 / 1, if I read the code right?

- Also when calculating argument for Sin / Cos, use `/` instead of `div` since we are operating on floating point values anyway, so `(state.dwPOV and $FFFF) / 100.0` seems better (doesn't lose any fractional information from the result of divison).

michaliskambi: All in all, I think this line could be

```
_value := Sin( DegToRad((state.dwPOV and $FFFF) / 100.0) );
```

Similar comment to  line with `Cos` below.
 Round( Sin( state.dwPOV and $FFFF div 100 / 180 * Pi ) );

eugeneloza: Well, maybe we can go further? Now notation of the database starts being clear for me: dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1. They simply make it from dwPOV without any sin/cos conversions - 0.1*$FFFF is dpup; 0.2*$FFFF is dpright and so on. That doesn't correspond well to the reported values (we have 0, Pi/2, Pi, 3*Pi/2 - i.e. equally spaced; and they have 0.1, 0.2, 0.4, 0.8 i.e. 2^n/10, n=0..3). But anyway we can think of it this way.

Let me try if multiple D-Pad keys can be pressed simultaneously and whether it provides a consistent result (then our current sin/cos approach is indeed better).

eugeneloza: Tested everything and it looks like our current approach is better and safer :) Fixed according to your suggestions.

michaliskambi:
Cool, thank you!

I understand  that the point of this calculation is to express an angle as an XY axis, similar to other axis. So we have an angle, but we only expose its Sin / Cos. If someone wants an angle, (s)he will have to do ATan2, reversing these calculations :)

P.S.Note that one can calculate sinus and cosinus of the same angle in one using SinCos. Although that would be too much nitpicking in this case, it doesn't matter, as long as code is clear :)

eugeneloza:
Yeah, I see it that some gamepad may have not D-Pad (up-down-left-right), but POV (point-of-view) which can change axis arbitrarily, not only fixed 8 angles as D-Pad. This way we can/should expose D-Pad as a separate than Axis[i] property. I see that both in Linux and Windows we can do that relatively easily.
On Linux we have D-Pad reported at 16-17 axes, already processed like `_value` on Windows - in discrete [-1, 0, 1] range (like the formula with `Round` was giving us before). However, Windows gives us angle explicitly - and we can have DPad.X, DPad.Y and DPad.Angle. Also it's very reasonable that D-Pad would generate `keyPadUp...keyPadLeft` events, not axis movement, unless we really deal with analog D-Pad with very smooth axis rotation.

PR #149, opened 2019-12-07
castle-editor: Fixed taskbar button hidden on Windows

Kagamma:
On Windows, when we hide the main form using TForm.Hide, the button on taskbar also got hidden.
It is better to set Application.ShowMainForm field on Windows instead.

### DIFF
--- a/tools/castle-editor/code/formchooseproject.pas
+++ b/tools/castle-editor/code/formchooseproject.pas
@@ -75,20 +75,20 @@ uses CastleConfig, CastleLCLUtils, CastleURIUtils, CastleUtils,
 
 procedure TChooseProjectForm.Show;
 begin
-  {$IFDEF WINDOWS}
+  {$ifdef MSWINDOWS}
   Application.ShowMainForm := True;
-  {$ELSE}
-  TForm(Self).Show;
-  {$ENDIF}
+  {$else}
+  inherited Show;
+  {$endif}
 end;
 
 procedure TChooseProjectForm.Hide;
 begin
-  {$IFDEF WINDOWS}
+  {$ifdef MSWINDOWS}
   Application.ShowMainForm := False;
-  {$ELSE}
-  TForm(Self).Hide;
-  {$ENDIF}
+  {$else}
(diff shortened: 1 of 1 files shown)


tools/castle-editor/code/formchooseproject.pas
```diff
@@ -70,6 +73,24 @@ implementation
 
 { TChooseProjectForm ------------------------------------------------------------- }
 
+procedure TChooseProjectForm.Show;
+begin
+  {$IFDEF WINDOWS}
```
michaliskambi: We use `{$ifdef MSWINDOWS}` throughout the CGE code. I.e.
- Compiler directives are lowercase like `ifdef` not `IFDEF` (same note about `{$else}` and`{$endif}` below)
- We use `MSWINDOWS`, because that's also supported by Delphi. (FPC understands both `WINDOWS` and `MSWINDOWS`)

tools/castle-editor/code/formchooseproject.pas
```diff
@@ -70,6 +73,24 @@ implementation
...
+  Application.ShowMainForm := True;
+  {$ELSE}
+  TForm(Self).Show;
+  {$ENDIF}
```
michaliskambi: Instead of calling `TForm(Self).Show` I think it's a bit cleaner to call `inherited Show`. This more clearly shows the intention -- you don't want to call your own method, you want `Show` of  ancestor.

Same note about `TForm(Self).Hide` below.

tools/castle-editor/code/formchooseproject.pas
Kagamma: Fixed in latest commit.

tools/castle-editor/code/formchooseproject.pas
Kagamma: Haven't used Pascal for 2 years and I already forgot we can use inherited to access ancestor methods... Thanks for your remarks!

michaliskambi:
Thank you, merged!

PR #153, opened 2020-02-06
Implement TJoysticks.OnDisconnect on Windows and Linux

eugeneloza:


src/ui/castleinternaljoystickslinux.pas
```diff
@@ -145,54 +147,75 @@ procedure TLinuxJoysticksBackend.Poll(const List: TJoystickList;
...
+                    end;
+                end;
+            end;
+        end;
+        BytesRead := FpRead( BackendInfo.Device, event, 8 );
+      until BytesRead <> 8
+    else
+      if fpgeterrno <> EAGAIN then
+      begin
+        WritelnLog('Joystick error, possibly it was unplugged. Trying to run Joysticks.Initialize.');
```
michaliskambi: This log is no longer valid, since we call `OnDisconnect` instead of `Joysticks.Initialize` directly. Better and simpler:

```
WritelnLog('Joystick %s was disconnected', [Joystick.Info.Name]);
```

src/ui/castleinternaljoystickslinux.pas
```diff
@@ -145,54 +147,75 @@ procedure TLinuxJoysticksBackend.Poll(const List: TJoystickList;
...
+        JoystickHadBeenDisconnected := true;
+        { Note that only last unplugged joystick will be sent to the Unplugged event
+          if multiple joysticks have been unplugged simultaneously
+          which may happen e.g. in case an USB-Hub with several joysticks had been disconnected }
+        JoystickDisconnected := Joystick;
```
michaliskambi: Since it's not reliable (we could have multiple disconnections simultaneously, as you write) I would say to not try to remember/pass the `JoystickDisconnected`. I.e. let's not implement something we don't immediately need, when  we know the initial implementation would not be 100% reliable :) This way also code is a bit simpler.

So `Joysticks.OnDisconnect` would be parameter-less.

In the future we can make it better, e.g. collecting list of disconnected joysticks, if it will be ever needed.

src/ui/castleinternaljoystickswindows.pas
```diff
@@ -283,8 +293,20 @@ procedure TWindowsJoysticksBackend.Poll(const List: TJoystickList;
         Joystick.State.BtnUp[ j ] := btn = 0;
         if Assigned(EventContainer.OnButtonDown) and (btn <> 0) then EventContainer.OnButtonDown(Joystick, j);
       end;
+    end else
+    if JoyError = JOYERR_UNPLUGGED then
+    begin
+      WritelnLog('Joystick was unplugged. Trying to run Joysticks.Initialize.');
```
michaliskambi: Same comment as above for Linux: the log line can be improved.

src/ui/castleinternaljoystickswindows.pas
```diff
@@ -283,8 +293,20 @@ procedure TWindowsJoysticksBackend.Poll(const List: TJoystickList;
...
+      JoystickHadBeenDisconnected := true;
+      { Note that only last unplugged joystick will be sent to the Unplugged event
+        if multiple joysticks have been unplugged simultaneously
+        which may happen e.g. in case an USB-Hub with several joysticks had been disconnected }
+      JoystickDisconnected := Joystick;
```
michaliskambi: Same comment as above for Linux: let's not remember/pass `JoystickDisconnected`.

src/ui/castleinternaljoystickswindows.pas
```diff
@@ -215,7 +224,8 @@ procedure TWindowsJoysticksBackend.Poll(const List: TJoystickList;
     if BackendInfo.Caps.wCaps and JOYCAPS_POVCTS > 0 then
       state.dwFlags := state.dwFlags or JOY_RETURNPOVCTS;
 
-    if joyGetPosEx( i, @state ) = 0 then
+    JoyError := joyGetPosEx( i, @state );
+    if JoyError = JOYERR_NOERROR then
```
michaliskambi: You silently ignore the case when JoyError is neither `JOYERR_NOERROR` nor `JOYERR_UNPLUGGED` now.

Better to make it

```
case JoyError of
  JOYERR_NOERROR:
    begin
       ...
    end;
  JOYERR_UNPLUGGED:
    begin
       ...
    end;
  else WritelnWarning('Joystick %s unexpected error %d', [Joystick.Info.Name, JoyError]);
end;
```

src/ui/castlejoysticks.pas
```diff
@@ -175,6 +177,11 @@   TJoysticks = class
       but it may be called in other cases (e.g. "explicit" joystick backend,
       used by Nintendo Switch, may call this at any moment). }
     property OnChange: TNotifyEvent read FOnChange write FOnChange;
+
+    { Called in case a previously initalized joystick had been unplugged unexpectedly. }
```
michaliskambi: "had been unplugged unexpectedly" -> "has been disconnected".

So using our terminology, and also removing the word "unexpectedly" -- it is always unexpected :)

src/ui/castlejoysticks.pas
```diff
@@ -175,6 +177,11 @@   TJoysticks = class
...
+    property OnDisconnect: TNotifyEvent read FOnDisconnect write FOnDisconnect;
```
michaliskambi: OnDisconnect could be of type `TSimpleNotifyEvent`, defined `TSimpleNotifyEvent = procedure of object;`

src/ui/castleinternaljoystickswindows.pas
```diff
@@ -215,75 +245,90 @@ procedure TWindowsJoysticksBackend.Poll(const List: TJoystickList;
...
+          JoystickHadBeenDisconnected := true;
         end;
-        Joystick.State.BtnDown[ j ] := btn <> 0;
-        Joystick.State.BtnUp[ j ] := btn = 0;
-        if Assigned(EventContainer.OnButtonDown) and (btn <> 0) then EventContainer.OnButtonDown(Joystick, j);
-      end;
+      JOYERR_PARMS:
+        WritelnLog('Joystick %s parameters are no longer valid.', [Joystick.Info.Name]);
+      else
+        WritelnLog('Joystick %s error %d', [Joystick.Info.Name, JoyError]);
```
michaliskambi: This should be `WritelnWarning` (unless it's something that can routinely occur during normal program execution and we consciously know that it can be ignored, and doesn't indicate any error in our code).

src/ui/castleinternaljoystickslinux.pas
```diff
@@ -145,54 +147,70 @@ procedure TLinuxJoysticksBackend.Poll(const List: TJoystickList;
   event : TLinuxJsEvent;
   Joystick: TJoystick;
   BackendInfo: TLinuxJoystickBackendInfo;
+  BytesRead: TSsize;
+  JoystickHadBeenDisconnected: Boolean;
```
michaliskambi: Rename to  `JoystickHasBeenDisconnected` (had been -> has been), I think this is more correct in English ("has been" is used to indicate something happened in the past, like "has been turned off"; "had been" indicates a longer task that started and finished in past,  like "had been learning"). https://www.quora.com/When-should-I-use-has-been-have-been-and-had-been-1?share=1 , https://www.quora.com/How-can-I-determine-when-to-use-had-have-had-been-etc , https://www.quora.com/Whats-the-difference-between-had-been-and-has-been .

I know, nitpicking :)

michaliskambi:
Thank you!!!

PR #154, opened 2020-02-07
Detect Joystick connect or disconnect through WinAPI

eugeneloza:


src/ui/castlejoysticks.pas
```diff
@@ -181,7 +188,7 @@   TJoysticks = class
     { Called in case a previously initalized joystick has been unplugged. }
```
michaliskambi: unplugged->disconnected (I know, this is from last PR)

src/ui/castlejoysticks.pas
```diff
@@ -181,7 +188,7 @@   TJoysticks = class
...
     property OnDisconnect: TSimpleNotifyEvent read FOnDisconnect write FOnDisconnect;
     { Called in case a joystick has been plugged in the system. }
```
michaliskambi: "has been plugged in the system" -> "has been connected to the system"

src/window/windows/castlewindow_winapi.inc
```diff
@@ -448,6 +453,10 @@ function TCastleWindowBase.WindowProc(uMsg: UINT; wParm: WPARAM; lParm: LPARAM):
     MaybeRelease;
   end;
 
+const
```
michaliskambi: These constants are part of `JwaDbt`. Since we use winunits-jedi anyway (`JwaWinUser` above), then we can use `JwaDbt` here too.

src/window/windows/castlewindow_winapi.inc
```diff
@@ -719,6 +728,34 @@ begin
...
+            end;
+          { A device had been succesfully removed from the system }
+          DBT_DEVICEREMOVECOMPLETE:
+            begin
+              WriteLnLog('Received a WM_DEVICECHANGE event with DBT_DEVICEREMOVECOMPLETE');
+              //it might be a joystick, therefore
+              Joysticks.InternalDisconnected;
+            end;
+          { Something unspecified changed in the system devices configuration }
+          //DBT_DEVNODES_CHANGED: WriteLnLog('Received a WM_DEVICECHANGE event with DBT_DEVNODES_CHANGED'); // Ignore these events for now
```
michaliskambi: To actually ignore other events, we should not do `Exit(0)` in this case.

`WindowProc` at the end calls `DefWindowProc` (and other stuff) if you don't exit early.

I don't know what is the best course of action, but either

1. We should do `Exit(0)` only for 2 handled events - DBT_DEVICEARRIVAL, DBT_DEVICEREMOVECOMPLETE (seems more natural)
2. Or remove the comment "Ignore these events for now" in the line above. If we do `Exit(0)`, then we don't ignore them, 

https://docs.microsoft.com/en-us/windows/win32/devio/wm-devicechange docs say about the return value
"""Return TRUE to grant the request.
Return BROADCAST_QUERY_DENY to deny the request.
"""

Which suggests that the proper solution is

3. Always do `Exit(1)`. I don't know why SDL does `Exit(0)`.

src/window/windows/castlewindow_winapi.inc
```diff
@@ -854,6 +891,51 @@ procedure TCastleWindowBase.OpenBackend;
       [FDpi, FDpi / DefaultDpi, DefaultDpi]);
   end;
 
+  { Request WinAPI to send DBT_DEVICEARRIVAL and DBT_DEVICEREMOVECOMPLETE events
+    when new Joysticks have been connected to the system }
+  procedure RegisterForJoystickEvents;
+  const
```
michaliskambi: These constants are also in `JwaDbt`.

src/window/windows/castlewindow_winapi.inc
```diff
@@ -854,6 +891,51 @@ procedure TCastleWindowBase.OpenBackend;
...
+    DBT_DEVTYP_DEVICEINTERFACE = 5;
+    DEVICE_NOTIFY_WINDOW_HANDLE = 0;
+  type
+    DEV_BROADCAST_DEVICEINTERFACE = {packed?} record
```
michaliskambi: This type is also in `JwaDbt`.

src/window/windows/castlewindow_winapi.inc
eugeneloza: Hmmm... I'm not sure either. I'm not even sure if we actually need to `Exit` at all. So, let's go with "3".

PR #160, opened 2020-05-31
allow first frame of CustomAnimation to play on first loop

michaelvanham:
set FFrame before Inc(FCustomFrame) to allow first frame to play on first loop
-------
In the current TSprite, the first frame of custom animation is not played on the first loop. Please see the demo program attached. 4 frames of animation should be played on each button click, but only 3 are.

After patch, all 4 frames of animation are played on each button click.

[tsprite_animation_test.zip](https://github.com/castle-engine/castle-engine/files/4708088/tsprite_animation_test.zip)

### DIFF
--- a/src/images/opengl/castleglimages_sprite.inc
+++ b/src/images/opengl/castleglimages_sprite.inc
@@ -446,6 +446,9 @@ begin
     Anim := FCustomAnimations.Items[FCurrentAnimation];
     while (FTime - FCustomAnimTimestamp) >= Anim.Frames[FCustomFrame].Duration do
     begin
+      // set FFrame before Inc(FCustomFrame) to allow first frame to play on first loop
+      FCustomAnimTimestamp := FCustomAnimTimestamp + Anim.Frames[FCustomFrame].Duration;
+      FFrame := Anim.Frames[FCustomFrame].Frame;
       // todo: custom animation loops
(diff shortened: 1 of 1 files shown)


michaliskambi:
Thank you for the excellent report and testcase. The testcase really allowed me to easily investigate it,  really thanks! And sorry for large delay in answering -- too busy times for me :)

I decided to go with a bit different solution. 

- In your case, you achieved "displaying the first frame" by making `FCustomFrame` and `FFrame` desynchronized by 1 (`FCustomFrame` would always point to the *next* frame after the `FFrame` displays).

- Instead I wanted to keep the synchronization `FCustomFrame` and `FFrame`, and I also wanted to remove the weird delay visible in your demo. After you click a button, it takes a while before the new animation actually plays.

This was all because, as far as I can tell, `TSprite.SetTime` didn't update `FFrame` at all for one second (when condition `while (FTime - FCustomAnimTimestamp) >= Anim.Frames[FCustomFrame].Duration do` was false, the `FFrame` was not touched). I fixed this.

So now the first custom animation frame *is* played, and also playing the animation works instantaneously :) Please let me know If I missed something and my fix isn't correct.

michaelvanham:
Looks great, thank you very much Michalis!

I had a feeling my fix wasn't quite correct, but I hoped it would help you find the true fix.

PR #162, opened 2020-06-27
Fix merge conflict

eugeneloza:


eugeneloza:
Ok, It won't even show me the conflicting diff because it wants to commit the patch to `master`, not in `castle-editor-undo` and doesn't seem to give me any other option... Ok, then it seems like the only way GitHub will allow me at least to see what's wrong is by creating a pull request to `master`.

PR #166, opened 2020-07-08
Add support for Spine weighted mesh

Kagamma:
This PR is an effort to add support for one of Spine's major features: Weighted mesh.

Things that have been done so far:
- Add deprecated warning to skinnedmesh attachment.
- Support for weighted mesh, by interpolating values between bone timelines.
- Make weighted mesh work with deform timeline.
- Additional frames for better transition, this will make animation more accurate in certain situation (rotation and non-linear interpolation types) in exchange for longer load time and more memory consumption.
- Non-linear interpolation type (Bezier curve) for weighted mesh (both bone timeline and deform timeline).

Limitations:
- Due to the way we handle weighted mesh (by baking every transformations, including bones, into TCoordinateInterpolatorNode), we cannot have a mix of weighted mesh attachments and other types of attachment in the same slot.

Test models: [weighted_mesh.zip](https://github.com/castle-engine/castle-engine/files/4977810/weighted_mesh.zip)

Video: https://youtu.be/AgnwHuohewU

### DIFF
--- a/packages/castle_base.lpk
+++ b/packages/castle_base.lpk
@@ -103,7 +103,7 @@ This is the same license as used by Lazarus LCL and FPC RTL.
 See https://castle-engine.io/license.php for details.
 "/>
     <Version Major="6" Minor="5"/>
-    <Files Count="693">
+    <Files Count="694">
       <Item1>
         <Filename Value="../src/3d/castleboxes.pas"/>
         <UnitName Value="CastleBoxes"/>
@@ -279,6 +279,10 @@ See https://castle-engine.io/license.php for details.
         <Filename Value="../src/base/castleinterfaces.pas"/>
         <UnitName Value="CastleInterfaces"/>
       </Item43>
+      <Item44>
+        <Filename Value="../src/x3d/castleloadgltf.pas"/>
+        <UnitName Value="CastleLoadGltf"/>
+      </Item44>
       <Item45>
         <Filename Value="../src/base/castleinternalgzio.pas"/>
         <UnitName Value="CastleInternalGzio"/>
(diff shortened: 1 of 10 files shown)


Kagamma:
Weighted mesh with deform timeline now work in latest commit.

Kagamma:
Latest commit allows to use non-linear interpolation type (Bezier curve) for weighted mesh.

Kagamma:
The PR is ready to be merged. @michaliskambi please help me review it when you have time :)

michaliskambi:
@Kagamma Thank you! I'm very sorry for such delay with this. 

I looked at the code, and it all seems good. I will want to understand the bone backup/restore mechanism better, but that's on me.

Questions:

1. I see you removed support for region type `skinnedmesh`. Is it possible to keep it supported? To keep the old models, exported from old Spine versions, working? 

    It can of course make a warning like _"Region type skinnedmesh may not be supported in future CGE versions, please export your model again from latest Spine version._ (skinnedmesh already makes a warning _"Spine region type "skinnedmesh" is for now not deformed during animation...."_, so we would just extend this warning).

    Unless it would be a big pain to keep it supported, I would prefer to not remove support for any feature in JSON file, without having earlier a "deprecation warning" emitted by CGE for some time.

2. You provided me a number of demos and testcases, very very thank you! Can I add them to https://github.com/castle-engine/demo-models ? Mentioning you as an author, of course. The https://github.com/castle-engine/demo-models are by default covered by GNU GPL >= 2 license, but we could change this to any open-source license you prefer. 

    If this is OK for you, but only for a subset of these models, of course that would be fine too. Let me know.

Again very sorry for taking such long time to review. I jump with joy when people contribute to CGE, I'm extremely grateful for this work -> please keep it coming. However I am lately overworked even more than usual, so I'm failing behind all tasks.

Kagamma:
Thank you Michalis.

1: The reason I removed support for skinnedmesh attachment is because:
  - The official Spine runtime removed it for quite awhile (since 2016 I think).
  -  CGE support is incomplete and it's impractical to load those models with current level of support.

But as you said, we should mark it as "deprecated" first and removed it sometimes in the future. Since skinnedmesh shares the same setup code with mesh attachment, I can bring it back without much trouble.

2: Those test models is released under public domain, feel free to add them to demo-models repository :)

michaliskambi:
Thank you enormously! Added examples to demo-models, merged to CGE, added small fix (FPC 3.2.0 on Linux/x86_64 failed with internal error, fortunately it was easy to workaround in https://github.com/castle-engine/castle-engine/commit/1abf3e389b8e1683618874e668d5aa008cc4235c ) ).

I'm testing it now and will announce this feature on CGE news around the weekend.

PR #168, opened 2020-07-10
Castle editor rename in treeview

eugeneloza:
Capability to edit names of components in TreeView in Castle Editor - by clicking or by pressing F2.


tools/castle-editor/code/framedesign.pas
```diff
@@ -1857,6 +1863,29 @@ procedure TDesignFrame.ControlsTreeDragOver(Sender, Source: TObject; X,
   ControlsTree.Invalidate; // force custom-drawn look redraw
 end;
 
+procedure TDesignFrame.ControlsTreeEditing(Sender: TObject; Node: TTreeNode;
+  var AllowEdit: Boolean);
+begin
+  { this event is fired when calling TCustomListView.CanEdit which itself is called in TCustomListView.ShowEditor
+    therefore this event preceeds initializing and showing of the editor }
+  //here we have to "restore" the pure name of the component (without class name) before starting edit
```
michaliskambi, line 1871: Minor: 
- Make a space between `//` and text, so `//here` -> `// here`.
- It is best to merge the 2 comments. I understand your style, that `{ }` is a comment about the method, while `// xxx` is a comment about the subsequent line -- but it's not a standard we follow in CGE.
- Start first comment with upper letter.

I am nitpicking today :) Sorry! But I would prefer to write it like this:

```
  { This event is fired when calling TCustomListView.CanEdit which itself is called in TCustomListView.ShowEditor
    therefore this event preceeds initializing and showing of the editor.

    Here we have to "restore" the pure name of the component (without class name) before starting edit. }
```

tools/castle-editor/code/framedesign.pas
```diff
@@ -1857,6 +1863,29 @@ procedure TDesignFrame.ControlsTreeDragOver(Sender, Source: TObject; X,
...
+  Node.Text := TComponent(ControlsTree.Selected.Data).name;
+end;
```
michaliskambi, line 1873: Should this maybe use `TComponent(Node.Data)` instead `TComponent(ControlsTree.Selected.Data)`?

It seems more reliable to use the argument `Node` as the thing that indicates the component, instead of relaying on `ControlsTree.Selected` (and assuming that this event is always called for selected node). Note that the tree hierarchy also supports selecting multiple components with Ctrl.

Minor: `Name` instead of `name`.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1857,6 +1863,29 @@ procedure TDesignFrame.ControlsTreeDragOver(Sender, Source: TObject; X,
...
+
+procedure TDesignFrame.ControlsTreeEditingEnd(Sender: TObject; Node: TTreeNode;
+  Cancel: Boolean);
+var
+  S: String;
+  Sel: TComponent;
+begin
+  Sel := TComponent(ControlsTree.Selected.Data);
```
michaliskambi, line 1881: Similar comment as the previous:

- Can we use `TComponent(Node.Data)` instead of `TComponent(ControlsTree.Selected.Data)`?

- We prefer CamelCase, so write `Sel.Name` instead of `Sel.name` in the line below.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1933,6 +1962,14 @@ procedure TDesignFrame.ControlsTreeEndDrag(Sender, Target: TObject; X,
   ControlsTree.Invalidate; // force custom-drawn look redraw
 end;
 
+procedure TDesignFrame.ControlsTreeKeyUp(Sender: TObject; var Key: Word;
```
michaliskambi, line 1965: Instead of handling F2 directly, please add a a menu item, with a shortcut F2.

This way, this functionality is "discoverable" by just looking at menu items.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1857,6 +1863,29 @@ procedure TDesignFrame.ControlsTreeDragOver(Sender, Source: TObject; X,
...
+var
+  S: String;
+  Sel: TComponent;
+begin
+  Sel := TComponent(ControlsTree.Selected.Data);
+  S := 'Rename ' + Sel.name + ' into ' + Node.Text;
+  Sel.name := Node.Text;
+  ModifiedOutsideObjectInspector;
+  RecordUndo(S); //It'd be good if we set "ItemIndex" to index of "name" field, but there doesn't seem to be an easy way to
+  Node.Text := ComponentCaption(Sel);
```
michaliskambi, line 1886: As per my comments in "undo" PR, I would suggest to merge `ModifiedOutsideObjectInspector` with `RecordUndo`, so 

```
ModifiedOutsideObjectInspector;
RecordUndo(S);
```

-> would be rather 

```
ModifiedOutsideObjectInspector(S);
```

I would also suggest to rename local variable `S` -> `UndoComment` (more descriptive).

tools/castle-editor/code/framedesign.pas
eugeneloza: Fixed in e7b77931ce5cd2af43489866d6a488af2ee0b311

tools/castle-editor/code/framedesign.pas
eugeneloza: I've changed to `Node.Data`, and it seems to work without any issues, so indeed this is more correct way of handling this. Also tested for renaming by pressing F2 and clicking - both work as expected and correctly.
Done in 92d304b2b6020559f9e91ebc16b3b473860afbd9

tools/castle-editor/code/framedesign.pas
eugeneloza: Done in 112b252dba6a348ee86456ee5ec06bd8383b68dc

tools/castle-editor/code/framedesign.pas
eugeneloza: > merge ModifiedOutsideObjectInspector with RecordUndo
That would be done as a separate PR, as we've discussed yesterday.

"S" fixed in 112b252dba6a348ee86456ee5ec06bd8383b68dc.

tools/castle-editor/code/framedesign.pas
eugeneloza: I wonder where should we add this new menu item? To Design? Or maybe we should split Design into Edit, Create and Insert?

Clipboard 1

I've added it to Design before Undo in 5e829e4874ec94db984887a6440472ce9be98994 but we might want to change it.

michaliskambi (approved):
Looks all good, thank you!

michaliskambi:
Note that I removed undefined `OnKeyUp` definition (`OnKeyUp = ControlsTreeKeyUp`) from LFM, https://github.com/castle-engine/castle-engine/commit/550b54cf80aa1580965230acc8b3869f18f51995

eugeneloza:
Ah, I've forgotten about it - sorry. I wonder why didn't I see it when testing? Thanks for fixing!

PR #171, opened 2020-07-18
Castle editor right click in treeview

eugeneloza:
Add a useful popup menu to TreeView.


tools/castle-editor/code/framedesign.pas
```diff
@@ -1970,6 +2027,39 @@ procedure TDesignFrame.ControlsTreeKeyUp(Sender: TObject; var Key: Word;
       ControlsTree.Selected.EditText;
 end;
 
+procedure TDesignFrame.ControlsTreeMouseUp(Sender: TObject;
+  Button: TLclMouseButton; Shift: TShiftState; X, Y: Integer);
+var
+  Sel: TComponent;
+begin
+  if (Button = TLclMouseButton.mbRight) then
```
michaliskambi, line 2035: Instead of detecting in `OnMouseUp` event the right-click, it seems cleaner to assign the popup menu to `PopupMenu` property of the `ControlsTree` .

Then some stuff would be handled automatically. Detecting right-click in our code would not be necessary, setting `MenuTreeView.PopupComponent` would not be necessary, calling `MenuTreeView.PopUp` would not be necessary.

tools/castle-editor/code/framedesign.pas
```diff
@@ -830,6 +847,44 @@ constructor TDesignFrame.Create(TheOwner: TComponent);
     Result.ShowGutter := false;
   end;
 
+  //This is a copy from FormProject - TODO: reuse the code; do not copy it
```
michaliskambi, line 850: Indeed I would ask to make it a single routine, used both here and in FormProject. It can be placed e.g. in `EditorUtils` unit.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1970,6 +2027,39 @@ procedure TDesignFrame.ControlsTreeKeyUp(Sender: TObject; var Key: Word;
...
+begin
+  if (Button = TLclMouseButton.mbRight) then
+  begin
+    Sel := SelectedComponent;
+    MenuTreeViewItemDuplicate.Enabled := Sel <> nil;
+    MenuTreeViewItemCopy.Enabled := Sel <> nil;
+    MenuTreeViewItemDelete.Enabled := Sel <> nil;
+    if (Sel is TCastleUserInterface) or ((Sel = nil) and (DesignRoot is TCastleUserInterface)) then
+    begin
+      MenuTreeViewItemAddUserInterface.Visible := true;
```
michaliskambi, line 2043: Best to toggle both `Enabled` and `Visible`, using `SetEnabledExists` routine.

Feel free to rename `SetEnabledExists` to `SetEnabledVisible` BTW if you consider it better, it seems better to me now :)

Reason: At least in case of other LCL controls, making control invisible is independent from making it disabled. Invisible-but-enabled control can still be activated by a key shortcut.

eugeneloza: Done in f578d4f183c8b3b97d94e2f49401cfae4d4b67cb

tools/castle-editor/code/framedesign.pas
eugeneloza: Took me some time to find our conversation on the topic on 28 July 2020:

>    Instead of detecting in OnMouseUp event the right-click, it seems cleaner to assign the popup menu to PopupMenu property of the ControlsTree .

It throws SIGSEGVs for some reason, some variables are still uninitialized when PopupMenu is called.

> Hm, this is something I would investigate then more. The code will be simpler if we don't manually detect mbRight.

But it doesn't work :) I didn't investigate this bug too deeply, but AFAIR it seems like the menu items.visible isn't available at that moment (because the popup isn't shown yet, so there is no menu instance) and causes a SIGSEGV.

> Which would be solved by my suggestion with actions :slight_smile: You would then control actions, and no need to change menu items.

tools/castle-editor/code/framedesign.pas
eugeneloza: Done in cb1dd93d27cc81e2cbc992d6408480aaffda8813 , 8f8304f50948a03fb2bd46c8587cdd1583d74aba and 395797de8532d345622a9faa706040b408af99c9

tools/castle-editor/code/editorutils.pas
```diff
@@ -144,12 +150,21 @@ procedure SetEnabledExists(const C: TControl; const Value: Boolean);
...
 
 uses SysUtils, Dialogs, Graphics, TypInfo,
   CastleUtils, CastleLog,
+  CastleComponentSerialize, CastleUiControls, CastleCameras, CastleTransform,
   ToolCompilerInfo;
 
+procedure TMenuItemHelper.SetEnabledVisible(const Value: Boolean);
+begin
+  Visible := Value;
+  Self.Enabled := Value;
```
michaliskambi, line 165: Write just `Enabled`, not `Self.Enabled`. It seems inconsistent you use `Visible` but `Self.Enabled` :)

tools/castle-editor/code/framedesign.pas
michaliskambi: I admit I'd like to require this change. That is, do not detect `mbRight`, instead use LCL `PopupMenu` property to assign the context menu to the control.

Manually detecting `mbRight` in `OnMouseUp` feels to me like a hackish way of invoking a popup menu. And e.g. you probably don't handle the "menu key" ( https://en.wikipedia.org/wiki/Menu_key ) this way. Making it cleaner would also mean you can remove `TLclMouseButton`.

You can use the `OnPopup` event to set initial enabled / visible properties, http://docs.embarcadero.com/products/rad_studio/delphiAndcpp2009/HelpUpdate2/EN/html/delphivclwin32/Menus_TPopupMenu_OnPopup.html . The menu should be fully initialized at this point (even more than in `TDesignFrame.ControlsTreeMouseUp` :) ).

michaliskambi (changes requested):
Thank you! I still have 2 requests, in particular I admit I want to push my previous request to *not* detect `mbRight`, instead use `PopupMenu` property. And use `OnPopup` event to set enabled/visible.

I would also ask to look at LCL actions. While I don't require for this  PR to remake this into actions, but actually actions may make it even easier in case of this PR. You would not need `OnPopup` event then at all, you would have actions, that have enabled/visible state, and these actions would control both menu menu and popup menu.

We talked about actions with Andrzej yesterday, and we both agree that it would be nice to use them in CGE editor :)

tools/castle-editor/code/editorutils.pas
eugeneloza: Yes, I wanted to get some completion hint, but then forgotten to delete the Self prefix. Fixed in 9de9cd08c6886184e18f5ca81d866b4b8be0ed0e

tools/castle-editor/code/framedesign.pas
eugeneloza: Ok, I have no idea what was wrong before, but now it just works. No SIGSEGVs whatsoever. All fixed, workarounds removed.

eugeneloza:
All seems to be fixed (without actions) and merge conflict solved.

michaliskambi (approved):
All perfect. Thank you very much, and sorry for taking so long to accept! Merging.

PR #181, opened 2020-08-15
Screen Space Reflections (SSR)

Kagamma:
This PR adds SSR screen effect to the engine.

The shader is based on Riccardo Balbo's implementation of SSR: https://github.com/riccardobl/SimpleSSRShader

TCastleViewport have 2 new properties:
- ScreenSpaceReflections: Enable(true)/Disable(false) SSR. Default is false.
- ScreenSpaceReflectionsSurfaceGlossiness: Adjust default surface glossiness (0..1). Default is 0.5.

Limitations:
- Since we lack various information about the scene (mainly because the engine doesn't have Deferred Rendering pipeline), the whole scene is reflectable - no bump mapping, roughness or metallic check.
- No filters (I can't figure out how best to implement it without resorting to low-level APIs).

Some screenshots:
image1
image
image4
image2

### DIFF
--- a/src/x3d/opengl/castleviewport.pas
+++ b/src/x3d/opengl/castleviewport.pas
@@ -333,7 +333,7 @@ type
     const
       DefaultScreenSpaceAmbientOcclusion = false;
       DefaultScreenSpaceReflections = False;
-      DefaultScreenSpaceReflectionsSurfaceGlossiness = 0.25;
+      DefaultScreenSpaceReflectionsSurfaceGlossiness = 0.5;
       DefaultUseGlobalLights = true;
       DefaultUseGlobalFog = true;
       DefaultShadowVolumes = true;
--- a/src/x3d/opengl/castleviewport.pas
+++ b/src/x3d/opengl/castleviewport.pas
@@ -106,6 +106,10 @@ type
         Viewport: TCastleViewport;
         function SetupUniforms(var BoundTextureUnits: Cardinal): Boolean; override;
       end;
+      TSSRScreenEffect = class(TGLSLScreenEffect)
+        Viewport: TCastleViewport;
+        function SetupUniforms(var BoundTextureUnits: Cardinal): Boolean; override;
+      end;
     var
       FCamera: TCastleCamera;
       FRenderParams: TManagerRenderParams;
@@ -134,6 +138,11 @@ type
       SSAOShader: TSSAOScreenEffect;
       SSAOShaderInitialized: Boolean;
 
+      FScreenSpaceReflections: Boolean;
+      SSRShader: TSSRScreenEffect;
+      SSRShaderInitialized: Boolean;
+      FScreenSpaceReflectionsSurfaceGlossiness: Single;
+
       FOnCameraChanged: TNotifyEvent;
       { Renderer of shadow volumes. You can use this to optimize rendering
         of your shadow quads in RenderShadowVolume, and you can read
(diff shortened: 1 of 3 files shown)


michaliskambi:
This looks great. Thank you!

My main concern is that I would not like to implement it by another Boolean variable inside `TCastleViewport` and I would not like it to add this inside (already large) `CastleViewport` unit. Which is all my fault -- `ScreenSpaceAmbientOcclusion` is implemented as a Boolean variable like that, and I didn't propose a better design for it yet.

My idea is to remake `ScreenSpaceAmbientOcclusion`, and also `ScreenSpaceReflections` now, as components, that could reside in a different unit (like `CastleViewportScreenEffects`? ). 

The TCasleScreenEffects already offers a flexible API to add/remove effects. 

We should not need to modify a single `TCastleViewport.ScreenEffectsCount` or `TCastleViewport.ScreenEffectsNeedDepth` to add new effects. Likewise, `TCastleViewport.SSRShaderInitialize` should not be necessary.

We should have a component like `TScreenSpaceReflections` with a property like `SurfaceGlossiness: Single`.

This would also enable natural way to control specific screen effects parameters by properties, and expose them as published for CGE editor. IOW, allow adding screen effects from editor.

So, all those comments kind of wait on me, to propose a better way to design "standard" screen effects inside CGE, like screen-space ambient occlusion and now screen-space reflections. Give me a bit of time to try to do it (if I will fail, I'll merge the current implementation, and will redesign it later :) ).

michaliskambi (approved):

michaliskambi:
First of all, sorry (again) for taking so long to review this. I think this is the 3rd time where your PR took me a looong time to apply, simply because of my lack of time. Sorry about this, and (this is important!) I hope that my long review times did not discourage you from doing more work on CGE. Over the years, you provided a number of big CGE features, so I hope that you will keep them coming :)

And I hope that my review times will get better.

I'm applying it now, as everything is cool -- the remaining work is on my side to

1. propose a new API to add screen effects inside CGE. See Trello ticket: https://trello.com/c/2hWdcmEO/97-make-screen-effects-like-ssao-ssr-and-more-as-components . When this happens, I'll remake SSR to be a nice component.

2. new GLSL functions to query screen point without any divisions. See Trello ticket: https://trello.com/c/wmKO7kn3/96-make-glsl-api-for-screen-effects-that-avoids-dividing-multiplying-by-screen-sizes

michaliskambi:
1. I solved https://trello.com/c/wmKO7kn3/96-make-glsl-api-for-screen-effects-that-avoids-dividing-multiplying-by-screen-sizes faster than I thought :)

    We now have `screenf_01_position` (previous `tex_coord_frag`, only renamed) and `screenf_01_get_color(...)` and `screenf_01_get_depth(...)`.

    Documented on https://castle-engine.io/x3d_extensions_screen_effects.php , and will be added to news soon.

    I updated GLSL code of SSR to use them. I needed to update this code a bit anyway, as

      ```
      float depth = texture2D(screen_depth, pos)
      ```
      
    was reported as invalid by my current OpenGL (renderer: Mesa DRI Intel(R) UHD Graphics 630 (Coffeelake 3x8 GT2), OpenGL version: 3.0 Mesa 18.3.6).
      
    It makes sense, as `texture2D` returns `vec4`, so it should be accessed like `texture2D(...).r` to get float. `screenf_01_get_depth` is doing that now.

2. I added option to play with it in view3dscene, "View -> Screen Effects -> Screen Space Reflections". It works cool!

scene_0
scene_0_ssr

PR #205, opened 2020-09-30
53 fix random test

eugeneloza:
Fix https://trello.com/c/QYekAxqL/53-rndrandom-can-sometimes-return-0

### DIFF
--- a/tests/code/testcastlerandom.pas
+++ b/tests/code/testcastlerandom.pas
@@ -41,21 +41,22 @@ begin
   // Zero seed hash
   AssertEquals(%11100000010001101011001110011100, StringToHash('String to hash test'));
   // Seeded hash
-  AssertEquals(%01111101100111100000101011011100, StringToHash('String to hash test',$9747b28c));
+  AssertEquals(%01111101100111100000101011011100, StringToHash('String to hash test', $9747b28c));
 end;
 
 procedure TTestCastleRandom.TestRandom;
 const
-  NTests = 10000000;
+  NTests = 1000000;
+  SumTests = 1000000;
 var
   Rnd: TCastleRandom;
-  i: Integer;
+  I: Integer;
   Sum: Double;
 begin
   Rnd := TCastleRandom.Create;
 
   //test that random is in 0..1 limits
-  for i := 0 to NTests do
+  for I := 0 to NTests do
     AssertTrue('Rnd.Random <= 1', Rnd.Random <= 1);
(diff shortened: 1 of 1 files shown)


tests/code/testcastlerandom.pas
```diff
@@ -41,33 +41,34 @@ procedure TTestCastleRandom.TestHash;
...
   Sum: Double;
 begin
   Rnd := TCastleRandom.Create;
 
   //test that random is in 0..1 limits
-  for i := 0 to NTests do
+  for I := 0 to NTests do
     AssertTrue('Rnd.Random <= 1', Rnd.Random <= 1);
-  for i := 0 to NTests do
-    AssertTrue('Rnd.Random > 0', Rnd.Random > 0);
```
michaliskambi, line 61: I would say to keep the test, just change `Rnd.Random > 0` to `Rnd.Random >= 0`.

michaliskambi (changes requested):
All makes sense (we just allow `Random` to be equal = 0, which is cool since our docs never claimed it is strictly `> 0`), and thanks for the formatting improvements too.

My only note is that I see you simply removed `Rnd.Random > 0` test. I would say to leave it, with relaxed `Rnd.Random >= 0` condition. Of course the new test of `Random32bit` is fine too.

tests/code/testcastlerandom.pas
eugeneloza: `Rnd.Random32bit` below actually does this just much better (see Trello issue), i.e. practically `Rnd.Random := (Rnd.Random32bit shl 1) / MaxInt;` - this "`shl 1` bug" was the cause of this test failing once in 400 tests as `1 shl 1 = 0`.

michaliskambi: Sure, I'm all after having the new `Rnd.Random32bit` test too :)

I'm just saying to have the `Rnd.Random >= 0` too. Just in case its implementation changes some day.

eugeneloza: Well, I don't think this test serves any useful purpose, but of course we can have it :) Added in bb0e3b70bf2971fb1dd92f6af1d7c922c55d56ab.

michaliskambi: For me, the test confirms that our understanding of `Random` method is correct :) Even if it's a simple routine, now I don't need to look at the code to claim with confidence "it is always >= 0".

michaliskambi (approved):

PR #207, opened 2020-10-01
Rename deprecated key constants

eugeneloza:
* Deprecate all K_Xxxxxx keys.
* Hopefully properly rename all deprecated key constants to their valid values everywhere: engine sources, tests, examples, etc.


michaliskambi (approved):
Cool! Thank you!

PR #208, opened 2020-10-03
Fix Frogger3d and castle-game regression

and3md:
This PR fix Player recreation issue. 

Trello issues:
- https://trello.com/c/VhnXB66N/60-fix-frogger3d-crashing-regression
- https://trello.com/c/oXmF5Pda/61-castle-game-regression

This fix simply reinitialize the camera/navigation after changing the Player. This works ok for level restart. 

I played with Frogger a bit longer and made an extra fix to fix problems when destroying an object causes changes to the camera settings, and then it is notified of these changes. What can cause SEGFAULT. Simply way to see this is change the order of creation/adding Player, and Level to scene (in Frogger3D) like this: 

```
  FpsPlayer := TPlayer.Create(SceneManager);
  //SceneManager.Items.Add(FpsPlayer);
  //SceneManager.Player := FpsPlayer;
  FpsPlayer.Blocked := true; // do not allow to move player in this game

  SceneManager.LoadLevel('1');
  SceneManager.Items.Add(FpsPlayer);
  SceneManager.Player := FpsPlayer;
```
I think that such behaviour is even easier to meet in larger projects and not only when using Player/Level api.

### DIFF
--- a/src/3d/castletransform.pas
+++ b/src/3d/castletransform.pas
@@ -2900,9 +2900,30 @@ end;
 procedure TCastleTransform.CameraChanged(const ACamera: TCastleCamera);
 var
   I: Integer;
+  TransformToNotify: TCastleTransform;
 begin
   for I := 0 to List.Count - 1 do
-    List[I].CameraChanged(ACamera);
+  begin
+    TransformToNotify := List[i];
+    { Checking the component status avoids informing an object that
+      changes or initiates change of camera settings during self destruction.
+      For example this situation may appear when TPlayer delete
+      causes reinitialization of the camera in TLevel).
+
+      To do this, see Frogger3d with the changed order of Level and Player
+      creation (Level first, Player next) and play two times:
+
+      FpsPlayer := TPlayer.Create(SceneManager);
+      FpsPlayer.Blocked := true; // do not allow to move player in this game
+      SceneManager.LoadLevel('1');
+      SceneManager.Items.Add(FpsPlayer);
+      SceneManager.Player := FpsPlayer;
+
+      I think that in the case of user games there may be more such situations,
+      so it is worth checking this. }
+    if not (csDestroying in TransformToNotify.ComponentState) then
+      TransformToNotify.CameraChanged(ACamera);
+  end;
 end;
 
 procedure TCastleTransform.CameraChanged(const ACamera: TCastleNavigation);
--- a/src/3d/castletransform.pas
+++ b/src/3d/castletransform.pas
@@ -2900,30 +2900,9 @@ end;
 procedure TCastleTransform.CameraChanged(const ACamera: TCastleCamera);
 var
   I: Integer;
-  TransformToNotify: TCastleTransform;
 begin
   for I := 0 to List.Count - 1 do
-  begin
-    TransformToNotify := List[i];
-    { Checking the component status avoids informing an object that
-      changes or initiates change of camera settings during self destruction.
-      For example this situation may appear when TPlayer delete
-      causes reinitialization of the camera in TLevel).
-
-      To do this, see Frogger3d with the changed order of Level and Player
-      creation (Level first, Player next) and play two times:
-
-      FpsPlayer := TPlayer.Create(SceneManager);
-      FpsPlayer.Blocked := true; // do not allow to move player in this game
-      SceneManager.LoadLevel('1');
-      SceneManager.Items.Add(FpsPlayer);
-      SceneManager.Player := FpsPlayer;
-
-      I think that in the case of user games there may be more such situations,
-      so it is worth checking this. }
-    if not (csDestroying in TransformToNotify.ComponentState) then
-      TransformToNotify.CameraChanged(ACamera);
-  end;
+      List[I].CameraChanged(ACamera);
 end;
 
 procedure TCastleTransform.CameraChanged(const ACamera: TCastleNavigation);
--- a/src/3d/castletransform.pas
+++ b/src/3d/castletransform.pas
@@ -2902,7 +2902,7 @@ var
   I: Integer;
 begin
   for I := 0 to List.Count - 1 do
-      List[I].CameraChanged(ACamera);
+    List[I].CameraChanged(ACamera);
 end;
 
 procedure TCastleTransform.CameraChanged(const ACamera: TCastleNavigation);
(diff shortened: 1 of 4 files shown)


and3md:
BTW. Perhaps other notification loops should also check whether the component is during destruction. I cannot judge it yet. I think it's worth considering.

src/game/castlelevels.pas
```diff
@@ -1311,6 +1190,122 @@ procedure TLevel.SetPlayer(const Value: TPlayer);
     end;
     Viewport.AvoidNavigationCollisions := Value;
   end;
+  { Reintialize camera and navigation only when level was loaded. }
```
michaliskambi, line 1193: Reintialize -> Reinitialize (typo :) )

src/3d/castletransform.pas
```diff
@@ -2900,9 +2900,30 @@ procedure TCastleTransform.VisibleChangeNotification(const Changes: TVisibleChan
...
+
+      FpsPlayer := TPlayer.Create(SceneManager);
+      FpsPlayer.Blocked := true; // do not allow to move player in this game
+      SceneManager.LoadLevel('1');
+      SceneManager.Items.Add(FpsPlayer);
+      SceneManager.Player := FpsPlayer;
+
+      I think that in the case of user games there may be more such situations,
+      so it is worth checking this. }
+    if not (csDestroying in TransformToNotify.ComponentState) then
```
michaliskambi, line 2924: If it is easy, I would suggest to add automatic test for this case, in `tests/code/testcastlelevels.pas`. 

- Make the test, 
- make sure that the test *fails* before the fix, 
- make sure it no longer fails after the fix :)

There's already one test `TTestCastleLevels.TestLoadLevelWithoutGLContext`, so you can see how it works :) The short docs of our testing application is in `tests/README.md`. It is automatically run by Jenkins after every commit, so once something is tested there, we *know* it will stay fixed :)

michaliskambi (changes requested):
Thank you!

I managed to make 2 minor notes (one about typo, the other one to add automatic test), see below.

I also have 1 major note: looking at what `TLevel.LoadCore` does, we need to do more things when player changes:

- `TLevel.LoadCore` calls `Items.Add(Player);`, see there for a large comment explaining why order on `Items` is important. So when changing `Player` in `SetPlayer`, you should 

    1. remove old `Player` value, if non-nil.

    2. add new `Player`, if non-nil, as the 2nd item, like `Items.Add(1, Value)`. `Player` must be after `MainScene` but before other items. It is possible that actually `Player` can be just before all other items, so it is simpler to do `Items.Add(0, Value)` -- but needs testing.

- Actually, above is more complicated, to account for 3rd-person navigation. `InitializeCamera` (both before and after these changes) may call `Items.Add(ThirdPersonNavigation.AvatarHierarchy)` or `Items.Add(ThirdPersonNavigation.Avatar)`. Any one of them may be equal to `Player`, thus Player gets automatically added to `Items`.

- And call `if (Player <> nil) then Player.LevelChanged;` after setting new player.

All in all, I think SetPlayer should be extended to:

- if old Value <> nil then remove old Value from `Items`

- after  calling `InitializeCamera`, do something like

```pascal
if new Player <> nil then
begin
  { Add Player to Items,
    as the first item (see LoadCore for explanation why),
    making sure it's not already there (because InitializeCamera
    may have already added it when UseThirdPerson). }
  if Player.World = nil then
    Items.Insert(0, Player);
  Player.LevelChanged;
end;
```

- Also, `InitializeCamera` should be changed to insert avatar (`Items.Add(ThirdPersonNavigation.AvatarHierarchy)` or `Items.Add(ThirdPersonNavigation.Avatar)`) at proper position. Maybe it should just do `Items.Insert(0, ThirdPersonNavigation.AvatarHierarchy)` or `Items.Insert(0, ThirdPersonNavigation.Avatar)`.

Note: examples/fps_game allows to test both 1st-person and 3rd-person navigation. You can uncomment there `SetupThirdPersonNavigation` to easily see 3rd-person navigation.

and3md:
Thanks for the detailed comment, I think now I understand the sources much better.

> add new `Player`, if non-nil, as the 2nd item, like `Items.Add(1, Value)`. `Player` must be after `MainScene` but before other items. It is possible that actually `Player` can be just before all other items, so it is simpler to do `Items.Add(0, Value)` -- but needs testing.

I found in sources that Player must be after scene (item[1]) because of wall sliding. I think this is because of `TCastleTransform.LocalMoveCollision`. Only item[0] is treated as walls.
 
> Actually, above is more complicated, to account for 3rd-person navigation. `InitializeCamera` (both before and after these changes) may call `Items.Add(ThirdPersonNavigation.AvatarHierarchy)` or `Items.Add(ThirdPersonNavigation.Avatar)`. Any one of them may be equal to `Player`, thus Player gets automatically added to `Items`.

On closer inspection, I think only `AvatarHierarchy` can be used with the `TPlayer` class because `Avatar` is `TCastleScene` and `TPlayer` is not inherited form `TCastleScene`.
 
> Note: examples/fps_game allows to test both 1st-person and 3rd-person navigation. You can uncomment there `SetupThirdPersonNavigation` to easily see 3rd-person navigation.

I made a small improvement on this example. Now you can recreate the player at any time during the game using the F1 key. FPP and TPP camera seems to be working properly. Frogger3d and Castle Game also works OK.

Please wait with the merge until I add the test you requested. Of course, any comments are welcome :)

and3md:
After changes in Player setup in Level I can't reproduce `CameraChanged()` problem so I reverted this change. 

> There's already one test TTestCastleLevels.TestLoadLevelWithoutGLContext, so you can see how it works

I had a minor problem with this test - it doesn't work when I chose only the TestCastleLevels in the FPCUnit interface, so I added a level loading line.
After all, I created a test to add and remove a Player after loading a level. If the problem occurs again, we will know.

michaliskambi (approved):

michaliskambi:
Thank you -- all looks cool! Merging :)

michaliskambi:
I pushed trivial formatting fixes and upgrade `K_F1` (deprecated) -> `keyF`. Tested fps_game (with 1st and 3rd person), frogger, castle-game -- everywhere fixed, nice! Thank you!

I also noticed a new problem in frogger3d. Maybe it is related to this fix, maybe not :) Created new ticket, https://trello.com/c/iz7uvjKN/79-frogger3d-crash-at-exit .

PR #209, opened 2020-10-05
Editor clear warnings button

and3md:
Resolves: https://trello.com/c/p49eYePV/65-clear-button-somewhere-in-the-warnings-panel-in-the-editor

Added Panel with one SpeedButton, also fixed clearing on open/create Design.


michaliskambi (approved):
All looks cool, thank you!

michaliskambi:
Looks like some UI sizes of the frame changed, and the font size of the text "To visually design user interface or 3D/2D game world..." changed. The diff of LFM file indeed had a number of changes, in particular

```
Font.Height = -33
Font.Height = -26
```

for the label with `'To visually design...`.

Attaching screenshots. Unfortunately, it is indeed easy to change sizes of things just by reopening in Lazarus on different system with different theme/scaling.

- Maybe you do not use the same text scaling? See "Contributing" section in https://github.com/castle-engine/castle-engine/tree/master/tools/castle-editor#contributing . Looking at diff, the line `DesignTimePPI = 120` disappeared.

- Maybe it is because of our Lazarus versions differences? I use older Lazarus 2.0.8 now, while I see you used later 2.0.11.0. If this is the culprit, I can upgrade.

It's not a problem for now -- I reverted the LFM changes, and repplied only addition of the new button. So this is just a note for future :), to watch out for this, and figure out "how to cooperate on castle-editor to keep the diffs in LFM files small / non-breaking".

michaliskambi:
I also played around and added a Caption and remade the button to TBitBtn (it looks somewhat better than TSpeedButton, when `AutoSize`). I didn't want to bother you with it, because I wasn't sure whether it's a good idea before I tried it :)

Issue #210, opened 2020-10-10
Cannot find program lazarus


KrzysztofJeszke:
The filename "lazarus" is obselete now, it should either use the executable "lazarus-ide" or "startlazarus".
Otherwise it will try to execute "lazarus" no matter what you input in config

PR #211, opened 2020-10-11
Add lazarus-ide as alternative Lazarus executable name

eugeneloza:
Summary: On Linux (at least on Debian) official Lazarus+FPC binaries no longer create `lazarus` executable, but `lazarus-ide` in `/usr/bin/`.

Tested and working for me, however, might need confirmation from the original reporter.

Should fix https://github.com/castle-engine/castle-engine/issues/210 Also we've been discussing the issue on Discord.


michaliskambi (approved):
Thank you!

PR #213, opened 2020-10-18
76 : Wrap renaming in TreeView in try..except

eugeneloza:
Fixes https://trello.com/c/C66HVb5s/76-bug-assigning-wrong-name-for-the-component-renames-it-in-treeview-and-keeps-the-wrong-name


michaliskambi:
Thank you! Merging. I changed the solution a tiny bit -- instead of try..except, I used try..finally with an additional comment. It seems cleaner (otherwise the code was doing `Node.Text := ComponentCaption(Sel)` is 2 places).

PR #214, opened 2020-10-24
75 bug access violation in castle editor

eugeneloza:
Workaround bug when we may have `ControlsTree.Selected = nil` in `TDesignFrame.PropertyGridModified` under some circumstances

Trello issue: https://trello.com/c/V6v2rBwv/75-bug-access-violation-in-castle-editor

### DIFF
--- a/tools/castle-editor/code/framedesign.pas
+++ b/tools/castle-editor/code/framedesign.pas
@@ -1823,7 +1823,8 @@ var
   UI: TCastleUserInterface;
 begin
   { Workaround possible ControlsTree.Selected = nil when the user deselects
-    the currently edited component by clicking somewhere else }
+    the currently edited component by clicking somewhere else.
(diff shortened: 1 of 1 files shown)


michaliskambi:
Merging, thank you! 

I'm leaving https://trello.com/c/V6v2rBwv/75-bug-access-violation-in-castle-editor open --weirdly enough, I still cannot reproduce the problem, doing the same thing as you show on the video. The solution makes sense of course, so I'm fine to keep it. Still, I would like to figure out why I cannot reproduce.

michaliskambi:
Note that your PR applies the commit to `castle-editor-right-click-in-treeview` branch, not to `master`, since I guess you derived the `75-bug-access-violation-in-castle-editor` branch from `castle-editor-right-click-in-treeview`.

This is OK, just noting that now it waits for #171 (going there soon) before it gets to `master`.

PR #218, opened 2020-10-27
New sprites - Loading Starling XML, Cocos2d Plist (format 2 and 3) and usual images directly to TCastleScene

and3md:
This is first part of https://trello.com/c/8h3vWwaF/64-new-approach-for-sprites-to-load-them-easily-into-tcastlescene

Code bases on spritesheet-to-x3d tool.

## Starling XML
As agreed, the files must have their extension changed from `.xml` to `.starling-xml`.

### Design assumptions 
To keep the speed of loading files as high as possible, I assumed that subsequent frames of the same animation in the XML file appear in the correct order. From my observations, it appears that the files look like this.

### Options
Can be set in anchor.

Option | Description
-------- | ---------------
`fps` | frame per second
`anim-naming` | subtextures names in starling file can be named freely, but in the case of our loader, we have to define what is the next frame of the animation and what should be recognized as a separate animation <br><br> Values: <br> - `strict-underscore`- default behavior treats as animation frames only those subtextures whose names ends with an underscore followed by a number. In this case, "walk_01", "walk_02" will be recognized as next frames of the same animation "walk", but "item1", "item2" will be treated as separate entities. <br>- `trailing-number` - in many cases, the consecutive frames of one animation are named without underscore, eg "walk1", "walk2". To load such subtextures as one animation use this option. 

### Mime type
Based on https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types I created `image/starling-texture-atlas`. Most likely, we should change this to something more appropriate. On the one hand, it is a texture atlas so it should be in `image` group, on the other hand, it needs processing in the appropriate application, so maybe it should be changed to `application` group. Maybe should start form x like `image/x-starling-texture-atlas` or `application/x-starling-texture-atlas`.

### Default frame naming
Kenney packs like https://kenney.nl/assets/toon-characters-1 uses `trailing-number` convention (one `walk` animation has frames named as `walk1` `walk2` ...) in other hand tutorial form Starling wiki [Hungry Hero](https://wiki.starling-framework.org/tutorials/start) uses `strict-underscore` convention. See file here https://github.com/hsharma-design/Hungry-Hero/blob/master/media/graphics/mySpritesheet.xml 

I think `strict-underscore` convention is more clear. You can do one frame "animations" like `idle1` `idle2` and multi frame animation like `die1_001` `die1_002` and `die2_001` `die2_002` and this looks more obvious (for me).

### TO DO
Find and test more files. 

## Cocos 2d
Here I left the standard extension - `.plist`. Maybe should be changed.

### Design assumptions 
To keep the speed of loading files as high as possible, I assumed that subsequent frames of the same animation in the dictionary file appear in the correct order. From my observations, it appears that the files look like this.

### Options
Can be set in anchor.

Option | Description
-------- | ---------------
`fps` | frame per second

### Format Versions
There are three format versions. I implemented version 2 and 3. First version is very old. 

Some code are doubled in `ParseFrameDictionaryFormat2` and `ParseFrameDictionaryFormat3` Currently I have not decided whether they should be merged or anchor calculate code separated. Keep in mind that format version 3 supports also PolygonSprites so there can be more changes in the future. I think some data types has been changed too from Integer to Single.

### Mime type
A lot of sources use `application/xml` for `.plist` but the second most popular option is `application/x-plist` so I used `application/x-plist`.

### TO DO
Find and test more files. 

## Load usual images directly to TCastleScene

In this case you can load entire image or chosen rect (left, bottom, width, height) 

### Options
Can be set in anchor.

Option | Description
-------- | ---------------
`left` | left coordinate of chosen rectangle in px
`bottom` | bottom coordinate of chosen rectangle in px
`width` | width of chosen rectangle in px
`height` | height of chosen rectangle in px

## Code structure

All imports have similar functions to prepare the X3D model (like `PrepareX3DRoot`, `CalculateFrameCoords`, `PrepareShape` `AddFrameCoords` `AddAnimation`) I did it especially so that after getting to know one import you can easily find yourself in others. However, I was not tempted to introduce a superior class because it is quite easy to imagine introducing changes in the parent class for one of the formats, which would "accidentally" destroy another import. I don't believe that a person changing this class will test all inheriting imports ;) When we add PolygonSprite a lot will change in Cocos2d imports and the code similarity will disappear.
IOW I think separating each import from the others is the best idea/approach.

## Play animation example

In play animation example I added loading Starling, Cocos2d and rectangle from PNG file examples. 
obraz

Starling load options are also in this example GUI so there we can easily test other files with different animation naming.convention checked.


michaliskambi:
Thank you! Initial comments (not yet reading the diff):

As for MIME types:

- I don't see "sprite sheet" as merely an image or even a texture atlas (even if Starling calls them a texture atlas), I see it as a description of animations.

  So I would say to go with

    `application/x-starling-sprite-sheet`
    `application/x-cocos2d-sprite-sheet`

  This is consistent with other MIME types for models that we had to invent, looking at `URIMimeType` implementation. In general it seems that people are using `image/`, `model/`, `application/` prefix without any consistency for models :)

As for Cocos2d extension:

- I would say to use and advise `.cocos2d-plist`, but also handle `.plist`. The plist is an Apple serialization format -- it's similarly "generic" as xml or json. (And yes, this raises the question do we want to one day recommend `.spine-json` for Spine, instead of just `.json`; to be decided at some point :) )

As for play_animation:

- In the UI: Use TCastleVerticalGroup with a border and padding for all "Load sample..." buttons. And as the fist element, add TCastleLabel with caption "Load sample...". And then remove from button's captions "Load sample" prefix. This can be postponed after PR, as it will be trivial to do in CGE editor later.

- In the UI: I would make some captions longer but with explicit info. So:

    - "Starling load options" -> "Starling Sprite Sheet load options:"
    - "Load sample 2D Starling" -> "Starling Sprite Sheet (2D)"
    - "Load sample 2D Cocos2d" -> "Cocos2d Sprite Sheet (2D)"
    - "Load sample 2D Spine" -> "Spine Skeletal Animation (2D)"
    - "Load sample 3D model" -> "glTF Skinned Animation (3D)"
    - "Load sample 2D image" -> "Image (2D)"

- After merging this PR I would convert this application to have UI done in CGE editor (but this is outside of this PR, just making a note here).

michaliskambi:
I created a documentation for this:

https://github.com/castle-engine/castle-engine/wiki/Sprite-sheets
https://github.com/castle-engine/castle-engine/wiki/Images

TODO:

* On https://github.com/castle-engine/castle-engine/wiki/Sprite-sheets , check that this sentence is true:

    """
    So e.g. walk1 and walk2 will work (resulting in animation name walk) just like walk_1 and walk_2 (underscore is optional, but still stripped, so it also results in animation name walk).
    """

* On https://github.com/castle-engine/castle-engine/wiki/Images , check that this sentence is true:

    """
    Note that if you specify area that is too large (outside of the actual image area), the excessive pixels will show the clamped image border (the image is simply rendered as a texture with repeat=FALSE).
    """

* Link to these pages from https://castle-engine.io/documentation.php sidebar, and https://github.com/castle-engine/castle-engine/wiki sidebar, once PR is applied

src/x3d/x3dloadinternalstarling.pas
```diff
@@ -0,0 +1,611 @@
...
+end;
+
+procedure TStarlingTextureAtlasLoader.AddRoutes(
+  const TimeSensor: TTimeSensorNode;
+  const CoordInterp: TCoordinateInterpolatorNode;
+  const TexCoordInterp: TCoordinateInterpolator2DNode);
+var
+  R1, R2, R3, R4: TX3DRoute;
+begin
+  { Create routes. }
```
michaliskambi, line 347: I added a bit friendlier overload to add routes some time ago (I know, many loaders have not yet updated to using it).

Instead of

```
R1 := TX3DRoute.Create;
...
R1.SetSourceDirectly(TimeSensor.EventFraction_changed);
R1.SetDestinationDirectly(CoordInterp.EventSet_fraction);
...
FRoot.AddRoute(R1);
```

use 

```
FRoot.AddRoute(TimeSensor.EventFraction_changed, CoordInterp.EventSet_fraction);
```

This is shorter and usually easier to read.

src/x3d/x3dloadinternalstarling.pas
```diff
@@ -0,0 +1,611 @@
...
+  end;
+
+  FAnimationList.Add(AnimationName);
+  Result := true;
+end;
+
+
+constructor TStarlingTextureAtlasLoader.Create(const URL: String);
+begin
+  FURL := URL;
```
michaliskambi, line 382: Same as before, call `inherited Create` at the beginning of the constructor.

Also, place `Destroy` in the implementation right after `Create`, so move `Destroy` implementation a bit higher. We follow this convention in CGE: first `Create`, then `Destroy`, then the rest of methods in completely random order :)

src/x3d/x3dloadinternalstarling.pas
```diff
@@ -0,0 +1,611 @@
...
+begin
+  ReadImportSettings;
+
+  FRoot := nil;
+  Doc := nil;
+  try
+    try
+      FRoot := TX3DRootNode.Create;
+      Doc := URLReadXML(FURL);
+      Node := Doc.FindNode('TextureAtlas');
```
michaliskambi, line 412: Is `TextureAtlas` the top-level element (looks like it, from example)? If yes, then do instead

```
Check(Doc.DocumentElement.TagName8 = 'TextureAtlas', 'Root of Starling sprite sheet file must be <TextureAtlas>');
```

src/x3d/x3dloadinternalstarling.pas
```diff
@@ -0,0 +1,611 @@
...
+      if (Node = nil) or (Node.NodeType <> ELEMENT_NODE) then
+        raise EInvalidStarlingXml.CreateFmt('Invalid Starling XML file "%s" - TextureAtlas node not found.', [FDisplayURL]);
+      AtlasNode := Node as TDOMElement;
```
michaliskambi, line 415: If this is root, then just `AtlasNode := Doc.DocumentElement;`  without any typecast should be OK.

src/x3d/x3dloadinternalstarling.pas
```diff
@@ -0,0 +1,611 @@
...
+        AnimationNameLength := Length(AnimationName);
+        if (AnimationNameLength > 1) and (AnimationName[AnimationNameLength] = '_') then
+          delete(AnimationName, AnimationNameLength, 1);
+      end;
+  end;
+
+  if AnimationName = '' then
+  begin
+    WritelnWarning('Starling', 'Incorrect animation name (%s), I set to "unknown"',
+    [SubTextureName]);
```
michaliskambi, line 548: Indent the line `[SubTextureName]);` with 2 spaces, i.e.

```pascal
WritelnWarning('Starling', 'Incorrect animation name (%s), I set to "unknown"',
  [SubTextureName]);
```

Another way to write some Format-like lines, that I also use (but both above and below are correct):

```pascal
WritelnWarning('Starling', 'Incorrect animation name (%s), I set to "unknown"', [
  SubTextureName
]);
```

I found it is easier to add / remove / comment-out particular parameters is the last form.

michaliskambi (changes requested):
Thank you enormously, and sorry for the long delay! Reviewed. 

As expected in a large PR, I found a number of small places to make some small notes and generally nitpick :) In general, I think all my comments are minor -- I like the design of everything.

src/x3d/x3dloadinternalstarling.pas
```diff
@@ -0,0 +1,611 @@
...
+        end;
+      end;
+    anTralingNumber:
+      begin
+        RemoveTrailingChars(AnimationName, ['0'..'9']);
+
+        { Remove trailing underscore if there is more than one char }
+        AnimationNameLength := Length(AnimationName);
+        if (AnimationNameLength > 1) and (AnimationName[AnimationNameLength] = '_') then
+          delete(AnimationName, AnimationNameLength, 1);
```
michaliskambi, line 541: Instead of the lines above, use `AnimationName := RemoveSuffix('_', AnimationName);` (it will do nothing if no such suffix).

michaliskambi (changes requested):
Found one more minor thing.

michaliskambi:
P.S. Once you rebase to the newest master, the Jenkins build of this PR should succeed. (you will see a "check" at "continuous-integration/jenkins/pr-merge" in the GitHub dialog at the bottom of this PR conversation page).

Right now it fails, only because it also builds view3dscene from master, and newest view3dscene from master already requires `RenderOptions` existence.

src/x3d/x3dload.pas
```diff
@@ -351,13 +334,78 @@ function LoadNode(const URL: string;
...
+    Result := LoadStarlingTextureAtlas(URL)
+  else
+
+  if (MimeType = 'application/x-plist') or
+     (URIMimeType(URIDeleteAnchor(URL, true)) = 'application/x-plist') then
+    Result := LoadCocos2d(URL)
+  else
+
+  { Support for simple graphics images like PNG }
+  if URIMimeExtensions.ContainsKey(ExtractFileExt(URIDeleteAnchor(URL, true))) then
```
michaliskambi, line 348: This checks that URL extension is defined somewhere in our MIME database in `URIMimeExtensions`. 

While only images are registered there now (so in practice this does the job OK) but in general various units (in CGE or user code) could register other file types in `URIMimeExtensions`, not necessarily images.

Rejected solution 1: The right array to browse is `ImageFormatInfos` inside `CastleImage`. Rejected: but it is internal in that unit, and I would prefer to keep it internal (to allow more flexible implementation with registered image formats in the future).

Rejected solution 2: Expose in `CastleImages` a public `ImageMimeExtensions` (function in interface, exposing an internal singleton, created inside the unit). `ImageMimeExtensions` would be populated only in `RegisterMimeTypes` in `src/images/castleimages_file_formats.inc`, with the same content as `URIMimeExtensions`. Rejected: This could be modified by user code.

Good solution: I would say to expose in `CastleImages` dedicated functions for this. They seem to be generally useful, so it makes sense to expose them:

- Expose in `CastleImages` a function `IsImageMimeType(const MimeType: String): Boolean` (that internally accesses `ImageFormatInfos`, iterating over every format, and in each format iterating over every `TImageFormatInfo.MimeTypes` item).

- For the ReadImageExtensions below, similarly: make in `CastleImages` function `AllImagesExtensions` that would return the desired String. I have a sketch of such function:

```pascal
function AllImagesExtensions: String;
var
  Format: TImageFormat;
  ExtIndex: Integer;
begin
  Result := '';
  for Format := Low(Format) to High(Format) do
    for ExtIndex := 1 to ImageFormatInfos[Format].ExtsCount do
      Result := SApppendPart(Result, ';', '*.' + ImageFormatInfos[Format].Exts[ExtIndex];
end;
```

src/x3d/x3dload.pas
```diff
@@ -351,13 +334,78 @@ function LoadNode(const URL: string;
...
   if NilOnUnrecognizedFormat then
     Result := nil
   else
     raise Exception.CreateFmt('Unrecognized file type "%s" for scene "%s"',
       [MimeType, URIDisplay(URL)]);
 end;
 
+function LoadScene_FileFilters: String;
+
+  function ReadImageExtensions: String;
```
michaliskambi, line 361: Instead of this, use `AllImagesExtensions` proposed above.

P.S. For constructions like below, the `SAppendPart` is useful.

michaliskambi (changes requested):
I found one more thing :)

src/x3d/x3dloadinternalimage.pas
```diff
@@ -0,0 +1,271 @@
...
+  FRoot.Meta['generator'] := 'Castle Game Engine, https://castle-engine.io';
+  FRoot.Meta['source'] := ExtractURIName(FURL);
+end;
+
+procedure TImageAsX3DModelLoader.ReadImageProperties(const URL: String);
+var
+  Image: TCastleImage;
+begin
+  FImagePath := URL;
+  if (FImageWidth > 0) and (FImageHeight > 0) then
```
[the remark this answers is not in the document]
and3md: This is just optimization, when user add anchor option `width` and `height` do not load image just to read image height and width. I will add comment here (in sources).

src/x3d/x3dloadinternalcocos2d.pas
```diff
@@ -0,0 +1,999 @@
...
+
+{ TCocos2dLoader }
+
+procedure TCocos2dLoader.ReadImportSettings;
+var
+  SettingsMap: TStringStringMap;
+  Setting: TStringStringMap.TDictionaryPair;
+begin
+  // default values
+  FFramesPerSecond := 4.0;
```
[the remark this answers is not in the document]
and3md: In other hand in spritesheet-to-x3d the default value is 4.0. In this way, we leave compatibility with the previous solution for sprite sheets in CGE. I think spritesheet-to-x3d was a lot modern solution than TSprite. So I decided to leave that value (4.0). Maybe 6 or 8 is better but it's hard to say. Usually, sprite sheet animations don't have many frames (I think) and 25 will force most of users to always change that value. Maybe "Kenney" test should have last word :)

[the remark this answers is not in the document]
and3md: Our zombie at 4.0 is a little slow, maybe should be faster :)

src/x3d/x3dloadinternalcocos2d.pas
```diff
@@ -0,0 +1,999 @@
...
+              HasAnchor := true;
+          end;
+        'spriteSourceSize':
+          begin
+            { full size of the sprite, the same as sourceSize in format 2 }
+            if ReadDual(ValueNode.TextData, FullFrameWidth, FullFrameHeight) then
+              WasFrameFullSize := true;
+          end
+        else
+          continue;
```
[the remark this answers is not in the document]
and3md: > Also does it deserve a WritelnWarning('Cocos2d', 'Unrecognized value %s', [KeyNode.TextData]); ?

There can be data that we don't use currently (like `triangles`, `vertices`, `verticesUV` for polygon sprites) so warnings can be to verbose. I think at some point I had here some log and removed it and `else` remianed as oversight.

src/x3d/x3dloadinternalstarling.pas
and3md: > Also, place Destroy in the implementation right after Create, so move Destroy implementation a bit higher. We follow this convention in CGE: first Create, then Destroy, then the rest of methods in completely random order :)

i thought this is just for the interface section, understand now, fixed.

src/x3d/x3dloadinternalstarling.pas
and3md: I can't find `RemoveSuffix` function, do you mean `RemoveTralingChars` ? I didn't use that here because I thought someone can name animation frames as `_1`, `_2`. So I check if this is only `_` as name then I don't remove this char.

src/x3d/x3dloadinternalcocos2d.pas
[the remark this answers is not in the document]
michaliskambi: :+1: I second the idea """Maybe "Kenney" test should have last word :)"""

We neither have to be consistent with TSprite, or with sprite-sheet-to-x3d. We can figure out good default independently.

src/x3d/x3dloadinternalstarling.pas
michaliskambi: Sorry, it's `SuffixRemove` :) https://castle-engine.io/apidoc-unstable/html/CastleStringUtils.html#SuffixRemove

and3md: `SuffixRemove` also removes '_' when animation name is '_'. So in this case animation name will be empty. See comment above.

src/x3d/x3dload.pas
and3md: For the `ReadImageExtensions` I think I found better solution. I propose using `LoadImage_FileFilters` and add `function FilterString :String` to `TFileFilterList`. This aproach also makes `TFileFilterList` more complete. What do you think about that?

and3md: Of course, we needed a TFilterList.AllExtensions: String function that omitted * and *. *.

src/x3d/x3dloadinternalstarling.pas
michaliskambi: Ah, sure. I would then say to write 

```pascal
if AnimationName <> '_' then // do not remove underscore if it's the only character in name
  AnimationName := SuffixRemove('_', AnimationName);
```

This is still a bit shorter and more obvious :)

src/x3d/x3dloadinternalimage.pas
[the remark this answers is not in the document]
michaliskambi: But to make this optimization, you should check then `  if (FWidth > 0) and (FHeight > 0) then ...`. That is, anchor sets `FWidth` and `FHeight`, not `FImageWidth` and `FImageHeight`.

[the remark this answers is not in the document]
and3md: Yes, you are right, I thought about Starling when writing comment. This check should be removed.

michaliskambi (approved):

michaliskambi:
Thank you! All perfect. Merging. I will announce it over the weekend, and will link to
https://github.com/castle-engine/castle-engine/wiki/Sprite-sheets
https://github.com/castle-engine/castle-engine/wiki/Images
everywhere.

PR #222, opened 2020-11-03
Clear pressed keys in progress bar. 

and3md:
This also can fix Frogger3d but I add it more for another reason. The `TProgress` is visible when some long operations are made. So passing the state of the keys may trigger unnecessary responses, especially because of `TGLMode` can do `SimulateReleaseAll()`. 

This PR can also eliminate some of the weaknesses of TGLMode. But in the future (maybe after 7.0) we need totally rebuild TProgressBar (and messageboxes also?). 

## Idea

I understand that TGLMode is supposed to enable state display without affecting the user-defined state stack. At first glance, in addition to the user stack, I would introduce something like `SystemState` that would be prioritized. If `SystemState = nil` then the user states would be used. Otherwise, the system state would take precedence. I don't know the implementation details yet, so it's just an idea. What do you think about it? Can we add something like this to trello?

### DIFF
--- a/src/window/castlewindowprogress.pas
+++ b/src/window/castlewindowprogress.pas
@@ -116,6 +116,10 @@ begin
     Bar.Background := UsedWindow.SaveScreen;
   Bar.YPosition := BarYPosition;
 
+  { Reset all keys because TGLMode.Create() can do a lot of uneeded things in
+    SimulateReleaseAll(). And probably no one expects the progress bar to
+    try to react to the pressed keys in some way. }
+  UsedWindow.Pressed.Clear;
   SavedMode := TGLMode.CreateReset(UsedWindow, nil, nil, @NoClose);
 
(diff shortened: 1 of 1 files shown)


michaliskambi (approved):

michaliskambi:
Thank you, for all 3 PRs (#220, #221, #222) !

As for `SystemState` -- I don't think we need to go there. I see it would be a better way to "express" the functionality of TGLMode (without some save/restore hacks), but it would also mean that this functionality is more coupled with code in CastleWindow / TUIContainer in CastleUIControls . And we want to avoid adding complication there.

The one advantage of current TGLMode is that it is implemented "on top" of mechanisms in CastleWindow / TUIContainer in CastleUIControls . So we know that any modification to TGLMode has a limited effect, e.g. it cannot affect how TCastleUserInterface works, it cannot affect how a game (that doesn't use CastleWindowProgress or CastleMessages) works.

PR #223, opened 2020-11-04
Improve undo recording and undos for gizmos

eugeneloza:
* Refactor ModifiedOutsideObjectInspector to handle Undo recording.
* Properly implement Undo for Gizmos on drag-n-drop.


tools/castle-editor/code/framedesign.pas
```diff
@@ -1760,9 +1755,9 @@ procedure TDesignFrame.CastleControlUpdate(Sender: TObject);
 
   if InternalCastleDesignInvalidate then
   begin
-    ModifiedOutsideObjectInspector;
     UpdateDesign;
     //WritelnWarning('CGE needed to explicitly tell editor to refresh hierarchy');
+    ModifiedOutsideObjectInspector('Internal update'); //WARNING: I don't really know when this happens and how it should behave
```
michaliskambi, line 1760: `InternalCastleDesignInvalidate` means that something in CGE changed hierarchy. 

Right now it happens when `ExposeTransforms` is set and in effect new `TCastleTransform` is created and added as child of some existing `TCastleTransform`. In general it will happen always when something in CGE needs to change UI/transform hierarchy but we did not handle this change (by calling `UpdateDesign` explicitly) elsewhere in editor (e.g. `InternalCastleDesignInvalidate` is also set when creating new component, but in this case it is handled by `UpdateDesign` earlier).

Calling it an "Internal update" is not accurate, something changed in response to user actions and the change is visible, it's not really "internal" :)

So just make it `ModifiedOutsideObjectInspector('');`

tools/castle-editor/code/framedesign.pas
```diff
@@ -725,7 +721,9 @@ function TDesignFrame.TDesignerLayer.Motion(const Event: TInputMotion): Boolean;
         end;
     end;
 
-    Frame.ModifiedOutsideObjectInspector;
+    //this will schedule recording Undo OnRelease
+    //and will prevent from recording Undo by applying changes below
+    Frame.ModifiedOutsideObjectInspector('', true);
```
michaliskambi, line 726: I find above 2 lines somewhat unclear. Also remember we usually make space between `//` and text in CGE comments :) I propose to change it to this:

```
{ By default ModifiedOutsideObjectInspector would record this as undo change.
  But we pass UndoOnRelease = true to defer it to the Release call. }
```

tools/castle-editor/code/framedesign.pas
```diff
@@ -2682,8 +2679,7 @@ procedure TDesignFrame.MenuItemViewportCamera2DViewInitialClick(
 begin
   V := SelectedViewport;
   V.Setup2D;
-  ModifiedOutsideObjectInspector;
-  RecordUndo('Camera Setup for 2D View and Projection');
+  ;ModifiedOutsideObjectInspector('Camera Setup for 2D View and Projection');
 end;
```
michaliskambi, line 2683: Accidental `;` at the beginning of line :)

tools/castle-editor/code/visualizetransform.pas
```diff
@@ -605,6 +617,13 @@ procedure TVisualizeTransform.GizmoHasModifiedParent(Sender: TObject);
     OnParentModified(Self);
 end;
 
+procedure TVisualizeTransform.GizmoStopDrag(Sender: TObject);
+begin
+  if Assigned(OnGizmoStopDrag) then
+    OnGizmoStopDrag(Self);
+end;
+
```
michaliskambi, line 625: Additional newline.

michaliskambi (changes requested):
Thank you! All looks good, I only pointed minor issues.

michaliskambi:
P.S. Notice cool Jenkins status visible in this PR, Jenkins is checking that this PR compiles now everywhere :)

tools/castle-editor/code/framedesign.pas
eugeneloza: I see...

Here let's make one more decision. Now Undo recorded with `Comment := ''` actually transforms into:

```Pascal
if UndoHistory[UndoI].Comment = '' then
begin
  if UndoHistory[UndoI].Selected = '' then
    Result := 'Modify global variable'
  else
    Result := 'Modify ' + UndoHistory[UndoI].Selected;
end
```

I.e. in this case we'll get 'Modify global variable' which is not actually correct. Maybe we should change this feature in core. E.g. in this case let's the menu item caption be just `'Undo'` without any specifics? WDYT?

michaliskambi: Oh, I see. Yes, I expected `UndoComment=''` to make just `'Undo'` menu caption without any specifics.

The fix should be a little more involved than just changing `'Modify global variable'` case. Because there may be something selected at this moment, and we still want to just show `'Undo'` menu caption.

I guess we have a problem with competing ideas how to treat `UndoComment=''`. E.g. I see `AdjustUserInterfaceAnchorsToKeepRect` also passes `UndoComment = ''`, and in this case it makes sense that it results in `'Modify ' + Selected`.

Proposed solution: When things are unclear, introduce an enumerated type to make it explicit and clear :) Like this:

```pascal
TUndoType = (
  // This undo has no description (UndoComment is ignored and should be left empty).
  utNoDescription,
  // This undo has description in UndoComment.
  utDescription,
  // This undo should have description like "Modify SelectedComponentName" (UndoComment is ignored and should be left empty).
  utModifiedSelected,
  // No undo is recorded, defer recording it to OnRelease. (UndoComment is ignored and should be left empty).
  utDeferToOnRelease
);
```

And now:

- `ModifiedOutsideObjectInspector` takes `const UndoType: TUndoType` as 1st parameter

- `ModifiedOutsideObjectInspector` no longed needs `UndoOnRelease` boolean, `utDeferToOnRelease` signals this purpose.

- We should explicitly forbid "Modify global variable", because that should not be really possible (you cannot do this in editor, and you could, we could not be able to undo it -- as we only save/restore the design).

So add to `ModifiedOutsideObjectInspector` 

```
{ When UndoType is utDescription, then non-empty UndoComment is required.
  When UndoType is something else, then UndoComment is ignored, and we require it to be empty for now. }
Assert((UndoComment <> '') = (UndoType = utDescription));

{ Using utModifiedSelected is only allowed when something is selected. }
Assert((UndoType <> utModifiedSelected) or (SelectedC <> nil));
```

tools/castle-editor/code/visualizetransform.pas
eugeneloza: fixed in 123ef82e873ac874b687605da5d13e5693b53fd8

tools/castle-editor/code/framedesign.pas
eugeneloza: Fixed in 8f02ccff9a05f7cafc9e5eb9f74cd137140148e2

tools/castle-editor/code/framedesign.pas
eugeneloza: Done, I've rewritten it in a (very) different way, please, check it out

tools/castle-editor/code/framedesign.pas
eugeneloza: After 2 days of thinking over and struggling to find a good solution (without asserts), I've came up with a (significantly) different idea. The problem was that we treated `''` as a "magic" comment that implied "low priority comment". And that "magic" actually harmed us in the way described above :) In the new approach I explicitly added "UndoCommentPriority" which specifies how high the priority of the given comment is.

As a by-product I also significantly improved the undo comments themselves, and made them as informative as it seems to be possible.

eugeneloza:
It seems like I've fixed everything here, ready for the next round :D

michaliskambi (changes requested):
I was wondering about the "priority" parameter. I admit I find it over-complicated. As a person writing new operation, I would have a a hard time figuring what priority to use -- I would just use "high priority" which seems used by all code that passes non-empty UndoComment.

And it also seems the parameter is ignored by `ModifiedOutsideObjectInspector` ?

Also, it seems error-prone that it matters whether something is selected when `InternalCastleDesignInvalidate` occurs -- it should not. As far as I can see, right now it could result in comment `'Modify ' + UndoHistory[UndoI].Selected` if I happen to have something selected when `InternalCastleDesignInvalidate` occurs, which should not be (the change signalled by `InternalCastleDesignInvalidate` may be anywhere).

I'm after changing it to my proposed `TUndoType` from above comment. This avoids speaking about "priority", it instead makes the intention explicit. The idea is to not do the `'Modify ' + UndoHistory[UndoI].Selected` automatically (when comment is empty), instead it should be done only when explicitly requested by using `UndoType` parameter as `utModifiedSelected`.

tools/castle-editor/code/framedesign.pas
```diff
@@ -3004,6 +3024,11 @@ procedure TDesignFrame.ModifiedOutsideObjectInspector;
     PropertyGridModified(nil);
 
   MarkModified;
+
+  if UndoOnRelease then
+    UndoSystem.ScheduleRecordUndoOnRelease := true
+  else
+    RecordUndo(UndoComment, HighUndoPriority); //it will overwrite Undo recorded in PropertyGridModified with a better comment
```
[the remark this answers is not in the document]
eugeneloza: Yes, thank you! Fixed.

eugeneloza:
Sorry, your comment left me absolutely confused, until I realized that I've forgotten to apply a significant part of the fix :D Had to rewrite my long-long answer from the square one.

Yes, it was intended that

> The idea is to not do the 'Modify ' + UndoHistory[UndoI].Selected automatically

will be removed. Instead I've left half-applied fix and in the end it was completely broken ('' with sel=nil would have resulted in no comment at all). Fixed it now in 3b46c0a3e6342368da2fa68cf751d6bb6b85e2a0. Now all `RecordUndo` are sent with proper comment. `''` means just "Undo". Hopefully I didn't screw it up this time.

I'm not sure if that would make the meaning of `UndoCommentPriority` more clear now?

> As a person writing new operation, I would have a a hard time figuring what priority to use -- I would just use "high priority" which seems used by all code that passes non-empty UndoComment.

Yes, that's exactly how it works. The Undo commnet with lower priority gets overwritten with Undo comment with higher priority. E.g. in case of `TDesignFrame.PropertyGridModified` sometimes may happen that `Sender is not TOICustomPropertyGrid` so we don't know what exactly happened (`Sel` somehow got modified). E.g. this happens when the new element is inserted - and the insertion procedure will call `ModifiedOutsideObjectInspector` which will in turn overwrite Low-priority comment of "Modify Edit1" with more meaningful "Inserted Edit1".

This way, yes, each and every `UndoComment` that is not "something happened to something" should have "high priority". This is how every new `UndoComment` should be added. They all have high priority, as all current low-level comments are already covered and new comments should overwrite the old ones.

That also means that technically our `UndoCommentPriority` has two values: High and Low. We may reduce it to "true"/"false", but that's less obvious to me, than the explicit words "high"/"low".

That said I may think of an event when we'll need more than 2 possible comment priorities. The current "highest" priority is used only internally and actually doesn't really do anything and will only mean something only in case we have more than base 2 priorities.

E.g. let's imagine we have "Add FPS Player". It would record:

* "Modified Player": LowUndoPriority
* "Rearrange Scene Cameras": MiddleUndoPriority
* "Added Player": HighUndoPriority

I.e. as any other big event it will fire `TDesignFrame.PropertyGridModified` which will record a "generic" "Modified something" comment. It's our worst case. Then we receive the callback that the Scene registered a new camera and record it as a new Undo event (which could be useful if we know nothing more of what exactly has happened) and finally in the end of the procedure we call `ModifiedOutsideObjectInspector` which tells accurately what has happened.

I'm not sure if I explained it better, or did I confuse even myself now? :D

michaliskambi:
@eugeneloza OK, I can see how the system can be useful to extend in the future. Thank you for explanation. Sorry for prolonging this so much -- I want to understand it best, as you write we have a "system of priorities with only 2 priorities being actually used". So we need to make sure this is not "over-engineering".

Final notes from me:

1. Change the "priority" type to an enum. As this is an internal system, we can always extend this enum. This way we have a "single source of truth" to know what priorities are actually used throughout the code.

2. I thought about the name. The name "undo priority" is a bit confusing, it's really "how meaningful is the provided undo comment". Hm, maybe just "undo comment priority".

So I would say to define it like this:

```
TUndoCommentPrlority = (ucLow, ucNormal);
```

eugeneloza:
> 
> 
> @eugeneloza OK, I can see how the system can be useful to extend in the future. Thank you for explanation. Sorry for prolonging this so much -- I want to understand it best, as you write we have a "system of priorities with only 2 priorities being actually used". So we need to make sure this is not "over-engineering".
> 
> Final notes from me:
> 
>     1. Change the "priority" type to an enum. As this is an internal system, we can always extend this enum. This way we have a "single source of truth" to know what priorities are actually used throughout the code.
> 
>     2. I thought about the name. The name "undo priority" is a bit confusing, it's really "how meaningful is the provided undo comment". Hm, maybe just "undo comment priority".
> 
> 
> So I would say to define it like this:
> 
> ```
> The name "undo priority" is a bit confusing, it's really "how meaningful is the provided undo comment". Hm, maybe just "undo comment priority"

Makes perfect sense. It's actually called like this around the UndoSystem - I just "shortened" one word for the constants definition (making them less obvious as a side-effect). I'll fix it.

eugeneloza:
Done! However, I'm not yet sure how clean now looks this line: https://github.com/castle-engine/castle-engine/pull/223/files#diff-3f396e9c2f2a1d4c5b9422520dbebc210369d3e80cf3a2ac4e0d502fdfb3eb6cR176 -> WDYT of comparing Ord(Enum)?

tools/castle-editor/code/castleundosystem.pas
```diff
@@ -11,6 +11,20 @@ interface
...
+    { Low priority comments will get overwritten by a higher priority comments
+      which correspond to a more specific description of what has happened,
+      but are not always available. }
+    ucLow = 0,
+    { High priority comments usually describe what exactly happened
+      and should overwrite less specific comments. }
+    ucHigh = 1,
+    { Highest priority comments should never be overwritten.}
+    ucHighest = 31
+    );
```
michaliskambi, line 25: Do not use explicit numeric values for the enums here. I see no reason for it. And you should not need `ucHighest`, as you can use `High(TUndoCommentPriority)`.

So this definition can be simpler, just 

```pascal
  TUndoCommentPriority = (
    { Low priority comments will get overwritten by a higher priority comments
      which correspond to a more specific description of what has happened,
      but are not always available. }
    ucLow,
    { High priority comments usually describe what exactly happened
      and should overwrite less specific comments. }
    ucHigh
);
```

tools/castle-editor/code/castleundosystem.pas
```diff
@@ -136,34 +157,36 @@ function TUndoSystem.UndoHistorySize: Integer;
...
-    else
+    if (SelectedComponent <> UndoHistory[CurrentUndo].Selected) then
     begin
+      WriteLnLog('New Undo is identical to previous Undo record. Only selection has changed from ' + UndoHistory[CurrentUndo].Selected + ' to ' + SelectedComponent + '. This change has been saved.');
       UndoHistory[CurrentUndo].Selected := SelectedComponent;
-      WriteLnLog('New Undo is identical to previous Undo record. Only selection has changed from ' + UndoHistory[CurrentUndo].Selected + ' to ' + SelectedComponent + '. Not saving.');
       OnUpdateUndo(Self);
     end;
-    if (UndoComment <> '') and (UndoHistory[CurrentUndo].Comment = '') then
+    if (Ord(UndoCommentPriority) > Ord(CurrentUndoCommentPriority)) then
```
michaliskambi, line 176: Do not cast to integers before comparison, it is not necessary. Just compare 

```
    if UndoCommentPriority > CurrentUndoCommentPriority then
```

michaliskambi (changes requested):

eugeneloza:
Done! :)

michaliskambi:
That was fast :) Resolved conflict (as I applied https://github.com/castle-engine/castle-engine/pull/231 in the meantime), merging. Thank you!

michaliskambi (approved):

PR #224, opened 2020-11-16
Update default Android target API level to 29

and3md:
I changed the default target API level to 29 so the build tool meets the minimum Play Store requirements.

https://developer.android.com/distribute/best-practices/develop/target-sdk

Tested by compiling and running my game, drawing_toy, simple_3d_demo, 2d_dragon_spine_game.

### DIFF
--- a/tools/build-tool/code/toolproject.pas
+++ b/tools/build-tool/code/toolproject.pas
@@ -368,7 +368,7 @@ constructor TCastleProject.Create(const APath: string);
     DefautVersionCode = 1;
     { iOS requires version display to be <> '' }
     DefautVersionDisplayValue = '0.1';
-    DefaultAndroidCompileSdkVersion = 28;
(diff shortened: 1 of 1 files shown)


michaliskambi (approved):

michaliskambi:
Thanks!

https://github.com/castle-engine/castle-engine/wiki/Android also should be updated now (Ctrl + F on that page for "28" to find occurrences where it needs to be updated :) ).

michaliskambi:
Also https://github.com/castle-engine/castle-engine-cloud-builds-tools should be updated, at https://github.com/castle-engine/castle-engine-cloud-builds-tools/blob/master/Dockerfile.no-cge#L72 it defines Android platform and Android build tools used for building.

I just made this repo public, it is used to make our Docker image ( https://github.com/castle-engine/castle-engine/wiki/Docker ).

PR #226, opened 2020-11-21
Types capitalization

eugeneloza:
This is an attempt to bring more order in CGE code. Most here is capitalization of types: Integer, String, Boolean, etc. The major idea is to have more order in overrides of virtual methods.

~~I'll need to review this one very careful and fix expected merge conflict.~~ done

Short summary of the changes:
* All types I could think of (String, Integer, Pointer...) are now capitalized both in CGE API, and in internal variables.
* I've left "third-party" files unchanged (those that don't have CGE copyright notice), for the sake of "cleaner updating" them in future should it be needed.
* I've made changes to autogenerated files, ignore those, they will be overwritten and that's ok. It was just "easier" to change types in there automatically than to analyze which file should or should not be changed. As the code inside the generators should also have been changed, that means they'll have correct capitalization on next regeneration.
* I've tried to keep smart capitalization in comments, though it wasn't always obvious in which cases the word was representing a _type_ and in which - just a regular word. E.g. "the integer value" doesn't refer to a type, but "returns Integer" was capitalized.
* I've tried to keep old capitalization in string variables or log messages unless it was absolutely obvious that the word represents a type here and should be capitalized and its capitalization is unlikely to break anything.
* From time to time I've also made some small changes, e.g. capitalize variables, usually "i" -> "I", etc. This was done unsystemmatically and is not complete, and after reviewing the changes it seems like a bad idea, as it makes review less obvious/clean. But anyway it's done and reviewed :)
* I've made "more modifications" to a couple of code chunks that were written by me very long ago and didn't follow any reasonable conventions.


eugeneloza:
Self-review and self-tests finished, the PR is ready for review.

michaliskambi:
Applying this PR is problematic for me, because applying it will inevitably put all the other work-in-progress forks/branches (including *all* other PRs to CGE) in a conflicting state. As this PR basically touches almost all lines in the whole CGE Pascal code.

So, I admit I don't know how to proceed best yet. I don't want this work go to waste, but the gain from applying this PR never seems justified.

One thing I'm wondering about (but this is also imperfect solution) would be to split the PR into more PRs, and apply e.g. 1 per week:

- Update code style in src/base/
- Update code style in src/x3d/
- ...
- Update code style in examples/

This way each particular PR will have less change of causing damage (conflicts). But this is also a poor solution -- as conflicts will happen anyway, and we will essentially make a painful process much longer this way. (Instead of applying one big-breaking PR at once, we would apply e.g. 1 somewhat-breaking PR per week, for 8 weeks.)

eugeneloza:
Yeah, and with every day maintaining this branch will require more and more effort to fix all the new merge conflicts.

I don't really see a serious/critical problem here, we can overwrite changes in this PR without questions - the merge conflicts do not seem to be as severe as they tend to be (because the change is trivial). I see at least 5 PRs in different state of work, so technically I could just fix merge conflicts and push changes to those PRs. Then once it's done they won't have a much larger amount of merge conflict than merge conflict with original master. If this branch will be without merge conflict with master that should not be hard (as all of those originate from master). And normally other PRs concentrate on several existing units, so they don't create an extensive merge conflict themselves. However, this may become tricky if the PR has its own merge conflict.

However, in the end... Looking at FPC code - maybe it doesn't really matter? I see two things that have been improved by types capitalization and can be further improved for better 1) code readability 2) code overriding. But this gain is actually very minor. If it creates too much problems for collaborative work, then the gain from collaborative work is higher, than the mythical gain from the two points above.

Such simple changes are much easier to be done extensively. I.e. to capitalize all instances of "boolean" it's a simple search-and-replace, paying the minimal essential attention to every instance. I.e. the effort to rework each and every single unit separately is significantly higher, than rework the whole folder in one go.

So I can see two solutions:

1. Drop this branch and forget about fixing code quality extensively, only do minor fixes randomly from time to time as the related code gets update. **This doesn't require any additional effort.**
2. Pull together all PRs, fix all merge conflicts and prepare for more, because type capitalization is only one of many things that can be improved (variables capitalization, indentation, line splitting, comments translation, etc.). **This way we add ourselves a lot of work, which doesn't contribute to new Engine features. This also creates danger of possible bugs introduced due to typos, accidentally skipped during review**. The fact that some of this work has already been done doesn't justify the future work that will be required.

michaliskambi:
Well, code quality and readability matters a lot. I don't want to just close this PR "because it is too prone to cause conflicts", as in the long run it matters to have consistent style, and conflict is a "one-time thing" to resolve. Having CGE code in a state that is consistent is important, esp. to new contributors. I like to think CGE is already here in a good shape, and I don't want to be wrong because of too-many-decisions-in-the-meantime-that-we-only-applied-halfway.

But, we also don't want to cause problems for parallel work of everyone on the project.

Let's wait with rebasing this (I'll let you know when to rebase, hopefully next week there will be 3 PRs less) and then merge.

Your proposal that you can fix current PRs is great (thank you!), although it does not remove all my fears:

- We have now 71 public forks of the engine :) We do not control everything that people do with CGE in parallel. I talk privately with people that play with CGE, sometimes without using GIT at all, sometimes without desire to make a public PR soon -- for various reasons.

- The existing PRs are in various states -- some new ones will probably get applied soon (so no need to fix them, this PR can wait for them). Some older ones are work-in-progress and it will be a larger work for you to resolve them. They do have conflicts.

- As for your note "prepare for more" --- please let's synchronize such things before you do them :) Otherwise such efforts are creating more work than they should. You need to rebase and rebase..., I have to plan when to apply it at a perfect moment. 

    - In case of some code changes, you can first warn, if you can predict _"this task will modify > 100 Pascal source files..."_ then it is better to discuss it first :) Then we can think of the best approach together.

    - I can propose many other CGE tasks that are also easy, but more valuable and much less troublesome to merge :) There is a value in minor code quality fixes, but there are also 100 other areas in CGE where we welcome contributions and where potential conflicts are not a problem :)

eugeneloza:
> hopefully next week there will be 3 PRs less

And hopefully at least 2 more (MaterialProperties and FMOD) :D

Overall, even if we make CGE code base perfect, we still go down to FPC code base, and it often looks ... the way it looks: no comments, no conventions. And it's not yet fixed for a reason - the same forks, PRs, possible issues - I don't believe they don't fix it just because they don't want/don't care. So, maybe still this is a good way to go.

michaliskambi:
@eugeneloza Now is the time :) Please fix any conflicts (but not introduce any new code-style fixes within this PR, keep it minimal now -- if in doubt, leave things unmodified) and I can apply it.

michaliskambi (changes requested):
Waits for resolving conflicts, as in comments above :)

PR #231, opened 2020-12-01
Ability to drag assets directly into the viewport

and3md:
Works in 2D and 3D, in case of no hitting any object, it takes default values of distance (3D - 20, In case of 2D try to use camera Z position or 100 if smaller than 100.

### DIFF
--- a/tools/castle-editor/code/framedesign.pas
+++ b/tools/castle-editor/code/framedesign.pas
@@ -1815,14 +1815,13 @@ var
   ShellList: TCastleShellListView;
   SelectedFileName: String;
   SelectedURL: String;
-  R: TRegisteredComponent;
   Scene: TCastleScene;
   UI: TCastleUserInterface;
   Viewport: TCastleViewport;
-  Pos2D: TVector2;
-  Z: Single;
-  Asset2D: Boolean;
-  Transform: TCastleTransform;
+  ScenePos: TVector3;
+  RayOrigin, RayDirection: TVector3;
+  RayHit: TRayCollision;
+  Distance: Single;
 
   function Is2DAsset(URL : String): Boolean;
   var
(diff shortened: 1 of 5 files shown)


tools/castle-editor/code/framedesign.pas
```diff
@@ -1776,6 +1786,114 @@ procedure TDesignFrame.CastleControlUpdate(Sender: TObject);
...
+      if RayHit <> nil then
+      begin
+        Distance := RayHit.Distance;
+        FreeAndNil(RayHit);
+      end else
+      begin
+        // if we don't hit any other scene set Distance to default
+        if Viewport.Camera.ProjectionType = ptOrthographic then
+        begin
+          // use camera position for 2D (minimum 100 distance)
```
michaliskambi, line 1875: I would suggest (but please test in practice) to make a different logic in this case (`Viewport.Camera.ProjectionType = ptOrthographic`, i.e. nothing was hit and we have ortho projection):

Place object at `Z = (Viewport.Camera.EffectiveProjectionNear + Viewport.Camera.EffectiveProjectionFar) / 2` (and this should make Z = 0 in case of default projection set by `TCastleViewport.Setup2D`, which is nice and most natural for user). So you would need to rearrange code, in this case we don't follow distance, but we do 

```pascal
PlaneZ = (Viewport.Camera.EffectiveProjectionNear + Viewport.Camera.EffectiveProjectionFar) / 2;
if not TrySimplePlaneRayIntersection(ScenePos, 2, PlaneZ, RayOrigin, RayDirection) then 
  Exit; // camera direction parallel to 3D plane with Z = constant
```

... and you get ready `ScenePos`, and don't care at all about calculating `Distance`.

Only in other cases, you do `RayOrigin + (RayDirection * Distance)`.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1776,6 +1786,114 @@ procedure TDesignFrame.CastleControlUpdate(Sender: TObject);
...
+    UI := FDesignerLayer.HoverUserInterface(Vector2(X, Y));
+    if not (UI is TCastleViewport) then
+      Exit;
+
+    Viewport := TCastleViewport(UI);
+    try
+      Scene := AddComponent(Viewport.Items, TCastleScene, nil) as TCastleScene;
+
+      if Is2DAsset(SelectedURL) then
+        Scene.Setup2D;
```
michaliskambi, line 1852: In the end, I think we can just remove 

```pascal
if Is2DAsset(SelectedURL) then
  Scene.Setup2D;
```

and remove hardcoded `Is2DAsset` detection. It's my fault, I was confusing you and myself when we talked about this :) Thinking about it clearly now, this `Scene.Setup2D` (which does `RenderOptions.BlendingSort := bs2D`) just doesn't matter in this case:

- for images or sprite sheets, `RenderOptions.BlendingSort` don't matter. They only show 1 layer.

- for Spine models, `RenderOptions.BlendingSort` don't matter, as imported Spine models use `NavigationInfo` to request blendingSort = 2D anyway.

All in all, calling `Scene.Setup2D` is only really useful if you import a model from a "general" (3D or 2D) format like glTF or X3D, but you want to use it as 2D, and the model has multiple layers that use blending. I made a note about it on https://github.com/castle-engine/castle-engine/wiki/Blending . Doing it automatically, in a hardcoded subset of cases, may be more confusing than helpful.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1776,6 +1786,114 @@ procedure TDesignFrame.CastleControlUpdate(Sender: TObject);
...
+        // if we don't hit any other scene set Distance to default
+        if Viewport.Camera.ProjectionType = ptOrthographic then
+        begin
+          // use camera position for 2D (minimum 100 distance)
+          if Viewport.Camera.Position.Z > 100 then
+            Distance := Viewport.Camera.Position.Z
+          else
+            Distance := 100;
+        end else
+          Distance := 20;
```
michaliskambi, line 1881: I fear such hardcoded values, but I see no other choice here :) But make it 10 instead of 20 here -- it seems more "obvious default".

michaliskambi (changes requested):
Thank you! I have some comments, see below. They are mostly minor stuff, except I wondered about `TDesignFrame.CastleControlDragDrop` -- how to avoid hardcoding there some assumptions, but at the same time keeping the behaviour natural for users.

tools/castle-editor/code/framedesign.pas
```diff
@@ -1224,6 +1232,27 @@ procedure TDesignFrame.BeforeProposeSaveDesign;
 
 procedure TDesignFrame.AddComponent(const ComponentClass: TComponentClass;
   const ComponentOnCreate: TNotifyEvent);
+var
+  Selected: TComponentList;
+  SelectedCount: Integer;
+  ParentComponent: TComponent;
+begin
+  if ControlsTree.Selected <> nil then
+    ControlsTree.Selected.EndEdit(true);
```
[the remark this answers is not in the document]
and3md: Thanks, fixed

tools/castle-editor/code/framedesign.pas
and3md: I tested 10 and 20. 10 also works OK on our default 3d template. However, I thought the greater value would be safer. So I doubled it up. Of course we can change it to 10 (I'll do that). Maybe that should be in project configuration in the future.

michaliskambi (approved):

michaliskambi:
All looks great, merging -- thank you!

PR #239, opened 2020-12-17
Add basic rendering functionality for TiledObject primitives (Polygon, Polyline, Point, Rectangle)

Free-Pascal-meets-SDL-Website:
Hello,

this commit adds 
- rendering functionality for ObjectGroupLayers TiledObjects (polygon, polyline, point, rectangle) from TiledMaps

this commit doen't add
- rendering of ellipses
- rendering of labels of TiledObject

desired TODO feature
- transforming TiledObjects into TCastleUserInterfaces (perhaps based on TCastleShape?)

Let me know, what you think.
Have a nice day,
Matthias

### DIFF
--- a/src/ui/opengl/castletiledmap.pas
+++ b/src/ui/opengl/castletiledmap.pas
@@ -34,7 +34,7 @@ uses
   Classes, SysUtils, DOM, XMLRead, base64, zstream, Generics.Collections,
   CastleVectors, CastleColors, CastleUtils, CastleURIUtils, CastleXMLUtils,
   CastleLog, CastleStringUtils, CastleUIControls, CastleGLImages,
-  CastleRectangles, CastleClassUtils;
+  CastleRectangles, CastleClassUtils, CastleRenderOptions;
 
 {$define read_interface}
 {$I castletiledmap_map.inc}
--- a/src/ui/opengl/castletiledmap_control.inc
+++ b/src/ui/opengl/castletiledmap_control.inc
@@ -223,20 +223,29 @@ var
   Lay: TTiledMap.TLayer;              // ObjectGroupLayer
   TiledObj: TTiledMap.TTiledObject;   // one TiledObject
   TiledObjPoints: array of TVector2;  // holds points of TiledObject as array of TVector2
-  ScaleX, ScaleY: Single;
+
+  function ScaleX: Single;
+  begin
+    { Denominator corresponds to width of object layer in pixels. }
+    Result := RenderRect.Width / (Map.Width * Map.TileWidth);
+  end;
+
+  function ScaleY: Single;
+  begin
+    { Denominator corresponds to height of object layer in pixels. }
+    Result := RenderRect.Height / (Map.Height * Map.TileHeight);
+  end;
 
   { Converts all TVector2's from a TVector2List to and array of TVector2.
     ArrIndex is the index of a TVector2 in the array, ListIndex is the TVector2
     in the list. }
-  procedure ConvertAndScalePoints(ArrIndex, ListIndex: Cardinal);
+  procedure ConvertAndScalePoints(const ArrIndex, ListIndex: Cardinal);
   begin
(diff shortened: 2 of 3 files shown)


michaliskambi (changes requested):
1. Thank you! The code looks cool. I noted 2 tiny things to correct, although this is really nit-picking :)

2. Can you point me to (or attach here) an example Tiled map to test this (see the lines, points on a map)? Nothing fancy-looking necessary, just any file where I could see it working and confirm it looks correct (same in CGE as in Tiled editor).

3. As for expressing more and more shapes this way (like ellipses or expressing things as TCastleShape, as you mention):

    I think that eventually in the future, we shall express the Tiled map as a graph of X3D nodes. We already have nodes to express things like circles and text (and yes, the existence of such nodes kind of duplicates some functionality of UI classes -- because sometimes one or other API is easier to use; I have a plan to reconcile it, by rendering all UI internally as nodes in the future). We have examples how to build such nodes using Pascal e.g. on https://castle-engine.io/x3d_implementation_geometry3d.php and https://castle-engine.io/x3d_implementation_geometry2d.php .

    Once this is done, the existing Tiled map UI control would become just a container for internal instance of TCastleViewport, with TCastleScene inside, and in that TCastleScene we would define nodes to draw things (rectangles with images, shapes, etc.).

    One advantage is that X3D nodes are quite powerful and they can be used for 3D too. A Tiled map expressed as X3D nodes could be as easily rendered in 3D too. And it could be enhanced by 3D objects. Like in the old "Orcs and Volcanoes" old game I made https://castle-engine.io/old_news.php?id=2013-09-15 which was played on a 2D map, but sometimes (just for visual effect) I actually showed it in 3D, with perspective, with some 3D objects (volcanoes and towers on the screenshots there).

    Another advantage is that TCastleScene rendering of nodes is already very optimized. E.g. we could just use frustum culling to avoid drawing invisible map parts. We could use `DynamicBatching` to use less draw calls when rendering same tiles.

    That being said, I know it's a large rework. Existing tiles rendering would need to be converted to this system too.

    For the time being, simply drawing additional things using existing API "simple 2D API", that is: `TDrawableImage` and `DrawPrimitive2D`, is cool.

Free-Pascal-meets-SDL-Website:
Hello there,

Re to 1) I hope the code refinement is correct. In the last commit I re-worked the scale functions which where not working correctly before (but missed this due to an bad example tmx file). 

Re. to 2) Find the attached TMX file for testing. I had to rename it to txt because "tmx" file suffix isn't supported to be uploaded here. 
[test.tmx.txt](https://github.com/castle-engine/castle-engine/files/5721421/test.tmx.txt)

3) Thanks for the detailed explanation. I see where this is going and really like it. I hope I'm able to contribute in the future, but time will tell. 

Let me know, if something is still missing.
Have a nice day!

michaliskambi (approved):

michaliskambi:
Thank you! All looks good, merging :)

michaliskambi:
@Free-Pascal-meets-SDL-Website After testing, I made some changes:

1. I made `ObjectsVisible` by default true.

2. I changed the way coordinates are calculated. We do not stretch the map to always fill `RenderRect` (which was in your code) -- instead we place the map in the lower-left corner of the `RenderRect`. You can test it with e.g. examples/tiled/map_viewer/ . I added subroutine `RenderPosition` to handle this.

3. I have added a test file `examples/tiled/map_viewer/data/maps/desert_with_objects.tmx`, which contains the old test map desert.tmx + summed with your test objects. The objects (polylines etc.) should "stick" to the map tiles in the same way they stick to them in Tiled editor. In your original implementation, the objects would scale, unlike the underlying map tiles. Now it is good :)

Note that you can zoom and move the map in this example, and the objects should keep being attached to the tiles.

Zrzut ekranu z 2020-12-20 22-44-58
Zrzut ekranu z 2020-12-20 22-45-15

Issue #44, opened 2017-01-26
Allow to access to loader instance after calling TCastleSceneCore.Load()


Kagamma:
Now that we officially have tsr8's tiledmap loader in the engine code, I am planning on writing a tiledmap loader for X3D in the same way as how the engine handles non-X3D formats like 3DS or Spine. The problem is I also want to retain TTiledMap instance so that I can get informations that can't be translated directly to X3D nodes like the Object layer. So I am thinking of modifying TCastleSceneCore and add 2 new properties: 
- RetainLoaderInstance: A boolean field that tells the engine to keep loader instance after loading process is done, default is false.
- LoaderInstance: A TObject field that holds loader instance.

Of course we also need to modify all existing x3dloadinternal*.pas code to follow this changes.

What do you think Michalis?

michaliskambi:
Hm, I'm unsure about adding such fields to the TCastleSceneCore, when it's only for TMX format. If in the future it will make sense for other formats, then I'm going to change my mind and do exactly like you propose:)

But for now, I would prefer a different way. Make the future tiledmap -> X3D loader as two overloads:

~~~~
function LoadTiledMap(const URL: string): TX3DRootNode;
function LoadTiledMap(const URL: string; out Map: TTiledMap): TX3DRootNode;
~~~~

If you want to keep the internal map instance, then instead of using

~~~~
Scene.Load('map.tmx');
~~~~

you would use:

~~~~
Scene.Load(LoadTiledMap('map.tmx', Map), true);
~~~~

and you get a <code>Map: TTiledMap;</code> instance, to use as you like:)

Oh, and a minor consequence of this: the unit to load TiledMaps should then not have "Internal" in it's name:) So, it should be named like <code>X3DLoadTiledMap</code> instead of <code>X3DLoadInternalTiledMap</code>. Since a user may have a good need to use <code>X3DLoadTiledMap</code>, to be able to call <code>LoadTiledMap</code> directly.

BTW, one can usually push a lot of information into the X3D graph, by adding them to the "metadata" field of X3D nodes. But in this case, this would be unnecessarily convoluted, it's much easier to just get TTiledMap instance indeed.

PR #241, opened 2021-01-05
Add basic support for loading of Tiled maps as x3d scene

Free-Pascal-meets-SDL-Website:
This branch basically is a first try of a basic implementation of support for tiled maps as x3d scenes.
It originates from the discussion at #239. This pull request is not considered to be merged very soon into the engine but after some discussions and refinement it could perhaps.

This branch allows loading of a tmx file as known for scenes by
`Scene.Load('castle-data:/test.tmx');`

I figured this way it would be possible to load tmx file just as a scene to view it (e.g. in the viewer), otherwise, if this works perfectly fine with rendered tiles in the end, it could easily be integrated with the UIControl as discussed in #239. - However, I wonder about the connection between a x3d node (used for rendering) and the UI system, maybe you could point me to some examples?

Let me know please what you think about this route/code?

Major TODO: 
- render tiles, too

Minor TODO:
- add check routine to figure out if the tmx file is really a tiled map file, the correct MIME type is not sufficient as it is too general (xml/text) 
- some check is done in the tiled map loading routine though

Possible TODO:
- implement whole unit as a loader class as in X3DLoadInternalCocos2d.pas; benefit?

Later TODO:
- integrate with UI system

[test.tmx.txt](https://github.com/castle-engine/castle-engine/files/5772905/test.tmx.txt)


michaliskambi:
Great that you start this! Initial code looks cool.

1. We need to figure out a more specific MIME type for Tiled maps.

    Current `text/xml` you use is too generic. It could be easily confused with other XML formats. We already have some XML-based formats, like X3D XML and Starling sprite sheets XML, they use specialized MIME type. If there's nothing already established (TODO: search web), I would say to use `.tmx` extension -> maps to `application/x-tiled-map`. The MIME types mapping in CGE is right now trivially hardcoded in `CastleURIUtils` unit.

2. >  However, I wonder about the connection between a x3d node (used for rendering) and the UI system, maybe you could point me to some examples?

    Once you have `TX3DRootNode` instance, you just load it to `TCastleScene` using `MyScene.Load(Node, true { owns node })`. And then you have `TCastleScene`, which you can put in `TCastleViewport`, and  `TCastleViewport` is itself an UI control, that can be a child of any other UI control. `TCastleViewport` can be big and full-screen (like in most CGE examples), or it can be something small (even added to a button, like in `examples/component_gallery/`).

    So, the examples are all around CGE examples and manual, like https://castle-engine.io/manual_load_3d.php :) Once you can express Tiled map as X3D node, then you have `TCastleScene`, and then all CGE docs/examples about `TCastleScene` applies.

Kagamma:
My concern with TMX Loader is how can we access TTiledMap instance directly from TCastleScene, instead of having to explicitly call TTiledMap to load the map again just for additional information needed for the game (as mentioned before in #44).

I also made a basic TMX -> X3D converter before that can be used as reference, although it probably faster to read the document and implement it instead of dealing with ancient & ugly, uncommented code :)
https://github.com/Kagamma/cge-tiledmap

Free-Pascal-meets-SDL-Website:
>     1. We need to figure out a more specific MIME type for Tiled maps.
>        Current `text/xml` you use is too generic. It could be easily confused with other XML formats. We already have some XML-based formats, like X3D XML and Starling sprite sheets XML, they use specialized MIME type. If there's nothing already established (TODO: search web), I would say to use `.tmx` extension -> maps to `application/x-tiled-map`. The MIME types mapping in CGE is right now trivially hardcoded in `CastleURIUtils` unit.
> 
>     2. > However, I wonder about the connection between a x3d node (used for rendering) and the UI system, maybe you could point me to some examples?
>        
>        
>        Once you have `TX3DRootNode` instance, you just load it to `TCastleScene` using `MyScene.Load(Node, true { owns node })`. And then you have `TCastleScene`, which you can put in `TCastleViewport`, and  `TCastleViewport` is itself an UI control, that can be a child of any other UI control. `TCastleViewport` can be big and full-screen (like in most CGE examples), or it can be something small (even added to a button, like in `examples/component_gallery/`).
>        So, the examples are all around CGE examples and manual, like https://castle-engine.io/manual_load_3d.php :) Once you can express Tiled map as X3D node, then you have `TCastleScene`, and then all CGE docs/examples about `TCastleScene` applies.

1. I did the web reasearch prior to the coding. Nothing. I will go with your proposal.
2. Seems, I will look into this.
 
> I also made a basic TMX -> X3D converter before that can be used as reference, although its probably faster to read the document and implement it than dealing with ancient & ugly, uncommented code :)
> https://github.com/Kagamma/cge-tiledmap

I wasn't aware there was something like this already. I will definitivly look into this.

> My concern with TMX Loader is how can we access TTiledMap instance directly from TCastleScene, instead of having to explicitly call TTiledMap to load the map again just for additional information needed for the game (as mentioned before in #44).
> 

As I understand it (see #239), this will resolve automatically once the UIControl is just kind of a container of the TiledMap. The container UIControl holds the castle scene alongside with the TTiledMap instance. From there you can do whatever you want with both of them :-).

michaliskambi:
Note that you will need to solve a conflict with `x3dload.pas` unit. We did change an important thing there recently: the `LoadNode` overload wants to call loaders specifying not just `URL` of the file, but a pair `const Stream: TStream; const BaseUrl: String`. You will need to rework your `LoadTiledMap2d` to follow this convention:

- the map file contents should be loaded now from provided stream `Stream: TStream`. You shall not just open the URL from `BaseUrl`.

- The URL from `BaseUrl` can be used to resolve the related resources (e.g. if the map file references some textures relative to it).

Free-Pascal-meets-SDL-Website:
Hi Michaelis,
sorry for the delayed reply and fractured activity, heavy lack of time recently. 

Up to the latest commit, it is possible to load tmx files (ugly/memory inefficient solution though). But while working on this and the code got bigger, I'm working on a class based code base (which I didn't commit). I am unsure if this PR should be continued or if a new PR would be the better approach, atm.

Thanks for the hint about the stream loading. I will examine this.

Would it be possible to get some more time to find the best way to go forward? Be sure, I would come  back to you, as soon as I am decided.

michaliskambi:
@Free-Pascal-meets-SDL-Website Absolutely, and thank you for the work. 

There is no time rush, I'm fine with this PR being in draft state as long as it should to reach a good quality solution. I only asked because I was afraid that maybe you now wait for my action (review or advice at any point) and I'm causing a stall :) Just make sure to ping me if you need any help/advise around this.

I'm happy to hear you're working on it, I appreciate all the help with CGE we can get. Yeah, time constraints always suck :), I totally understand, no worries.

Free-Pascal-meets-SDL-Website:
Hi Michalis, 
after some time I was able to continue working on this. 

This PR now allows to load a tmx file (Tiled map) via Scene.Load() and generate internally a x3d scene graph, hence I changed the title. The resulting X3D graph can then be used to do whatever with it. I added a demo source (TestProject1 main and the state file).
**src folder**
[testproject1state.pas.txt](https://github.com/castle-engine/castle-engine/files/9128316/testproject1state.pas.txt)
[TestProject1.pas.txt](https://github.com/castle-engine/castle-engine/files/9128317/TestProject1.pas.txt)
**data folder**
[map64-final-test.tmx.txt](https://github.com/castle-engine/castle-engine/files/9128358/map64-final-test.tmx.txt)
[image3x3.tsx.txt](https://github.com/castle-engine/castle-engine/files/9128360/image3x3.tsx.txt)
image3x3

**Original**
OriginalMap
**Loaded/converted X3D Scene**
ConvertedMap

**Converted Scene and some tiles pushed in the back (shows functionality)**
ConvertedMap-Functionality

**Major todos for the future which not necessarily prevent this PR from being merged:**
- Ellipsoids support (see example above, is missing)
- Image layer support
- 90° rotated tiles support (= combination of vertical/horizontal flip and diagonal flip)
- Minor: Fix Warning about LineSet objects having less then 2 vertices (object layers)

Please have a look at this PR and let me know about what can be done to improve it to be merged. - The mentioned todos are, to my mind, smaller projects in themselves which I would not want to push into this PR but should be considered for the future. Exception is the minor todo (not sure atm. what causes the warning).

There is a debug mode which can be turned on if you set the `DebugMode` boolean in function `LoadTiledMap2d` to `true`. This has to be done in the code though. Maybe the debug mode source code should be removed from the converter, not sure on this.

Thanks and I see forward to your response :-)
Matthias

michaliskambi:
Great, thank you very much!

From a quick glance, it looks great, all the logic is inside `src/scene/x3dloadinternaltiledmap.pas` that returns X3D graph, and the rest of code just uses it. And it's generating X3D nodes based on `TTiledMap` information. This is perfect. It means also that Tiled maps will be fully functional also in CGE editor, if one adds TCastleScene with URL pointing to a Tiled map, and then one should be able to easily play with the map, placing other scenes on top of it.

So I will likely only have some minor notes :) 

1. I see that Jenkins and GitHub actions notice an error: `x3dloadinternaltiledmap.pas` is not part of `castle_base.lpk`. You can see it as failure on https://github.com/Free-Pascal-meets-SDL-Website/castle-engine/actions for your branch, and the exact error message is at https://github.com/Free-Pascal-meets-SDL-Website/castle-engine/runs/7379467034?check_suite_focus=true . Go ahead and add the unit to LPK. You can check locally, before committing, that it is fixed by 

    ```
    cd tools/internal/check_lazarus_packages/
    castle-engine compile --mode=release
    castle-engine run -- ../../../
    ```

    Once this works, the rest of Jenkins and GitHub actions automatic checks will go. That's good, we want Jenkins and GitHub actions to confirm that all automatic tests work before merging the PR.

2. About initial comment in `x3dloadinternaltiledmap.pas`:

    - Note that it is user-visible and processed by PasDoc, https://pasdoc.github.io/ . So I generally advise to keep "

michaliskambi:
Ups, very sorry for accidentally closing it. This was just a misclick during writing my comment above. Of course I reopened it.

michaliskambi (approved):

michaliskambi:
Big big thank you, and merged!

I did some testing, made some minor tweaks (details below) and merged.

I like this new approach to Tiled maps very much. Going forward, I would like this (_"load your Tiled maps into TCastleScene"_) to become our most feature-complete and advised way of using Tiled maps, eventually even deprecating our UI `TCastleTiledMapControl`.

This matches with CGE direction -- we add a lot of features around TCastleViewport / TCastleScene, we plan to optimize X3D nodes memory usage already (it will help e.g. in cases when one builds a 2D map using X3D nodes, with or without Tiled). And this way placing objects (additional TCastleScenes, like sprite sheets) on top of map is obvious and can be done in editor too.

I'm happy you want to work more on it :) My proposed TODOs would be:

A. Proper Z ordering of layers, so that objects don't "fight in Z" when rendered by default. Looking at your example code (doing `TileNode.Translation := TileNode.Translation + Vector3(0, 0, -1)` on mouse click) I guess you're already thinking of it :)

B. Support for other map types - hexagonal etc., like `TCastleTiledMapControl` does. The `examples/tiled/strategy_game_demo/` has samples of all types.

C. In general, I would say to test on sample maps in both projects under `examples/tiled/` in CGE and see if you have all the functonality to create something like `examples/tiled/strategy_game_demo/` but now using TCastleViewport / TCastleScene with Tiled maps.

Some minor notes (including about my changes):

1. I extended `LoadScene_FileFilters` to include *.tmx file filter.

    Now I can comfortably load Tiled maps in view3dscene ( https://github.com/castle-engine/view3dscene ), CGE editor and all other CGE viewers.

    TODO for me: I should introduce a proper registration procedure for node formats, so that file filters are automatically updated and there's no need to modify currently-hardcoded conditions in `LoadNode` that test for supported MIME types.

2. I did a small fix ( https://github.com/castle-engine/castle-engine/pull/241/commits/ac29076399de9c7e73ec0173baf06f42f8ed2a87 ) and now I can open example maps in `examples/tiled/...` too. This also makes easier testing.

    Zrzut ekranu z 2022-07-27 14-35-03

    Note that there are additional borders visible on `desert.tmx` map below, this is normal. Strictly speaking, it's an error in the tileset image. The UI control has a hack `SmoothScalingSafeBorder` to avoid it, though admittedly it is a hack, ideally tilesets should be prepared with a safe border between tiles to avoid that.

4. PasDoc requires parenthesis around multi-line `@item`. Like

    ```
    @item(xxx
      yyy
      zzz)
    ```

    You only omit parenthesis for single-line usage, see https://pasdoc.github.io/TagsParametersWithoutParenthesis .


5. A minor style note: we generally in CGE compare references by `XXX <> nil` not `Assigned(XXX)`. We use `Assigned` only for callbacks. Of course it's exactly equivalent, and it's totally minor issue, just making a note for future.

6. As for object creation in

    ```
    TiledMapFromStream := TTiledMap.Create(Stream, BaseUrl);
    if not Assigned(TiledMapFromStream) then
      Exit;
    try
      TiledMapConverter := TTiledMapConverter.Create(TiledMapFromStream, DebugMode);
      TiledMapConverter.ConvertMap;
      Result := TiledMapConverter.RootNode;
    finally
      if Assigned(TiledMapFromStream) then
        FreeAndNil(TiledMapFromStream);
      if Assigned(TiledMapConverter) then
        FreeAndNil(TiledMapConverter);
    end;
    ```

    A few notes:

    - Checking `if Assigned(Xxx) then` before `FreeAndNil` is not necessary. `FreeAndNil(Xxx)` does this automatically (by doing `Xxx.Free` which contains `if Self <> nil`...).

    - Checking `if not Assigned(TiledMapFromStream) then` after creating `TiledMapFromStream := TTiledMap.Create` is not necessary. If there was no exception, the `TiledMapFromStream` is guaranteed to be non-nil.

    - In the above code, if there will be an error during `TTiledMapConverter.Create` constructor then the value of `TiledMapConverter` is undefined. It may be not `nil` -- it's a local variable, it can have memory garbage (non-nil but also invalid pointer). You could initialize it earlier to `nil` explictly, or just make nested try..finally, see https://castle-engine.io/modern_pascal_introduction.html#_finally_doing_things_regardless_if_an_exception_occurred .

    - You also don't need `Result := nil;` if all code paths (that don't raise exception) guarantee to set `Result`.

    So I simplified it to

    ```
    function LoadTiledMap2d(const Stream: TStream; const BaseUrl: String
      ): TX3DRootNode;
    var
      TiledMapFromStream: TTiledMap;
      TiledMapConverter: TTiledMapConverter;
    begin
      { The Tiled converter unit expects a TTiledMap object instance,
        hence create one. }
      TiledMapFromStream := TTiledMap.Create(Stream, BaseUrl);
      try
        TiledMapConverter := TTiledMapConverter.Create(TiledMapFromStream, TiledDebugMode);
        try
          TiledMapConverter.ConvertMap;
          Result := TiledMapConverter.RootNode;
        finally FreeAndNil(TiledMapConverter) end;
      finally FreeAndNil(TiledMapFromStream) end;
    end;
    ```

7. I totally agree with you about the `TObjectGroupLayer` / `RenderObjects` comment you added at `TCastleTiledMapControl`. Feel welcome to rework it and remove `ObjectsVisible` from `TCastleTiledMapControl`, so that rendering of `TObjectGroupLayer` just happens in the same loop as "regular" map layers.

8. I exposed the `TiledDebugMode` global variable.

    I'm fully OK and encourage such debug visualization -- it is useful to you, it will undoubtedly be useful to anyone else working on this code. Of course feel free to change this debug visualization to whatever seems comfortable (right now maybe it's indeed too much -- but I leave this in your hands, of course make as much debug visualization as is helpful to you).

    So by using global variable, we enable it to be used by others, without modifying CGE :) Others can just use `X3DLoadInternalTiledMap` unit and set `TiledDebugMode`. As the unit is marked "internal", this is undocumented and not guaranteed to be kept forever... but you can keep using it/changing as you need for the development of this.

Free-Pascal-meets-SDL-Website:
Thank you very much for tweaking and merging this PR! I'm really glad to see this feature now being natively implemented in the engine. Also this makes improving in an CI way simpler. 

> A. Proper Z ordering of layers, so that objects don't "fight in Z" when rendered by default. Looking at your example code (doing `TileNode.Translation := TileNode.Translation + Vector3(0, 0, -1)` on mouse click) I guess you're already thinking of it :)
> 

From a layer perspective this is already realized. The objects (tiles or tiled objects) of a certain layer will not Z-fight with objects of another layer. Within a certain layer this should be improved, I agree.

The example you are quoting was actually to prove that it is (easily) possible to extract a certain "tile node" from the scene to manipulate it in a desired way. This is what most users will want to do: Load the scene created in Tiled and be able to respond to on-clicks on a certain tile. 

> Some minor notes (including about my changes):
> 
> 1. I extended `LoadScene_FileFilters` to include *.tmx file filter.
>        Now I can comfortably load Tiled maps in view3dscene ( https://github.com/castle-engine/view3dscene ), CGE editor and all other CGE viewers.

I admittedly neglected integration with view3dscene but I'm happy to try it out now that it works :-)!

> 2. I did a small fix ( [ac29076](https://github.com/castle-engine/castle-engine/commit/ac29076399de9c7e73ec0173baf06f42f8ed2a87) ) and now I can open example maps in `examples/tiled/...` too. This also makes easier testing.

Thanks! I missed that!

About topics 3. - 6.: Thanks for these very instructive/encouraging hints! Especially the information about the try-block were eye opening.
 
>7. I exposed the `TiledDebugMode` global variable.
>        I'm fully OK and encourage such debug visualization -- it is useful to you, it will undoubtedly be useful to anyone else working on this code. Of course feel free to change this debug visualization to whatever seems comfortable (right now maybe it's indeed too much -- but I leave this in your hands, of course make as much debug visualization as is helpful to you).
>        So by using global variable, we enable it to be used by others, without modifying CGE :) Others can just use `X3DLoadInternalTiledMap` unit and set `TiledDebugMode`. As the unit is marked "internal", this is undocumented and not guaranteed to be kept forever... but you can keep using it/changing as you need for the development of this.

Great! Yes this is better, not sure why I didn't expose this variable globally in the first place.
I like your approach about the debug vis. and you are right btw., it is too much atm. as it was most helpful in the beginning of the converter class creation. It may get useful in the future again, though.

PR #243, opened 2021-01-26
Check if component name is valid when renaming

eugeneloza:
Fixes https://trello.com/c/kvAMJ3x2/116-it-is-possible-to-rename-component-in-castle-editor-to-empty-name-in-hierarchy

### DIFF
--- a/tools/castle-editor/code/framedesign.pas
+++ b/tools/castle-editor/code/framedesign.pas
@@ -344,7 +344,7 @@ implementation
 
 uses // use Windows unit with FPC 3.0.x, to get TSplitRectType enums
   {$ifdef VER3_0} {$ifdef MSWINDOWS} Windows, {$endif} {$endif}
-  TypInfo, StrUtils, Math, Graphics, Types, Dialogs, LCLType,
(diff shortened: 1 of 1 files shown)


michaliskambi:
Cool, thanks for doing this. I added a comment explaining the rationale, merging :)

My first reaction was to say that "when changing `Name`, usually the `TComponent.SetName` should do the job of checking validity", but I see in FPC sources now:

```pascal
Procedure TComponent.SetName(const NewName: TComponentName);

begin
  If FName=NewName then exit;
  If (NewName<>'') and not IsValidIdent(NewName) then
    Raise EComponentError.CreateFmt(SInvalidName,[NewName]);
...
```

So they explicitly make setting `Name` to empty OK. And it makes sense, it is a valid and even default name, and is OK if you don't care to uniquely identify it.

I was curious how did they make the object inspector (TOICustomPropertyGrid) catch it, and I see in `TComponentNamePropertyEditor.SetValue` they do the same as you:

```pascal
  if not IsValidIdent(NewValue) then
    raise Exception.Create(Format(oisComponentNameIsNotAValidIdentifier, [NewValue]));
```

So here they don't allow empty Name.

michaliskambi (approved):

Issue #244, opened 2021-01-26
CGE editor is not visible on Windows 10 taskbar once you open the project


michaliskambi:
That's because we hide the main application form once we open the project. The application is still available on Alt + Tab list, but the Windows 10 taskbar shows like the application is not open.

We had a hack to fix it, but this hack was causing even worse (at least sometimes) problems, in that the main form was visible when it was not supposed to (and sometimes even on top of project window). So the hack was removed in https://github.com/castle-engine/castle-engine/commit/1d1bd7816246915abe22b2dad0e5ec57beb25ed9 .

Looking for a proper solution :)

PR #245, opened 2021-02-07
Fix system tray icon hidden on non-main forms on Windows 10

eugeneloza:
Fixes https://github.com/castle-engine/castle-engine/issues/244


michaliskambi:
Oh they have a dedicated property to fix it in LCL I see `ShowInTaskBar`. Thank you!

Issue #216, opened 2020-10-25
CGE Editor - Remember Window State + Re-Open last session


peardox:
A couple of simple but desirable suggestions for the editor...

1) Remember the Window State of the last session. Basically if you had editor maximised in your previous session you probably want it in this session (etc) so start it the same as you last used it. OK - this is one click BUT that's not a reason the feature should be missing. This one is simply for user convenience.

2) Slightly more contentious I guess... Remember the design you were editing, the viewport settings etc from your last session and re-open where you last were. Presently you have to Re-open project, open design, click file and probably position things as you like. Lazarus remembers just about everything about your previous session - what files were in the editor, what tab you were editing, which form you were designing, it even remembers the positions of your palettes etc (does not apply to editor ATM)

3) A don't load past session checkbox on the main load interface - just in case, results in the current operation of editor.

I fully expect comments about it "not being worth doing" on these - I would argue that this kind of attention to detail is what makes the difference between a product and an Alpha test. 

So what if the user has to click five things to get back to their previous state? My question is why should they have to in the first place? Saving the user any unrequired action is always worth the effort.

michaliskambi:
I agree about 1 and 2.

AD 1 - Should be in user-specific settings (`ApplicationConfig` inside editor). It has to be done carefully -- to preserve the state of the window (maximized or not), not only the window dimensions. It also has to store current monitor and size (LCL has ways to access it by `Screen`) and if they change, then last saved window dimensions should possibly be adjusted (or discarded?).

AD 2 - Should be in project directory (`CastleEngineUserPreferences.xml` with comments to discourage putting this in version control). It would be possible to instead store it in editor settings (like a map project path -> project settings) which is simpler, but ultimately worse solution. 

AD 3 - I don't see this as necessary. 

- Blender has something this, but only because they mixed the "user preferences" inside .blend file with "content of 3D model".

- Lazarus doesn't have it, and that's cool, you can just `rm -f *.lps` to have a clean state.

- So, just by doing AD 2 correctly (`CastleEngineUserPreferences.xml` in project dir) we effectively make AD 3 not needed. If you screwed your project, just `rm -f CastleEngineUserPreferences.xml`.

michaliskambi:
As for AD 2 - to be clear, I propose `CastleEngineUserPreferences.xml` fully knowing that it would be a very trivial file at start, with just a single "project-relative path to currently open design". But this file means we have "room to grow" and introduce more project-specific and user-specific stuff.

PR #246, opened 2021-02-07
Save castle editor window state

eugeneloza:
Save Castle Editor window state (maximized/minimized, width/height) and some internal parameters (three splitters positions) in UserConfig and load them upon reloading of the Form.

Fixes the first part of https://github.com/castle-engine/castle-engine/issues/216

### DIFF
--- a/tools/castle-editor/code/formproject.lfm
+++ b/tools/castle-editor/code/formproject.lfm
@@ -11,6 +11,8 @@ object ProjectForm: TProjectForm
   OnCloseQuery = FormCloseQuery
   OnCreate = FormCreate
   OnDestroy = FormDestroy
+  OnHide = FormHide
+  OnShow = FormShow
   Position = poMainFormCenter
   ShowInTaskBar = stAlways
   LCLVersion = '2.0.8.0'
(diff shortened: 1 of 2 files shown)


eugeneloza:
@michaliskambi also note that here I used the "default values" from current LFM files - which for me (1366x768) go a bit off screen, we might want to adjust those to something small - visible on all monitors or define them based on `Screen.Height` and `Screen.Width`.

michaliskambi:
- I think we store too many values . Do we need to store both `Width` and `ClientWidth` in many cases? Seems like only one of them should be necessary. If really both must be saved/restored then setting their order must be carefully documented, as one of them sets also the other.

- Do not store window state as Minimized or Fullscreen, it would be unexpected to restore in minimized state. Store only normal vs maximized.

- When restoring, make sure to not restore at all if these values were not recorded in settings file. E.g. use a boolean like `ProjectForm_Saved`. We have already a good default behavior on all screen sizes thanks to LCL form settings, we do not want to break them. Neither hardcoding them nor basing on `Screen` should be necessary here, the default behavior is just good already.

- When restoring, check that form doesn't go offscreen. In case user e.g. changes monitors.

eugeneloza:
All seems to be fixed! :)

michaliskambi (approved):
All looks cool, thank you!

PR #251, opened 2021-02-25
Fix ExtractURIPath() to work the same on all platforms

and3md:
Currently on linux it removes filename with anchors (I think works as expected) but on windows its stops on last : in anchor URL.

I used `URIDeleteAnchor` with `RecognizeEvenEscapedHash=true` because GTK2 open dialog sometimes returns paths with # changed to `% code`. So i think this is more secure.

### DIFF
--- a/src/files/castleuriutils.pas
+++ b/src/files/castleuriutils.pas
@@ -1116,7 +1116,7 @@ end;
 
 function ExtractURIPath(const URL: string): string;
 begin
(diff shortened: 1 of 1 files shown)


michaliskambi:
Thank you! This is a good solution IMO, and I will actually want to extend it to some other routines in CastleURIUtils :)

michaliskambi (approved):

PR #252, opened 2021-02-26
Allows to set transform feedback varyings

Kagamma:
This PR allows to set transform feedback varyings before linking a GLSL program.
This will be ignored if OpenGL ES is detected, and raise EGLSLTransformFeedbackError if OpenGL version is < 3.0

### DIFF
--- a/src/images/opengl/castleglshaders.pas
+++ b/src/images/opengl/castleglshaders.pas
@@ -314,7 +314,7 @@ type
     procedure DetachAllShaders;
 
     { Specify values to record in transform feedback buffers.
-      This must be called before @link(Link) method }
+      This must be called before @link(Link) method. }
     procedure SetTransformFeedbackVaryings(const Varyings: array of PChar; const IsSingleBufferMode: Boolean = True);
 
     { Link the program, this should be done after attaching all shaders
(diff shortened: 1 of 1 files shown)


src/images/opengl/castleglshaders.pas
```diff
@@ -1778,6 +1783,34 @@ procedure TGLSLProgram.AttachGeometryShader(const S: string);
   AttachShader(stGeometry, S);
 end;
 
+procedure TGLSLProgram.SetTransformFeedbackVaryings(const Varyings: array of PChar; const IsSingleBufferMode: Boolean);
+var
+  TransformFeedbackBufferMode,
+  ErrorCode: TGLuint;
+  VaryingLength: Cardinal;
+begin;
+  {$ifndef OpenGLES}
```
michaliskambi, line 1792: Since we make an exception on desktops when transform feedback is not supported, it seems consistent to also make an exception on OpenGLES (always, when `VaryingLength > 0`) like

```pascal
 raise EGLSLTransformFeedbackError.Create('Transform feedback not supported by OpenGLES');
```

michaliskambi (changes requested):
Thank you! Aside from one suggestion, this looks cool.

Additional question (but it doesn't have to be in this PR): can you prepare a simple example of usage? E.g. in `examples/3d_rendering_processing/transform_feedback/` . I understard your GPU particles will also use it and thus provide an example.

Kagamma:
Thanks. I will add an example in the next PR (probably around this weekend).

michaliskambi (approved):

michaliskambi:
Thank you!

PR #253, opened 2021-02-26
Do not set Focused state for TCastleUserInterfaces on Touch devices

eugeneloza:
To avoid setting them into focused state after clicking a UI element in the same position (in this or previous state)


michaliskambi (approved):

michaliskambi:
Cool, simple enough :) Thank you!

PR #254, opened 2021-02-27
TCastleTheme.Images as TCastleImagePersistent

eugeneloza:
Use `TCastleImagePersistent` instead of `TCastleImage` in `TCastleTheme`, e.g. to allow pixelart splash screens, etc.

Also fix a bug with `TCastleImagePersistent.SmoothScaling` not being applied if set runtime.

Should fix https://trello.com/c/lNJYLskZ/129-theme-images-as-tcastleimagepersistent


michaliskambi:
I updated https://castle-engine.io/manual_2d_user_interface.php#section_theme about it.

PR #255, opened 2021-02-27
ClearExceptions(false) on TCastleApplication.CastleEngineInitialize

eugeneloza:
This should fix Xiaomi Mi 9 SE ARM64 bug that injects dangling external exceptions into the app. Note that this bug might also be related to the NS bug in TCasScriptExpression.Execute - and the workaround applied may fix it too.

Note that it was NOT TESTED on the Xiaomi phone - it was made based on a workaround that worked on the device.

### DIFF
--- a/src/window/castlewindow.pas
+++ b/src/window/castlewindow.pas
@@ -2963,15 +2963,15 @@ procedure TCastleWindowBase.OpenCore;
 
     {$IFDEF android}
     { Workaround an ARM64 Android-specific bug which manifests on some devices
-      where some "external" exception flags are "injected" into unrelated app
-      when it is launched/initialized
-      Reproduced on Xiaomi MI 9 SE
+      (reproduced on Xiaomi MI 9 SE) that creates a dangling EInvalidOp after calling Theme.Draw.
+      It doesn't seem to interfere with normal porgram workflow,
+      neither it causes any visible graphic glitches in the image rendered by Theme.Draw,
+      but causes ClearExceptions(true) to raise unrelated dangling exceptions,
+      which in turn makes e.g. CastleScript misbehave.
       Also seems to be reproduced outside of Castle Game Engine:
       https://forum.lazarus.freepascal.org/index.php/topic,42933.msg318965.html?#msg318965
-      These do not seem to interfere with normal porgram workflow,
-      but cause ClearExceptions(true) to raise unrelated dangling exceptions,
-      which in turn makes e.g. CastleScript misbehave.
-      It seems that clearing them once is perfectly enough. }
+      It seems that clearing them once per app run is perfectly enough,
+      all subsequent calls to Theme.Draw do not raise the exception. }
     ClearExceptions(false);
     {$ENDIF}
   end;
--- a/src/window/castlewindow.pas
+++ b/src/window/castlewindow.pas
@@ -2960,6 +2960,20 @@ procedure TCastleWindowBase.OpenCore;
 
     // just like TCastleWindowBase.DoRender
     if DoubleBuffer then SwapBuffers else glFlush;
+
+    {$IFDEF android}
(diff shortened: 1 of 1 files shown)


michaliskambi:
1\. Hm, looking at a place where `Application.CastleEngineInitialize` is called, this gives me some idea: maybe we can localize the place that "sets the exception" better in the end.

 `Application.CastleEngineInitialize` is called from `TCastleWindowBase.OpenCore` *after* OpenGL(ES) context is initialized. And initializing it I guess has a high chance of setting some floating-point error, since OpenGL(ES) implementations on various systems do it. Can you make a test with modified `TCastleWindowBase.OpenCore` to place

```pascal
WritelnLog('before xxx'); 
ClearExceptions(true);
xxx
WritelnLog('after xxx'); 
ClearExceptions(true);
```

... around some things it calls:

- `OpenBackend`
- `MakeCurrent`
- `GLInformationInitialize`
- `RenderLoadingBackground`
- oh, and make sure it doesn't enter the `if GLVersion.BuggyDepth32 ...` condition on that Android device.

Maybe we will be able to pinpoint where exactly is this exception set.

2\. I still cannot understand why

```
SetExceptionMask([exInvalidOp, exDenormalized, exZeroDivide, exOverflow, exUnderflow, exPrecision])
```

doesn't help in this case. Can you test it again with Vlada? To make sure it was not some mistake in testing?

eugeneloza:
1. Yes, sure, I'll make the tests.

2. > Can you test it again with Vlada? To make sure it was not some mistake in testing?

Vlada made a test today (with SVN version through Jenkins - no mistakes are possible, see 3579) - setting `Math.SetExceptionMask([exInvalidOp, exDenormalized, exZeroDivide, exOverflow, exUnderflow, exPrecision]);` in `initialization` still crashes.

eugeneloza:
Done! Asked Vladlena to test.

eugeneloza:
Thanks to Vladlena for testing:

Clipboard 1

So it fails before `Application.CastleEngineInitialize` i.e. after `RenderLoadingBackground`. Note that the first log is actually the middle one - it "passed", and doublechecking: yes, it passes always if the player reenters the application without closing it and the game starts fine - i.e. clearing exceptions _once_ is enough - second time the same procedure doesn't cause an exception.

There doesn't seem to be anything specific inside `RenderLoadingBackground`, maybe some rendering thing (like `RenderContext.Clear`) grabs external exceptions when it's not supposed to?

eugeneloza:
Vladlena made a video of loading - there are no visual bugs or something. I.e. render of the `RenderLoadingBackground` goes perfectly as expected.

michaliskambi:
> There doesn't seem to be anything specific inside RenderLoadingBackground, maybe some rendering thing (like RenderContext.Clear) grabs external exceptions when it's not supposed to?

I do not really suspect "grabbing external exceptions", I more suspect that just "some OpenGL operation sets floating-point exception". Because this is exactly what happens on desktops, it's known (but there `SetExceptionMask` helps).

Hm, we could go deeper into `RenderLoadingBackground` but probably it takes too much time :) Can you test a slightly modified version of it, with `ClearExceptions(false);` done right at the end of `RenderLoadingBackground` call, and not inside `CastleEngineInitialize`? If this will work, I'm happy with that :)

eugeneloza:
> but probably it takes too much time

Not that much, if we need it, I'll do it :D

Meanwhile will prepare a 

> test a slightly modified version of it, with ClearExceptions(false); done right at the end of RenderLoadingBackground call, and not inside CastleEngineInitialize

michaliskambi:
If the `ClearExceptions(false)` works when done at the end of `RenderLoadingBackground`, then I think this is enough. `Theme.Draw` does a number of OpenGL(ES) things. I am curious... but it will probably not help us much knowing which exactly OpenGL(ES) instruction makes the error.

michaliskambi (approved):

PR #258, opened 2021-03-01
Better component base name for Castle Editor

eugeneloza:
* Moves ProposeName logic into TRegisteredComponent
* Enables manual specification of TRegisteredComponent.ComponentNameBase
* Ensures that under no circumstances ProposeName starts with a digit

### DIFF
--- a/src/3d/castlecomponentserialize.pas
+++ b/src/3d/castlecomponentserialize.pas
@@ -97,14 +97,14 @@ type
   TRegisteredComponent = class
   strict private
     FComponentNameBase: String;
-    { Constructs a reasonable ComponentBaseName based on ComponentClass.ClassName }
-    procedure ConstructComponentBaseName;
-    { Tests if the ComponentBaseName is valid and corrects it }
+    { Constructs a reasonable ComponentNameBase based on ComponentClass.ClassName }
+    procedure ConstructComponentNameBase;
+    { Tests if the ComponentNameBase is valid and corrects it }
     procedure ValidateComponentNameBase;
-    { Gets ComponentBaseName or constructs one if it is empty }
-    function GetComponentBaseName: String;
-    { Sets the ComponentBaseName and silently corrects it if it is invalid }
-    procedure SetComponentBaseName(const Value: String);
+    { Gets ComponentNameBase or constructs one if it is empty }
+    function GetComponentNameBase: String;
+    { Sets the ComponentNameBase and silently corrects it if it is invalid }
+    procedure SetComponentNameBase(const Value: String);
   public
     { Class of the component. Never leave this @nil. }
     ComponentClass: TComponentClass;
@@ -117,7 +117,7 @@ type
     IsDeprecated: Boolean;
     { Base for naming this component instance,
       e.g. to be used in Castle Editor for naming components }
-    property ComponentNameBase: String read GetComponentBaseName write SetComponentBaseName;
+    property ComponentNameBase: String read GetComponentNameBase write SetComponentNameBase;
   end;
   TRegisteredComponents = {$ifdef CASTLE_OBJFPC}specialize{$endif} TObjectList<TRegisteredComponent>;
 
@@ -192,7 +192,7 @@ begin
     FComponentNameBase := 'Component' + FComponentNameBase;
 end;
 
-procedure TRegisteredComponent.ConstructComponentBaseName;
+procedure TRegisteredComponent.ConstructComponentNameBase;
 begin
   FComponentNameBase := ComponentClass.ClassName;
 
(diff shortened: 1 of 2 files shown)


tools/castle-editor/code/framedesign.pas
```diff
@@ -3033,28 +3033,10 @@ procedure TDesignFrame.SetParent(AParent: TWinControl);
...
-    ResultBase := PrefixRemove('T', ResultBase, true);
-
-  // remove 2D, as component name cannot start with that
-  if IsPrefix('2D', ResultBase, true) then
-    ResultBase := PrefixRemove('2D', ResultBase, true);
-
-  // in case the replacements above made '', fix it (can happen in case of TCastleUserInterface)
-  if ResultBase = '' then
-    ResultBase := 'Group';
+  ResultBase := FindRegisteredComponentByComponentClass(ComponentClass).ComponentNameBase;
```
michaliskambi, line 3039: This makes the `ProposedName`, and so also `AddComponent` that calls it, work only with components that are on the `RegisteredComponents` list.

While it is indeed right now the only use-case, but 

1. If we decide to go this way, we could make it simpler, and `ProposedName` and `AddComponent` could be called with just a `const R: TRegisteredComponent` parameter. We have this information, right now `AddComponent` is called like this:

```
procedure TDesignFrame.MenuItemAddComponentClick(Sender: TObject);
var
  R: TRegisteredComponent;
begin
  R := TRegisteredComponent(Pointer((Sender as TComponent).Tag));
  AddComponent(R.ComponentClass, R.OnCreate);
end;
```

   So it "deconstructs" the `TRegisteredComponent`, while your code would make it connected again.

2. But I *do not* think we should go this way. Better make `ProposedName`, and so also `AddComponent`, work with any class, not necessarily registered one -- for the future.

3. Also, `FindRegisteredComponentByComponentClass` is not really something we can implement. Some classes are present multiple times, just with different `TRegisteredComponent.OnCreate` value (see how `castlescene.pas` registers components like `Scene (Primitive: Box)`).

4. Points 1-3 mean that we should take `const R: TRegisteredComponent` parameter in `AddComponent` and `ProposeName`, and accept it may be `nil` also, and just use the base `ClassName` if `R` is `nil`. I say that this is too unnecessary complication. Let's resign from it, and do not modify `TRegisteredComponent` for it.

src/3d/castlecomponentserialize.pas
```diff
@@ -170,6 +184,54 @@ implementation
...
+  if IsPrefix('2D', FComponentNameBase, true) then
+    FComponentNameBase := PrefixRemove('2D', FComponentNameBase, true) + '2D';
+  if IsPrefix('3D', FComponentNameBase, true) then
+    FComponentNameBase := PrefixRemove('3D', FComponentNameBase, true) + '3D';
+
+  ValidateComponentNameBase;
+
+  // in case the replacements above made '', fix it (can happen in case of TCastleUserInterface)
+  if FComponentNameBase = '' then
+    FComponentNameBase := 'Group';
```
michaliskambi, line 219: This should be done before the `ValidateComponentNameBase`. Otherwise, it will never run (`ValidateComponentNameBase` will turn '' into 'Component').

src/3d/castlecomponentserialize.pas
```diff
@@ -170,6 +184,54 @@ implementation
...
+begin
+  if FComponentNameBase = '' then
+    ConstructComponentNameBase;
+  Result := FComponentNameBase;
+end;
+
+procedure TRegisteredComponent.SetComponentNameBase(const Value: String);
+begin
+  FComponentNameBase := Value;
+  ValidateComponentNameBase;
```
michaliskambi, line 232: It's a bit complicated that getter does some job (ConstructComponentNameBase), and setter does some job (ValidateComponentNameBase). I would much prefer to just have `ComponentNameBase` a simple field, and move the logic into a complete routine like `GetFinalComponentName`.

michaliskambi (changes requested):
Hmm, I found 2 things I'd like to change, which in effect change the way it should be implemented completely.

I'm sorry! Do you mind if I reject this PR, and do it simpler myself, using your code? I think I really would like to just use your lines

```pascal

 if IsPrefix('3D', FComponentNameBase, true) then
    FComponentNameBase := PrefixRemove('3D', FComponentNameBase, true) + '3D';

...

  if SCharIs(FComponentNameBase, 1, ['0'..'9']) then
    FComponentNameBase := 'Component' + FComponentNameBase;
```

... in the existing `ProposeName`, and they will make the issue fixed simpler.

tools/castle-editor/code/framedesign.pas
eugeneloza: Sure! If there is a more simple way to - great :)

src/3d/castlecomponentserialize.pas
eugeneloza: ValidateComponentNameBase should turn into 'Component' only if first symbol is digit, AFAIU

michaliskambi: Ups, yeah, that was invalid statement from my side. `ValidateComponentNameBase` would leave '' unchanged.

PR #262, opened 2021-03-04
FMOD service on android

eugeneloza:
I'm not sure how nice this solution is, so creating draft for now. @michaliskambi let's think of if we need to improve this. I see at least 4 problems here:

1. We create ad-hoc solution which mimics iOS behavior, but works differently, and differently than other Android services.
2. It asks for a different syntax for service definition. I already saw there is a comment on the wiki about "we should make it more universal" - in this commit it's otherwise.
3. As we have *two* dynamic library files (`libfmod.so` and `libfmodL.so`), we create *two* services. Maybe I'm doing something wrong here?
4. The path where the libraries are copied to is hardcoded. It mimics how the main library is handled, but other services don't do that.


michaliskambi:
> 3. As we have two dynamic library files (libfmod.so and libfmodL.so), we create two services. Maybe I'm doing something wrong here?

Simply remove fmodL for now, IMHO. We don't need in practice this complication, and we can avoid packaging 2 libraries when only 1 would be used.

If we decide in the future that we need "L" (lighter) version of FMOD, it should be a parameter in the manifest. I mean, a single service should work with both "L" or non-L version, and only the chosen one should be packaged.

> 4. The path where the libraries are copied to is hardcoded. It mimics how the main library is handled, but other services don't do that.

You mean the hardcoded output to `'app' + PathDelim + 'src' + PathDelim + 'main' + PathDelim + 'jni' + PathDelim` ? Yeah, it would be nicer to do this without any modification to `toolandroid.pas`, but for now this is fine :)

michaliskambi:
> If we decide in the future that we need "L" (lighter) version of FMOD, it should be a parameter in the manifest. I mean, a single service should work with both "L" or non-L version, and only the chosen one should be packaged.

Note that I was wrong in this paragraph. The "L" stands for "logging" version of FMOD. It is *not* something "lighter" in any way. Only the logging version reacts to FMOD_Debug_Initialize ( https://www.fmod.com/resources/documentation-api?version=2.1&page=core-api-common.html#fmod_debug_flags ).

I confused in my head this "L" with 2 Tremolo versions ( https://github.com/castle-engine/android-tremolo ), they have "low-precision" version for faster (and low quality) OggVorbis decompression. It is unrelated to FMOD "L" version.

Still my course was right -- let's not create separate classes/code to handle the "L" version. If anything, it could be requested by a service parameter some day. In practice, one can also just rename `libfmodL.so` -> `libfmod.so` for temporary debug purposes.

michaliskambi:
I did a few things:

1.  added copying of `fmod.jar`

2. added calling special FMOD init/close methods in Java

3. allowed FMOD to open Android assets (so reading e.g. wav or OggVorbis is done directly by FMOD)

4. called `AttachCurrentThread`

5. added InitializeFmodLibrary call for Android

1-4 are done following https://www.fmod.com/resources/documentation-api?version=2.1&page=platforms-android.html . I think I did everything described there as necessary now :)

michaliskambi (approved):

PR #264, opened 2021-03-06
Adds transform feedback example

Kagamma:


examples/3d_rendering_processing/transform_feedback.lpi
```diff
@@ -0,0 +1,236 @@
...
+        <Filename Value="../../../../../sdks/fpc/fpcsrc/packages/opengl/src/glext.pp"/>
+        <UnitName Value="GLext"/>
+        <EditorIndex Value="-1"/>
+        <WindowIndex Value="1"/>
+        <TopLine Value="4877"/>
+        <CursorPos X="3" Y="4896"/>
+        <UsageCount Value="10"/>
+      </Unit3>
+    </Units>
+    <JumpHistory Count="28" HistoryIndex="27">
```
michaliskambi, line 89: Please configure your Lazarus to not store "session information" inside the LPI file (store it anywhere else -- e.g. in separate LPS file, or in Lazarus config dir), and resave this project. This way the LPI file will have less information, and will not have "meaningless" modifications whenever someone opens it.

examples/3d_rendering_processing/transform_feedback.lpr
```diff
@@ -0,0 +1,135 @@
...
+
+  ----------------------------------------------------------------------------
+}
+
+{ Example how to use TGLSLProgram.SetTransformFeedbackVaryings }
+
+program transform_feedback;
+
+{$macro on}
+{$define nl:=+ LineEnding +}
```
michaliskambi, line 21: Please use `NL` constant from `CastleUtils` instead, without a macro, like 

```pascal
`xxx` + NL +
...
```

It looks only a little longer than your

```pascal
`xxx`nl
...
```

but it more obvious and consistent with other CGE code.

examples/3d_rendering_processing/transform_feedback.lpr
```diff
@@ -0,0 +1,135 @@
...
+'out vec4 outColor;'nl
+'void main() {'nl
+'  outColor = vec4(1.0);'nl
+'}';
+
+var
+  Window: TCastleWindowBase;
+  Viewport: TCastleViewport;
+  VAOs,
+  VBOs: array[0..1] of GLint;
```
michaliskambi, line 63: We usually place variable declarations of the same type on the same line in CGE, so make it

```
VAOs, VBOs: array[0..1] of TGLint;
```

We usually write declarations like that. In some exceptions we make it multi-line but then we indent the 2nd line, like this:

```pascal
  SomeVariable, SomeLongLongVariableName,
    SomeReallyReallyLongVairiableName: TSomething;
```

This way it is easier to "parse" the variable declarations by a human eye. I fully know that it is only a matter of taste :) -- so in this case it's just better to follow existing conventions of CGE.

Also use `Txxx` types, so `GLint` -> `TGLint`.

michaliskambi (changes requested):
Thank you! I have small suggestions to address :)

michaliskambi (approved):

michaliskambi:
Thank you! All cool :)

Issue #261, opened 2021-03-04
Origin not centered using an IsoMetric viewport


peardox:
It's the lastest version of sample project : https://github.com/peardox/IsometricTest - needs graphic assets adding (instructions in the README.md)

![origin](https://user-images.githubusercontent.com/1549219/109960791-1d89fb80-7ce1-11eb-8064-5f7f721049aa.png)

pngcrunch should fix the CRC issues in the originals

Basically I wanted to have an extra viewport that renders 3D content over a 2D viewport so the project now has 3 viewports, all except the rear-most being transarent...

Backport (background - just a JPG for now)
Viewport (2D for Isometric tiles)
Birdport (3D - was gonna put a parrot here, changed to a cross as it's very simple to debug the object - it's 0,0,0 origin and -23 to + 23 in X, Y and Z)

You advised setting the camera to look in -1,-1,-1 in an orthogonal viewport so I set that up very simply in lines 163-188 of MainGameUnit.pas - I've hooked Resize as well but commented it out (tried all sorts of things)

Using the advised setting of 0.5, 0.5 for the origin to screen center it places the origin in the top RH side for Birdport. Experimentation shows that 0.25, 0.25 is close to center on startup

When you run it make sure you resize the window - that is really odd, Viewport (2d) remains entered while Birdport does odd things depending on the origin setting used.

You know, we could really do with some more boiler-plate SetupXXX methods - e,g, SetupIsometric (and there are many others of course). Manually working out the interaction is a time-consuming process for the uninitiated (me)

Oh - why is the camera set to (0, 0, 500) in Setup2D? It's the Z I'm wondering about, There's still a FOV in orthohonol ain't there. To my thinking the center needs to be based on the width + height of the viewport for the Z in use. I did some experimentation a few months back in this respect to allow me to chose the FOV so rotating in the Y axis minimised the total X/Y area used (3D looking straight on from 0,0)

Kagamma:
The camera was set to (0,0,500) due to an old bug, see https://sourceforge.net/p/castle-engine/discussion/general/thread/b66c948e/
This setup (Z = 500) is not important if you don't rely on TouchSensor in 2D mode.

michaliskambi:
[...]

PR #265, opened 2021-03-08
Navigation for 2D (TCastle2DNavigation) 

and3md:
This PR introduces a new type of navigation adapted to 2D games, and in particular, facilitating work in the editor on 2D games. Just press the right mouse button to easily move the content of the viewport. It's base on `TCastleExamineNavigation`. 

The way the camera works has been changed when `Camera.ProjectionType` is `ptOrthographic` (as we agreed in the interview). But maybe better solution is add property like `PixelPerfectMovement`, to not change the default behaviour of `TCastleExamineNavigation` which can also be useful or can break CGE users games.

### DIFF
--- a/src/3d/castlecameras.pas
+++ b/src/3d/castlecameras.pas
@@ -3769,8 +3769,10 @@ begin
       begin
         Size := GoodModelBox.AverageSize;
         Translation := Translation - Vector3(
-          DragMoveSpeed * Size * (Event.OldPosition[0] - Event.Position[0]) / (2*MoveDivConst),
-          DragMoveSpeed * Size * (Event.OldPosition[1] - Event.Position[1]) / (2*MoveDivConst),
+          DragMoveSpeed * Size * (Event.OldPosition[0] - Event.Position[0])
+          / (2 * MoveDivConst),
+          DragMoveSpeed * Size * (Event.OldPosition[1] - Event.Position[1])
+          / (2 * MoveDivConst),
           0);
       end;
     Result := ExclusiveEvents;
--- a/src/3d/castlecameras.pas
+++ b/src/3d/castlecameras.pas
@@ -1387,23 +1387,11 @@ type
   public
     constructor Create(AOwner: TComponent); override;
 
-    { Drag with this mouse button to move the model. }
-    property MouseButtonMove: TCastleMouseButton
-      read FMouseButtonMove write FMouseButtonMove default buttonRight;
-    { Drag with this mouse button to zoom the model (look closer / further). }
-    property MouseButtonZoom: TCastleMouseButton
-      read FMouseButtonZoom write FMouseButtonZoom default buttonMiddle;
-
-    { Optional movement algorithm, moves the camera exactly as many units as
-      the mouse position has changed especially useful in 2D in the editor. }
-    property ExactMovement: Boolean read FExactMovement write FExactMovement default true;
+    property MouseButtonMove default buttonRight;
+    property MouseButtonZoom default buttonMiddle;
+    property ExactMovement default true;
   published
-    { Enable rotating the camera around the model by user input.
-      When @false, no keys / mouse dragging / 3D mouse etc. can cause a rotation.
-
-      Note that this doesn't prevent from rotating by code, e.g. by setting
-      @link(Rotations) property or calling @link(SetView). }
-    property RotationEnabled: Boolean read FRotationEnabled write FRotationEnabled default false;
+    property RotationEnabled default false;
   end;
 
 
--- a/src/3d/castlecameras.pas
+++ b/src/3d/castlecameras.pas
@@ -1378,6 +1378,26 @@ type
       read FRotationAccelerate write SetRotationAccelerate default true;
   end;
 
+  TCastle2DNavigation = class (TCastleExamineNavigation)
+  public
+    constructor Create(AOwner: TComponent); override;
+
+    { Drag with this mouse button to move the model. }
+    property MouseButtonMove: TCastleMouseButton
+      read FMouseButtonMove write FMouseButtonMove default buttonRight;
+    { Drag with this mouse button to zoom the model (look closer / further). }
+    property MouseButtonZoom: TCastleMouseButton
+      read FMouseButtonZoom write FMouseButtonZoom default buttonMiddle;
+  published
+    { Enable rotating the camera around the model by user input.
+      When @false, no keys / mouse dragging / 3D mouse etc. can cause a rotation.
+
+      Note that this doesn't prevent from rotating by code, e.g. by setting
+      @link(Rotations) property or calling @link(SetView). }
+    property RotationEnabled: Boolean read FRotationEnabled write FRotationEnabled default false;
+  end;
+
+
   TCastleWalkNavigation = class;
 
   { What mouse dragging does in TCastleWalkNavigation. }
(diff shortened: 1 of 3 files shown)


and3md:
Added `ExactMovement` to switch to old behaviour if needed. Ready to test and merge :)

src/3d/castlecameras.pas
```diff
@@ -1125,6 +1125,7 @@       TExamineVectors = record
       FRotationEnabled: Boolean;
       FZoomEnabled: Boolean;
       FDragMoveSpeed, FKeysMoveSpeed: Single;
+      FExactMovement: Boolean;
```
michaliskambi, line 1128: I think you want to move this private variable to `TCastle2DNavigation` now?

src/3d/castlecameras.pas
```diff
@@ -1249,6 +1250,10 @@       TExamineVectors = record
     { How fast user moves the scene by pressing keys. }
     property KeysMoveSpeed: Single read FKeysMoveSpeed write FKeysMoveSpeed default 1.0;
 
+    { Optional movement algorithm, moves the camera exactly as many units as
+      the mouse position has changed especially useful in 2D in the editor. }
+    property ExactMovement: Boolean read FExactMovement write FExactMovement default false;
```
michaliskambi, line 1255: I think you want to remove this declaration. `ExactMovement` is now declared inside `TCastle2DNavigation`.

I guess these changes remained because you initially made it as a modification of `TCastleExamineNavigation` and only later reworked to be a new class :)

src/3d/castlecameras.pas
```diff
@@ -1378,6 +1383,30 @@       TExamineVectors = record
...
+    { Optional movement algorithm, moves the camera exactly as many units as
+      the mouse position has changed especially useful in 2D in the editor. }
+    property ExactMovement: Boolean read FExactMovement write FExactMovement default true;
+  published
+    { Enable rotating the camera around the model by user input.
+      When @false, no keys / mouse dragging / 3D mouse etc. can cause a rotation.
+
+      Note that this doesn't prevent from rotating by code, e.g. by setting
+      @link(Rotations) property or calling @link(SetView). }
+    property RotationEnabled: Boolean read FRotationEnabled write FRotationEnabled default false;
```
michaliskambi, line 1406: This declares a new property `RotationEnabled` that "obscures" the property `RotationEnabled` at ancestor.

To only change the default value of ancestor property, use

```
property RotationEnabled default false;
```

(without type specified).

Same goes for `MouseButtonMove`, `MouseButtonZoom`.

michaliskambi (changes requested):
All cool, I only found some minor issues :)

peardox:
I take it this will alter the navigation in my 2D app

Is there going to be a way to pick which nav is used?

#261 uses 3D nav a bit in a 2D viewport

michaliskambi:
@peardox There will either be

1. no compatibility break, if we decide to make `ExactMovement` = false the default.

2. or we decide that `ExactMovement` = true is a better default, which will break compatibility, and it will be documented on https://github.com/castle-engine/castle-engine/wiki/Upgrading-to-Castle-Game-Engine-7.0 and will just have to toggle it back to `false` to have exactly the same behavior as previously.

I'll experiment a bit before merging to decide it. It may be that AD 2 is just much better (so `ExactMovement` = true will be just better, also for view3dscene, maybe also for your game).

michaliskambi:
@peardox To make it clear, the new class `TCastle2DNavigation` will *not* be automatically used, in any projects. It is your choice what navigation you use (you do this by choosing proper navigation class in editor, or in code, in any case -- it will not automatically change to `TCastle2DNavigation`).

So the only compatibility break may happen if we decide that `ExactMovement` is `true` by default, then `TCastleExamineNavigation` movement speed (in case of orthographic projection) will change. But that is the only thing that wlil change.

src/3d/castlecameras.pas
michaliskambi: That was invalid comment from me, closing :) All is OK, TCastle2DNavigation just changes default.

src/3d/castlecameras.pas
michaliskambi: Same for `ExactMovement`. My initial comments around it were invalid. TCastleExamineNavigation implements the `ExactMovement`, and TCastle2DNavigation just makes it by default true in your PR, this it will have 100% backward compat.

and3md: Fixed, also removed comments as you suggested.

michaliskambi (approved):

michaliskambi:
ExactMovement didn't make exact movement for me on the "2d game" template. I saw it applies

- mouse position change (Event.OldPosition - Event.Position)
- Camera.Orthographic.Scale

But that is not everything yet :) Mouse position change must be scaled by UI scaling. And that is assuming Camera.Orthographic.Width = Camera.Orthographic.Height = 0, otherwise projection doesn't match control size.

I reimplemented it using `InternalViewport.PositionTo2DWorld`, PositionTo2DWorld was invented to hide these details :) Now it works nicely with "2d game" template always. Please test it also works with platformer demo.

michaliskambi:
I made some changes around this:

- I changed the mouse to use left mouse button (not right), as "the simplest -> most discoverable".

    It does raise a concern, but this concerns will have to be solved for other navigation methods too anyway: It is now easy to move camera, when you instead wanted to move some transform. (But this is now the same problem as in 3D with TCastleExamineNavigation, where one can accindentally rotate the camera.)

- ExactMovement is now more secured, and also by default `true`, even in ancestor TCastleExamineNavigation -- as it is really much better. Test e.g. view3dscene on `demo-models/navigation/ortho_viewpoint.x3dv`.

- view3dscene now has a button "2D". Just as in editor, it makes sense to view 2D models with a dedicated navigation for 2D.

and3md:
> I made some changes around this:
> 
>     * I changed the mouse to use left mouse button (not right), as "the simplest -> most discoverable".
>       It does raise a concern, but this concerns will have to be solved for other navigation methods too anyway: It is now easy to move camera, when you instead wanted to move some transform. (But this is now the same problem as in 3D with TCastleExamineNavigation, where one can accindentally rotate the camera.)
> 
The advantage is the ability to add a menu under the right button. But to the minuses, apart from those you indicated, I would add **change the selected scene when you want move camera**. So now when I have some scene selected and I need change camera position after changing cam position I need select scene I'm working on again. This is a little distracting. And I start have more clicks to do the same thing (after click to change camera position need extra click to select scene that was selected before). 

So was something like this (right click): 
- move mouse cursor away from my scene (bug in gizmos, see bellow)
- change camera position, 
- drag my scene to new position, 
- move mouse cursor away from my scene (bug in gizmos, see bellow)
- change camera position, 
- drag my scene to new position

And with left click to move camera (left click):
- move mouse cursor away from my scene (bug in gizmos, see bellow)
- change camera position, 
- click to select my scene
- drag my scene to new position, 
- move mouse cursor away from my scene (bug in gizmos, see bellow)
- change camera position, 
- click to select my scene
- drag my scene to new position

Currently right click on gizmo arrow start changing selected scene position, what is a bug in gizmos for me. When we fix that (only left click do something with gizmos). The difference between right and left click to change camera position will be even bigger:

Right click to move camera:
- change camera position, 
- drag my scene to new position, 
- change camera position, 
- drag my scene to new position

Left click to move camera:
- move mouse cursor away from my scene
- change camera position, 
- click to select my scene
- drag my scene to new position, 
- move mouse cursor away from my scene
- change camera position, 
- click to select my scene
- drag my scene to new position

So we will have twice more steps to do the same job. I think it's better to test that with https://github.com/castle-engine/castle-engine/pull/279.

michaliskambi:
The problem above boils down to the fact "left-clicking in viewport changes the selection". This is a general problem of using left mouse button for anything in TCastleXxxNavigation -- it is also a problem for rotating in TCastleExamineNavigation.

So I agree, but it needs a general fix :) And it's a fix only for editor. For view3dscene, or for "normal" applications, it's not a problem, and using left mouse button may be most natural.

PR #267, opened 2021-03-08
Implemented GravityStrength

and3md:
Because we set one kraft unit to one pixel (in the case of 2D). The default force of gravity `9.81` is very weak. This PR adds the ability to increase it in the physics settings (TPhysicsProperties.GravityStrength).


michaliskambi (approved):

michaliskambi:
Thank you!

PR #270, opened 2021-04-02
Material properties refactor

eugeneloza:
Rework syntax of `material_properties.xml` and implement better packaging conditions flexibility.

@michaliskambi, please pay special attention to `function LargestScaleValue` - the logic behind this value seems to contradict to the new logic of handling scales but `TMaterialProperties.TAutoGeneratedTextures.LoadImageEvent` seems to be inherently expecting uninterrupted sequence of scales, which is no longer the case. I.e. what should happen if we request Scale = 2 when generated only Scale = 1 and Scale = 3? Maybe we should rework that `Min` into "chose the smallest scale less or equal than `TextureLoadingScale`"?


michaliskambi (approved):
Thank you!

I did some minor code-style changes, but no logical changes. It all looks good.

I also removed "// TODO : more specific exception types". For now raising `Exception` is OK -- we do this in other CGE places too. When need arises, we can introduce more specific exception types, but in practice I see no need for them.

michaliskambi:
I did some tests and updated docs on https://castle-engine.io/creating_data_auto_generated_textures.php .

@eugeneloza Please merge now relevant "Unholy" branch to trunk, as this feature will now work in CGE master, and one `material_properties.xml` of Unholy should be OK for all platforms.

PR #272, opened 2021-04-06
PhysicsRayCast implementation

and3md:
This PR add `PhysicsRayCast()` to `TRigidBody`. It works using Kraft engine. From CGE `RayCast()` it differs by:
* you need set `MaxDistance` (in parent coordinate system)
* only kraft colliders are used - this ray can hit only physics object
* `Pickable` is not used - so you don't need make `Pickable` to true only on object you want to be raycastable in your game (what makes them also unselectable in editor). I think `Pickable` will be replaced with collision masks.

This PR is first step to make more things in CGE with real physics engine. Next we can start experimenting with collision masks and replacement of simplified CGE physics into a full physics engine.

This functionality is needed/can be tested by my platformer example https://github.com/and3md/cge-platformer-demo but you need my [platformer_fixes CGE branch](https://github.com/and3md/castle-engine/tree/platformer_fixes) to compile.

### DIFF
--- a/src/3d/castletransform_physics.inc
+++ b/src/3d/castletransform_physics.inc
@@ -245,6 +245,11 @@
     { Transformations that we collide with currently. }
     function GetCollidingTransforms: TCastleTransformList;
 
+    function PhysicsRayCast(const RayOrigin, RayDirection: TVector3;
+      const MaxDistance: Single): TCastleTransform;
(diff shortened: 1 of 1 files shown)


michaliskambi (approved):
Great, thank you! I had small comments and went ahead and just did them :)

PR #274, opened 2021-04-07
Load TMX files (Tiled Maps) by TStream/BaseUrl pair

Free-Pascal-meets-SDL-Website:
- this is needed/preparation for loading x3d nodes (see PR #241)
- this prevents redundant loading code for TMX files (here and for x3d nodes)

### DIFF
--- a/src/ui/opengl/castletiledmap_map.inc
+++ b/src/ui/opengl/castletiledmap_map.inc
@@ -315,8 +315,11 @@ type
     FTilesets: TTilesetList;
     FProperties: TPropertyList;
     FLayers: TLayerList;
+    { Load TMX file from stream. }
     procedure LoadTMXFile(const Stream: TStream; const BaseUrl_: String); overload;
+    { Load TMX file by URL. }
     procedure LoadTMXFile(const AURL: String); overload;
+    { Load TMX file internally from XMLDocument instance. }
     procedure LoadTMXFileInternal(var Doc: TXMLDocument);
   public
     property Layers: TLayerList read FLayers;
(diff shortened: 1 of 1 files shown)


michaliskambi (approved):

michaliskambi:
Sorry for the delay in reviewing! Just getting on my feet after COVID.

I noticed some small code-style things I wanted to improve, and I thought it will be faster to actually do them and then show you what I mean by showing my actual changes :) So please look at the commits associated with this PR.

Things I did, and why:

1. You created a 2nd constructor, `TTiledMap.Create(const Stream: TStream; const BaseUrl_: String)`, that had some implementation code copy-pasted from the original constructor `TTiledMap.Create(const AURL: string)`. E.g. they both did `FTilesets := TTilesetList.Create`.

    This is error-prone, it means that some future changes to `TTiledMap` would need to update both constructors, in the same way. It is easy to make an error then, and forget to update one or the other constructor.
    
    It is better, when having such overloaded constructors, to have one constructor call another. That is what I did now.

2. Minor: We generally don't use underscores in CGE to indicate internal things. (There are exceptions, but mostly because I was not strict enough in past reviews.) So `BaseUrl_` -> `ABaseUrl`.

3. Minor: Avoid the class name comment like `{ TTiledMap }` added by Lazarus Code Tools by default. You can turn it off in Lazarus Code Tools settings. In CGE we want to replace it with a good comment, that actually documents the class purpose, in this case we already had `{ Loading and manipulating "Tiled" map files (http://mapeditor.org). }`.

4. In general try to "free things" as close as possible to the code that "creates these things". This is even more important than "avoid code duplication" sometimes in my mind. So I removed the `FreeAndNil(Doc)` from `LoadTMXFileInternal`, and instead I call `FreeAndNil(Doc)` 2 times, from 2 overloaded versions of `TTiledMap.LoadTMXFile`. This way `Doc` is freed each time very close to the line of code that creates `Doc`.

Free-Pascal-meets-SDL-Website:
I'm glad you recovered from COVID!

> I noticed some small code-style things I wanted to improve, and I thought it will be faster to actually do them and then show you what I mean by showing my actual changes :) So please look at the commits associated with this PR.

I appreciate this approach and all the detailed comments I can learn from. Thanks for that!

PR #279, opened 2021-04-13
Editor 2D mode: support translation in X and Y axis at the same time

and3md:
Having to catch the right axis to move an object in 2d significantly reduces productivity and is a bit annoying. I added an option that in the case of 2D, catching the Z axis allows you to move object in two dimensions at the same time (X,Y). Implemented only for translation. 

xy

### DIFF
--- a/tools/castle-editor/code/visualizetransform.pas
+++ b/tools/castle-editor/code/visualizetransform.pas
@@ -44,7 +44,8 @@ type
         InsideInternalCameraChanged: Boolean;
 
         { Point on axis closest to given pick.
-          Axis may be -1 to indicate we drag on all axes with the same amount. }
+          Axis may be -1 to indicate we drag on all axes with the same amount
+          or -2 to indicate we drag X and Y axes for 2D. }
         function PointOnAxis(out Intersection: TVector3;
           const Pick: TRayCollisionNode; const Axis: Integer): Boolean;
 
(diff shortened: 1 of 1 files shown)


and3md:
Jenkins fails with `Disk Full` error: 

```
Free Pascal Compiler version 3.0.2 [2017/02/05] for x86_64
Copyright (c) 1993-2016 by Florian Klaempfl and others
Fatal: I/O error: Disk Full
Error: /usr/local/fpclazarus/3.0.2/fpc/bin/ppcx64 returned an error exitcode
Exception "Exception":
Failed to compile
```

michaliskambi (changes requested):
The implementation looks nice and simple.

But as we talked, I'd like to make a functional change. Instead of detecting a click on "Z" axis as a special case for 2D transformations, instead add a dedicated transparent box inside the gizmo that user can drag.

- Such transparent box should be added by Blender to the gizmo 3D model (in .blend and then exported to glTF). Just like an equivalent transparent box in case of scaling gizmo (used to scale in XYZ easily).

- Then it can be used to make a special translation in XY in case of "standard 2D projection". It will be easier to drag for user than "just Z axis".

- And it opens the door for future other translation mode, in general 3D. Maybe the 3D object can be translated in the plane parallel to the camera when dragged by this transparent box? Maybe it should even "stick" to walls / ceilings?

and3md:
I added cube in center point, but it's a little hard to decide how big should it be (in 2D it can be bigger but in 3D maybe it's too big). I scaled it to 0.19 in translation.blend and to 0.20 in translation_collider. Now dragging works great in 2d template/platformer and it's a lot convenient :)

PR #283, opened 2021-04-19
Not existing TRigidBody should not be ksfRayCastable (found by ray cast)

and3md:
In Kraft rigid body without `ksfCollision` still can be found by ray cast. But we don't want that when `TRigidBody.Exist = false`. This PR fixes that.


michaliskambi (approved):

michaliskambi:
Thank you!!

PR #288, opened 2021-05-04
More work/fixes in Platformer demo

and3md:
Fixes:
* fix using Push() with game over state
* fix spikes never ends hurting player
* once the spike hits the ground, it should no longer be able to hurt the player
* dead player shouldn't do anything
* set big friction on movable platforms only (TODO: player moves slower on them)

### DIFF
--- a/examples/platformer/code/gamestateplay.pas
+++ b/examples/platformer/code/gamestateplay.pas
@@ -225,10 +225,6 @@ begin
   RBody := TRigidBody.Create(Platform);
   RBody.Dynamic := false;
 
-  { Movable platforms are animated }
-  {if Platform.Tag <> 0 then
-    RBody.Animated := true;}
-
   if Platform.Tag <> 0 then
     RBody.Dynamic := true;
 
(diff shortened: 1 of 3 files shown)


examples/platformer/code/gamestateplay.pas
```diff
@@ -1375,9 +1387,10 @@ procedure TStatePlay.Update(const SecondsPassed: Single; var HandleInput: Boolea
   inherited;
   { This virtual method is executed every frame.}
 
-  if GameOver then
+  { If player is dead and we did not show game over state we do that }
```
michaliskambi, line 1390: I do not see the point of using `Push` here. It makes code harder to reason about: underneath `StateGameOver` you will have, still playing, `StatePlay`, and you need to account for it.

Let's keep the demo simple, and just have `TUIState.Current := StateGameOver`.

And remove the extra modifications (if some above checks ` if IsPlayerDead then` were required because of this only).

michaliskambi (changes requested):

examples/platformer/code/gamestateplay.pas
michaliskambi: @and3md convinced me -- we'll actually make `StateGameOver` transparent.

PR #292, opened 2021-05-10
Change rigid bodies existence when TCastleTransform.Exists changes.

and3md:
Before this PR, when you set `TCastleTransform.Exists` to `False` assigned `TRigidBody` still exists. So this makes very funny bugs;)

Before this PR you need set two `Exists` to `False`:
```
CastleTransform.Exists := False;
CastleTransform.RigidBody.Exists := False;
```
to make something really not existing. And you need make loop on all children to make them not existing too.

This PR add existence tracking mechanism to `TCastleTransform` and use it only to update Kraft flags.  So it's also disabled when `FWorld.FKraftEngine = nil`. Thanks to change Kraft flags you can remember `RigidBody.Exists` value.

Currently `TCastleAbstractRootTransform` existence is not tracked. So when you set `TCastleAbstractRootTransform.Exists` to `False` children will not be updated. This can be done (with small code complications) but I don't know is it worth to do it (or we want to support that case?

### DIFF
--- a/src/3d/castletransform.pas
+++ b/src/3d/castletransform.pas
@@ -341,10 +341,10 @@ type
     FLastUpdatedGetExists: Boolean;
     FLastUpdatedGetExistsValid: Boolean;
 
-    { When castle transform existance changes to false, we need update existance
+    { When castle transform existence changes to false, we need update existence
       in all castle transform children rigid bodies }
     procedure UpdateExistRecursively(const AExists: Boolean);
-    { Update rigid body existance when castle transform (or it parent) GetExists
+    { Update rigid body existence when castle transform (or it parent) GetExists
       changes. }
     procedure UpdateExist(const AExists: Boolean);
   private
--- a/src/3d/castletransform.pas
+++ b/src/3d/castletransform.pas
@@ -343,10 +343,15 @@ type
 
     { When castle transform existence changes to false, we need update existence
       in all castle transform children rigid bodies }
-    procedure UpdateExistRecursively(const AExists: Boolean);
+    procedure UpdateExistsRecursively(const AExists: Boolean);
     { Update rigid body existence when castle transform (or it parent) GetExists
       changes. }
-    procedure UpdateExist(const AExists: Boolean);
+    procedure UpdateExists(const AExists: Boolean);
+
+    { Check current item exist taking into account the existence of the
+      unique parents of this item, needed only before first Update() call.
+      See GetRealExistance(). }
+    function GetExistsRecursively: Boolean;
   private
     type
       TEnumerator = class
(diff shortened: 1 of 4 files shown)


src/3d/castletransform_physics.inc
```diff
@@ -720,6 +721,7 @@ procedure TRigidBody.InitializeTransform(const Transform: TCastleTransform);
     FKraftBody.Finish;
 
     UpdateCollides(Transform);
+    UpdateExist(Transform.GetExistsRecursively);
```
michaliskambi, line 724: Ideally `TRigidBody` should not call a recursive function here, but only access a ready field `Transform.ExistsRecursively`. Disclaimer: I did not analyze if this is feasible to implement :)

michaliskambi (changes requested):
I made some comments, mainly I'm interested is it possible to have a `TCastleTransform` private field `FExistsRecursively` updated, instead of method `GetExistsRecursively`. 

As for `TCastleAbstractRootTransform.Exists` -- this is a case we want to handle indeed, too.

Also (most important) we need automatic tests to make sure it works OK. For this I say to add public to `TRigidBody`

```
{ @exclude Only for testing. }
property InternalKraftBody: TKraftRigidBody read FKraftBody;
```

and looks in auto-tests whether InternalKraftBody.Flags has the desired flags aftet some operations. At the minimum, we need a test that would fail before this PR, and works afterwards :) 

Same for exposing TKraftShape inside a `TCollider`,

```
{ @exclude Only for testing. }
property InternalKraftShape: TKraftRigidShape read FKraftShape;
```

src/3d/castletransform.pas
```diff
@@ -447,6 +452,10 @@       TEnumerator = class
     procedure RemoveFromWorld(const Value: TCastleAbstractRootTransform);
     procedure RemoveBehaviorIndex(const BehaviorIndex: Integer);
     procedure SetListenPressRelease(const Value: Boolean);
+
+    { Check current item exist taking into account the existence of the
+      unique parents of this item }
+    function GetRealExistance: Boolean;
```
michaliskambi, line 458: Proper English is "existence" not "existance" as far as I see.

Although "existance" seems to be a common misspelling, some dictionaries redirect it to "existence" automatically ( https://duckduckgo.com/?t=ffab&q=Existance&ia=web and note that it's not DuckDuckGo fixing the misspelling, but various particular pages ).

michaliskambi:
We figured out a way forward with this in https://trello.com/c/4gja9P0d/157-implement-existsinroot-in-a-simple-way :)

Hopefully the `examples/platformer/` can have a simple workaround in the meantime.

michaliskambi:
I have implemented `ExistsInRoot`, mostly following our plans on https://trello.com/c/4gja9P0d/157-implement-existsinroot-in-a-simple-way . It's now on branch `light-components`. 

I didn't do (so far) much actions to make it useful for physics, as @and3md is working on new rigid body/collider classes anyway :) I merely adjusted in `TRigidBody.UpdateCollides` check

```pascal
  if Transform.Exists and Transform.Collides then
```

->

```pascal
  if Transform.ExistsInRoot and Transform.Collides then
```

(following TODO that was already there). @and3md lets decide what remains from this PR (maybe we can take auto-tests), probably after both my branch `light-components` and your branch with new physics behaviors are merged to master.

michaliskambi:
Closing this -- this is effectively done by `ExistsInRoot` and new physics (already in `master`) using it,

PR #299, opened 2021-05-25
Fix TFontFamily deprecation message

eugeneloza:


michaliskambi:
Thanks for noticing it!

PR #302, opened 2021-06-12
Package debian

eugeneloza:
Implement Deb package creation using Castle Game Engine by --package-format=deb

* Build binaries
* Create appropriate folder structure for DEB
* Convert FPC architecture names to Debian architecture names
* Fill-in basic required metadata for DEB package
* Create menu items and run scripts
* Add <debian> node to CastleManifest with 4 items: `install_folder`, `section`, `categories` and `comment` to fill in required metadata (see https://gitlab.com/EugeneLoza/kryftolike/-/commit/4f79ebd291332361604aa27708858dab46223e60)

And two minor fixes:

* Ability to save XPM (not a good one though - we are lacking transparency and the resulting file size is not optimal)
* Fix a bug in TImageFileNames.FindExtension that always searched only for '.ico' files.


tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+  // Post-installation running - assign executable permissions
+
+  TextWriter := TTextWriter.Create(PackageFolder + PathDelim + 'DEBIAN' + PathDelim + 'postinst');
+  TextWriter.Write(
+    '#!/bin/sh' + NL + NL +
+    'chmod +x ' + PathToExecutableUnix + '/' + Manifest.ExecutableName + NL + NL +
+    'exit 0'
+    );
+  FreeAndNil(TextWriter);
+  fpChmod(PackageFolder + PathDelim + 'DEBIAN' + PathDelim + 'postinst', &555);
```
michaliskambi, line 168: Jenkins reports a compilation error here, and is right ( https://jenkins.castle-engine.io/job/castle_game_engine_organization/job/castle-engine/job/PR-302/5/console  ):

```
... for Win64:
...
Free Pascal Compiler version 3.2.0 [2020/06/14] for x86_64
Copyright (c) 1993-2020 by Florian Klaempfl and others
tooldebian.pas(168,3) Error: Identifier not found "fpChmod"
```

Indeed `FpChmod` should be under `{$ifdef UNIX}`.

Actually, make it even better: move the `DoMakeExecutable` from `toolpackage.pas` into `toolutils.pas`. 

Then
- `TPackageDirectory.Add` can call `DoMakeExecutable(Path + Name)`
- and your new code can call `DoMakeExecutable(PackageFolder + PathDelim + 'DEBIAN' + PathDelim + 'postinst')`
- and the `$ifdef` is hidden inside `DoMakeExecutable`.

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+    'Installed-Size: ' + IntToStr(BinariesSize div 1024) + NL +
+    'Depends: libopenal1, libpng16-16, zlib1g, libvorbis0a, libvorbisfile3, libfreetype6, libgl1-mesa-dri, libgtk2.0-0' + NL +
+    'Description: ' + Manifest.Caption + NL +
+    ' ' + Manifest.DebianComment + NL //final new line
+    );
+  FreeAndNil(TextWriter);
+
+  // Post-installation running - assign executable permissions
+
+  TextWriter := TTextWriter.Create(PackageFolder + PathDelim + 'DEBIAN' + PathDelim + 'postinst');
```
michaliskambi, line 161: To simply create a file from a string contents, use `StringToFile`. It is simpler to use than creating / destroying `TTextWriter` instance.

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+      m68k: Result := 'm68k';
+      riscv64: Result := 'riscv64';
+      sparc64: Result := 'sparc64';
+      // "alpha"
+      // "sh4"
+      // "x32" = X32 is an ABI for amd64/x86_64 CPUs using 32-bit integers, longs and pointers
+      else {powerpc,sparc,avr,jvm,i8086,riscv32,armeb}
+      begin
+        Result := CpuToString(Cpu);
+        WriteLn('WARNING: Architecture ' + CpuToString(Cpu) + ' is not officially supported by Debian.');
```
michaliskambi, line 68: Use `WriteLnWarning` to report warnings in the build tool.

CGE build tool main program does `ApplicationProperties.OnWarning.Add(@ApplicationProperties.WriteWarningOnConsole)` to make sure that warnings are displayed on console too.

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+  CreateDir(ShareDir + PathDelim + 'pixmaps');
+  PathToIconFileUnix := '/usr/share/pixmaps/' + Manifest.ExecutableName + '.xpm';
+  PathToIconFileLocal := StringReplace(PathToIconFileUnix, '/', PathDelim, [rfReplaceAll]);
+  if Manifest.Icons.FindExtension(['.xpm']) <> '' then
+    CheckCopyFile(Manifest.Icons.FindExtension(['.xpm']), PackageFolder + PathToIconFileLocal)
+  else
+  begin
+    // using ImageMagic - FPWriteXPM first doesn't properly write alpha channel, second uses palette char size = 2 which is not a good idea for an icon
+    //RunCommandSimple(FindExe('convert'), [Manifest.Icons.FindExtension(['.png']), PackageFolder + PathToIconFileLocal]);
+    if not RunCommand('/bin/convert', [Manifest.Icons.FindExtension(['.png']), PackageFolder + PathToIconFileLocal], OutString) then
```
michaliskambi, line 109: Better to not specify absolute path `/bin/convert`.

Reason: Is it `/bin/convert` or `/usr/bin/convert`? On my Debian I have both, because on latest Debian `/bin` is just a symlink to `/usr/bin`. But that is not true on all Unix systems, and not on older Debian versions. `/usr/bin` is actually a traditional place for non-critical programs, like ImageMagick I guess. Better to search on `$PATH`.

To do this, use `FindExe`:

```
ConvertExe := FindExe('convert');
if ConvertExe = '' then 
  raise Exception.Create('Cannot find "convert", install ImageMagick or GraphicsMagick');
if not RunCommand(ConvertExe, ...
```

Bonus gain is that it makes the code more portable. In the (distant probably...) possibility that e.g. Cygwin will one day package Debian tools, maybe this can work on Windows? Above `FindExe` approach would work OK.

michaliskambi: Moreover, in this case, as you don't really want to catch `OutString`, just use `RunCommandSimple`.

And it will automatically search `$PATH`. So just

```
RunCommandSimple('convert', [Manifest.Icons.FindExtension(['.png']), ...
```

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+
+var
+  PathToExecutableLocal, PathToExecutableUnix: String;
+  PathToIconFileLocal, PathToIconFileUnix: String;
+  ShareDir: String;
+  TextWriter: TTextWriter;
+  OutString: String;
+  PackageFolder: String;
+begin
+  PackageFolder := 'castle-engine-output' + PathDelim + PackageFileName;
```
michaliskambi, line 81: Do not depend that `castle-engine-output` is a subdirectory of current working directory, do not even depend that it is a subdirectory of project directory. CGE build tool may be called from other directories (like any project subdirectory) and user can use `--output` to write the `castle-engine-output` anywhere.

So use `TempOutputPath(Project.Path)` above, like 

```
PackageFolder := TempOutputPath(Project.Path) + PackageFileName;
```

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+  if DirectoryExists(PackageFolder) then
+    RemoveNonEmptyDir(PackageFolder);
+
+  // Copy the binaries
+
+  PathToExecutableUnix := '/usr/' + Manifest.DebianInstallFolder + '/' + Manifest.Name;
+  PathToExecutableLocal := StringReplace(PathToExecutableUnix, '/', PathDelim, [rfReplaceAll]);
+  CopyDirectory(Path, PackageFolder + PathToExecutableLocal);
+  ShareDir := PackageFolder + PathDelim + 'usr' + PathDelim + 'share';
+  CreateDir(ShareDir);
```
michaliskambi, line 91: Check `CreateDir` result -- it does not raise exception in case of problems, only returns `false`.

tools/build-tool/code/tooldebian.pas
michaliskambi: Moreover I generally don't use term `Folder` in CGE code. Call it `PackageDir`. In CGE code I generally call things `XxxDir` (if they don't end with path delim) or `XxxPath` (if they do end with path delim).

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
...
+  TextWriter.Write(
+    'cd castle-engine-output' + NL +
+    'cd ' + PackageFileName + NL +
+    'find -type f | egrep -v ''^\./DEBIAN'' | xargs --replace=hh -n1 md5sum "hh" | sed ''s/\ \.\///'' > DEBIAN/md5sums' + NL +
+    'cd ..' + NL +
+    'dpkg-deb --build ' + PackageFileName
+    );
+  FreeAndNil(TextWriter);
+
+  RunCommand('/bin/bash', ['castle-engine-output/package-debian.sh'], OutString);
```
michaliskambi, line 182: It is rather unclean to create a shell script, only to execute it :) Moreover this script has a problem: it doesn't use `set -e` so it will work OK even if some command in the middle will fail (see http://redsymbol.net/articles/unofficial-bash-strict-mode/ for nice discussion on how to make bash scripts more reliable).

Better to use a series of `RunCommandSimple` calls (or other routines from `ToolCommonUtils` or FPC `Process`). If you need pipes for 1 line, run this line with `bash -c`.

Like this:

```
RunCommandSimple(PackageDir, 'bash', ['-c', 
  'find -type f | egrep -v ''^\./DEBIAN'' | xargs --replace=hh -n1 md5sum "hh" | sed ''s/\ \.\///'' > DEBIAN/md5sums']);

// you may need to fix quoting above. Since you don't create shell script, you have one less thing to escape.

RunCommandSimple(TempOutputPath(Project.Path), 'dpkg-deb', ['--build', PackageFileName]);
```

tools/common-code/toolmanifest.pas
```diff
@@ -233,6 +238,23 @@   TCastleManifest = class
...
+      We make sure this directory will be set as a working directory when launched
+      Default: games }
+    property DebianInstallFolder: String read FDebianInstallFolder;
+    { App menu section, see man menufile for more information,
+      Default: Games }
+    property DebianSection: String read FDebianSection;
+    { Freedesktop category of the app,
+      see https://www.freedesktop.org/wiki/Specifications/menu-spec/ for more info
+      Default: Game }
+    property DebianCategories: String read FDebianCategories;
```
michaliskambi, line 253: Calling it `DebianCategories` limits the usage. Call it more generic `FreeDesktopCategories`. This way it will be useful for all cases when we need to make `xxx.desktop` file for the game.

tools/common-code/toolmanifest.pas
```diff
@@ -233,6 +238,23 @@   TCastleManifest = class
...
+    { A short description of the project.
+      See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html for more info }
+    property DebianComment: String read FDebianComment;
```
michaliskambi, line 256: Same as above: call it `FreeDesktopComment`.

I know, it is also used as Debian package description later. That is OK -- we can always add new field specifically for multiline Debian description in the future. In the meantime, it makes sense we reuse desktop file short comment for it.

tools/build-tool/code/tooldebian.pas
```diff
@@ -0,0 +1,192 @@
+unit ToolDebian;
```
michaliskambi, line 1: Add our CGE copyright header, like

```
{
  Copyright 2021-2021 Eugene Loza.

  This file is part of "Castle Game Engine".

  "Castle Game Engine" is free software; see the file COPYING.txt,
  included in this distribution, for details about the copyright.

  "Castle Game Engine" is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

  ----------------------------------------------------------------------------
}
```

michaliskambi (changes requested):
Thank you! I did an initial review, and suggested some changes :)

tools/build-tool/code/tooldebian.pas
eugeneloza: Many thanks for all the suggestions! Should be easy to fix. For now just a note here:

```
Manifest.Icons.FindExtension(['.png'])
```

Literally finds only "png" icon file. I wasn't sure how to proceed here. We need some "main" icon file, but most certainly not "ico" which produces broken results when converted (all icon sizes dumped into one image). But this icon may be any CGE-readable.

I thought maybe it's even a good idea to write our own XPM exporter for this purpose. It's really straightforward and in case of an app icon can be simplified even further. It won't be universal, as will work reliably only for small images, but specifically for this purpose.

tools/build-tool/code/tooldebian.pas
eugeneloza: Done in 888690fdac181b6f140a71fd0d5ac20a7d7b59e6

tools/build-tool/code/tooldebian.pas
eugeneloza: Fixed in 492b70590837280d3642fd17b669ed4cf5b84465 and fc48d68471b1e5639df5260b441a4e25d0ceafd7

tools/build-tool/code/tooldebian.pas
eugeneloza: Done in 89fb8d670ddeef0fa8863b33811140d9084d2f8c

tools/build-tool/code/tooldebian.pas
eugeneloza: Fixed in 89fb8d670ddeef0fa8863b33811140d9084d2f8c

tools/build-tool/code/tooldebian.pas
eugeneloza: Done in b64cfed4e5a8b51f9546fc04b14bb9c81421306f

tools/build-tool/code/tooldebian.pas
eugeneloza: Fixed in 0b9dcf3c2e0b3e4833211baa678487cc6bce9e5e

tools/build-tool/code/tooldebian.pas
eugeneloza: Done in cc5a5a32d5ef9cf591182ced9edc7b128479f273

tools/common-code/toolmanifest.pas
eugeneloza: Done in 2f1630dfc4257ef5d5879b52875dea04200dfc65

tools/common-code/toolmanifest.pas
eugeneloza: Done in 2f1630dfc4257ef5d5879b52875dea04200dfc65

tools/build-tool/code/tooldebian.pas
eugeneloza: Fixed + improved a bit. Still the question about png...

michaliskambi:
I am cherry-picking https://github.com/castle-engine/castle-engine/pull/302/commits/888690fdac181b6f140a71fd0d5ac20a7d7b59e6 into master, to have cleaner line for AAB building.

tools/build-tool/code/tooldebian.pas
michaliskambi: For now, I'd say it's OK to require just PNG. Indeed we need anything that is readable by ImageMagick, so in general this could be better, but for now I suppose just requiring PNG is cool.

michaliskambi (approved):

michaliskambi:
Thank you for being patient with me and updating this PR to master multiple times! :)

I tested it, made some adjustments, and I'm merging now.

Most important changes:

1. I reorganized the target files:

    - We now copy exe to `/usr/bin/<project-name>`
    - data to `/usr/share/<project-name>`

    This seems suitable for all "normal" applications (games and not games), and our ApplicationData is already ready to detect such system-wide data location since a long time.

    This also means that .desktop and menu files just call the executable, without the need for "cd ..." into application directory.

2. I removed DebianInstallFolder, as it had 2 meanings (subdirectory of /usr/, for which really only "bin" and "games" were valid, and section in Debian CONTROL file). Instead we just install to hardcoded /usr/bin, and the section for CONTROL is controlled by DebianControlSection. The section for menu is in DebianMenuSection (renamed from DebianSection).

    The manifest format is now

    ```
    <debian control_section="..." menu_section="..." />
    <free_desktop categories="..." comment="..." />
    ```

    so it follows TCastleManifest properties, and it follows separation between Debian-specific and freedesktop standards.

3. I experimented and in the end, reverted the split int TPackageDirectoryAbstract/TPackageDirectory. The main Debian packaging is now just a routine called from TPackageDirectory.Make. This makes it a bit more consistent in usage to zip/tar.gz.

PR #314, opened 2021-07-05
Android app bundle

eugeneloza:
Generating Android App Bundle for Google Play Store.

As in

* `--package-format=default` (will generate `android-apk`)
* `--package-format=android-apk`
* `--package-format=android-app-bundle`

Tested and generates a proper signature key.

Note: It **does not use resources**, the internal structure of the project remains the same as it was before, when only APK was generated. Google guidelines suggest that different resources should be packaged for different devices (e.g. different images sizes for different screen resolutions, or different strings for different device localizations)

### DIFF
--- a/tools/build-tool/code/toolandroid.pas
+++ b/tools/build-tool/code/toolandroid.pas
@@ -635,7 +635,7 @@ var
   end;
 
 var
-  ApkName: string;
+  PackageName: string;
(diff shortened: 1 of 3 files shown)


michaliskambi:
Updated https://github.com/castle-engine/castle-engine/wiki/Build-Tool about it too (at `package` description).

PR #317, opened 2021-07-07
Important: Load from more than one tileset + Add LogWarning + TODO comment

Free-Pascal-meets-SDL-Website:
Still working on the conversion class (#241), so I have no time to fix this atm. 

Best regards

Matthias

### DIFF
--- a/src/ui/opengl/castletiledmap_map.inc
+++ b/src/ui/opengl/castletiledmap_map.inc
@@ -1041,9 +1041,11 @@ begin
            above and consequently are not found in the Tiles list. }
   if not (TileCount = Tiles.Count) then
   begin
-    WritelnWarning(URIDisplay(URL) + ': Could not load all tiles of tileset!' +
-      ' Expected tiles: ' + IntToStr(TileCount) + ' Loaded tiles: ' +
-      IntToStr(Tiles.Count));
+    WritelnWarning('%s: Could not load all tiles of a tileset. Expected tiles: %d, loaded tiles: %d', [
+      URIDisplay(URL),
+      TileCount,
+      Tiles.Count
+    ]);
   end;
 
   if TileCount = 0 then
--- a/src/ui/opengl/castletiledmap_map.inc
+++ b/src/ui/opengl/castletiledmap_map.inc
@@ -1034,6 +1034,14 @@ begin
     end;
   finally FreeAndNil(I) end;
 
+  { Fallback: TileCount determination if is not found in (old) tsx-files prior
+    to version 0.13 (Aug 2015). }
+  if TileCount = 0 then
+  begin
+    Image.DetermineSize;
+    TileCount := (Image.Width div TileWidth) * (Image.Height div TileHeight);
+  end;
+
   { TODO : Handle tiles which are not explicitly found in the tsx-file.
     Hint : Tiles which have no special setting (probability equals 1.0,
            no terrains set, no type set) are not explicitly written down
@@ -1046,12 +1054,6 @@ begin
       TileCount,
       Tiles.Count
     ]);
-  end;
-
-  if TileCount = 0 then
-  begin
-    Image.DetermineSize;
-    TileCount := (Image.Width div TileWidth) * (Image.Height div TileHeight);
   end;
 end;
 
--- a/src/ui/opengl/castletiledmap_map.inc
+++ b/src/ui/opengl/castletiledmap_map.inc
@@ -947,6 +947,7 @@ begin
   Properties := TPropertyList.Create;
   Tiles := TTileList.Create;
   Image := TImage.Create;
+  FirstGID := 1;
 end;
 
 destructor TTiledMap.TTileset.Destroy;
@@ -984,9 +985,7 @@ begin
   Margin := 0;
 
   if Element.AttributeString('firstgid', TmpStr) then
-    FirstGID := StrToDWord(TmpStr)
-  else
-    FirstGID := 1;
+    FirstGID := StrToDWord(TmpStr);
   if Element.AttributeString('source', TmpStr) then
   begin
     URL := CombineURI(BaseUrl, TmpStr);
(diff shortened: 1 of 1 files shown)


src/ui/opengl/castletiledmap_map.inc
```diff
@@ -1034,6 +1034,18 @@ begin
...
   finally FreeAndNil(I) end;
 
+  { TODO : Handle tiles which are not explicitly found in the tsx-file.
+    Hint : Tiles which have no special setting (probability equals 1.0,
+           no terrains set, no type set) are not explicitly written down
+           in the tsx file. This means they are not loaded by the iteration
+           above and consequently are not found in the Tiles list. }
+  if not (TileCount = Tiles.Count) then
+  begin
+    WritelnWarning(URL + ': Could not load all tiles of tileset');
```
michaliskambi, line 1044: Show URL using `URIDisplay(URL)`.

Also make it a single warning, not 2 warnings. This is desired in case some code captures warnings, counts them, sends them somewhere (e.g. to a remote server) etc. -- so a long warning should still be 1 warning, not 2 warnings :)

And in general use format (`%d` instead of `IntToStr` concatenation) -- it looks a bit better for longer strings.

So like

```
WritelnWarning('%s: Could not load all tiles of a tileset. Expected tiles: %d, loaded tiles: %d', [
  URIDisplay(URL),
  TileCount,
  Tiles.Count
]);
```

src/ui/opengl/castletiledmap_map.inc
```diff
@@ -1034,6 +1034,18 @@ begin
     end;
   finally FreeAndNil(I) end;
 
+  { TODO : Handle tiles which are not explicitly found in the tsx-file.
+    Hint : Tiles which have no special setting (probability equals 1.0,
+           no terrains set, no type set) are not explicitly written down
+           in the tsx file. This means they are not loaded by the iteration
+           above and consequently are not found in the Tiles list. }
+  if not (TileCount = Tiles.Count) then
```
michaliskambi, line 1042: Writing `if TileCount <> Tiles.Count then` looks more straightforward to me.

src/ui/opengl/castletiledmap_map.inc
```diff
@@ -1034,6 +1034,18 @@ begin
...
+  begin
+    WritelnWarning(URL + ': Could not load all tiles of tileset');
+    WritelnWarning('Expected tiles: ' + IntToStr(TileCount) +
+      ' Loaded tiles: ' + IntToStr(Tiles.Count));
+  end;
+
   if TileCount = 0 then
```
michaliskambi, line 1049: Question: Should we move this check later, *after* the block 

```
  if TileCount = 0 then
  begin
    Image.DetermineSize;
    TileCount := (Image.Width div TileWidth) * (Image.Height div TileHeight);
  end;
```

? This block potentially adjusts `TileCount`. So the check you propose should be after it?

michaliskambi (changes requested):
Thank you! I have some small requests below.

Also, can you attach an example of such map? So that we know exactly how to close this TODO in the future.

src/ui/opengl/castletiledmap_map.inc
Free-Pascal-meets-SDL-Website: The following files can be used in the Tiled editor. Anyway, if you look at the tsx-file you will find, that only one tile is explicitly found (with id = 2). For this specific tile I changed the probability to illustrate this. All the Tiled examples used in CGE have modified tiles in the tileset, hence they can be loaded. 

test512
[Test512-missing-tiles.tsx.txt](https://github.com/castle-engine/castle-engine/files/6780748/Test512-missing-tiles.tsx.txt)
[map512-missing-tiles.tmx.txt](https://github.com/castle-engine/castle-engine/files/6780749/map512-missing-tiles.tmx.txt)

src/ui/opengl/castletiledmap_map.inc
Free-Pascal-meets-SDL-Website: Yes, you are right. I would even propose to add a Log entry to indicate that the tsx-file is outdated if it doesn't carry a tilecount entry, as this is the case with tsx files since ver. 0.13 (2015). 

Or could you imagine any other scenario in which TileCount = 0 is possible even in recent tsx-files?

Free-Pascal-meets-SDL-Website:
Thanks for the quick reply. I hope, I resolved all the issues. The example files are uploaded above at the request.

Best regards
Matthias

Free-Pascal-meets-SDL-Website:
As this PR makes just minor changes and the last commit (2b5b179) is also in the same file, I think it is ok to just add it here. I will adapt the PR title after finishing this text.

About the commit 2b5b179: Before the fix it was basically only possible to have maps from ONE tileset. Although the code in principle suggest to have several tilesets (see Map.Tilesets). Anyway, the fixed if-statement made it so, that all tilesets always got FirstGID = 1 allocated even if it was stated otherwise in the tmx file. 

You can easily test it with the example tiled/map_viewer. 
It is showing the desert.tmx and the tiles from desert.tsx perfectly fine. 
But add the attached 3 files (new tilemap, 1 additional tileset, 1 additional tileset image file) to the data/maps folder.
Change line 88 in the code/gameinilize.pas to load the new map:
`TiledMap.URL := 'castle-data:/maps/desert2tilesets.tmx';`
and you see, the loading doesn't work correctly.

The commit patches this. After it, it loads perfectly fine. The result should look like this (screenshot from Tiled).
result-small

(had to rename all files to txt-files due to upload-restrictions)
[desert2tilesets.tmx.txt](https://github.com/castle-engine/castle-engine/files/6831985/desert2tilesets.tmx.txt)
[Test512.tsx.txt](https://github.com/castle-engine/castle-engine/files/6831986/Test512.tsx.txt)
test512

Free-Pascal-meets-SDL-Website:
I'm closing this PR in favour of a new PR (soon).

PR #319, opened 2021-07-14
PVRTexToolCLI

eugeneloza:
Use new version of PVRTexToolCLI (PVRTexTool v5.1.0) in Castle Game Engine.

Tested and working. ~~There are some glitches with loading the generated textures, but they may be specific to ASTC_4x4 format, as not all textures are affected.~~ yes, investigating more - it's ASTC issue, which doesn't have relation to this PR, however, maybe we'll need to address it from our side so that users will be able to use it without hassle.

We still need to comply with:

> If object code is bundled with a product, all branding should be kept as it was originally, and the following acknowledgement should be displayed clearly in any associated documentation or other collateral in printed or electronic form distributed with the product incorporating the Software: “This product includes components of the PowerVR Tools Software from Imagination Technologies Limited”

I believe "associated documentation" here can simply mean putting this notice at https://castle-engine.io/creating_data_auto_generated_textures.php is enough.


eugeneloza:
Hmm... I wonder if there is any need in this PR anymore? PVRTexToolCLI may be useful, but it's glitchy for ASTC format. And the version we're currently using in Docker image seems enough for all its uses.

I believe we should use ASTCENC as in https://github.com/castle-engine/castle-engine/pull/321  - also FOSS compatible, included in Debian repository - it seems to work perfectly for our usecase and without any glitches of PVRTexToolCLI and NvidiaTextTools.

michaliskambi:
I agree with your last comment @eugeneloza . I mean I still want this PR, because I want this "infrastructure" -- I believe we should package some texture tools with CGE, as we want https://castle-engine.io/creating_data_auto_generated_textures.php to work more "out of the box" (whether someone uses our Docker or not).

But the `PVRTexToolCLI` may indeed be the wrong candidate for this integration. As it's not critical, and not open-source.

I say to remake this PR to include open-source texture compression tools, that we can distribute without hassle, so

1. [ASTCENC](https://developer.arm.com/tools-and-software/graphics-and-gaming/mali-texture-compression-tool/astc-encoder) (license: Apache 2.0, https://github.com/ARM-software/astc-encoder )
2. [nvcompress](https://developer.nvidia.com/gpu-accelerated-texture-compression) (license: MIT)
3. [AMD Compressonator](https://gpuopen.com/compressonator/) (license: combination of MIT, BSD-3, other -- https://github.com/GPUOpen-Tools/compressonator/tree/master/license )

From what I see, they are all open-source and thus we could "bundle" them without getting entangled in any additional legalese. The licenses listed above (MIT, BSD-3, Apache 2.0) are very permissive.

I would say to go with [nvcompress](https://developer.nvidia.com/gpu-accelerated-texture-compression) for the start. It creates S3TC compressed textures, thus is most useful for desktop apps. This will allow to see the benefits of the system "tools are bundled in CGE" quickly.

michaliskambi (changes requested):
See my above comments. In short, I'd say we should bundle with "nvcompress" for start.

eugeneloza:
I've implemented the changes from this PR into https://github.com/castle-engine/castle-engine/pull/351 with PVRTexToolCLI binaries never touching CGE commit history - I believe we may close this PR.

michaliskambi:
> I've implemented the changes from this PR into #351 with PVRTexToolCLI binaries never touching CGE commit history - I believe we may close this PR.

Thank you, and thanks for thinking about this important detail ("PVRTexToolCLI binaries never touching CGE commit history", good point). Closing this PR then, I merged #351 already.

PR #320, opened 2021-07-17
Include version name in Android package name (by default)

eugeneloza:
Before:

* kryftolike-debug.apk
* kryftolike-release.apk

After:

* kryftolike-1.7.471-android-debug.apk
* kryftolike-1.7.471-android.apk

(adds project version, and does not "inform" about "release" version)

Also honors `PackageNameIncludeVersion`


michaliskambi (changes requested):
1. Do not add `-android`, this is superfluous information. The "a" in `apk` or `aab` already stands for "Android" :) I know that it would be extra future-proof (in case e.g. some day packaging to a common `.zip` on Android would make sense), but right now it seems too unnecessary.

2. Keep the suffix `-release` when it is release. So restore doing

    ```
    PackageName += '-' + PackageModeToName[PackageMode];
    ```

    always, even in release mode. Omitting it is not good -- while the name looks simpler, it is also easy to mistake. A person who would not know this convention (_"release build has no `-debug` in the name"_) would not know whether this is a debug or release build. 

    This way also `TCastleProject.PackageOutput` keeps working, as it uses code: 

    ```
    if { avoid Android packages }
      SameFileName(FileName, Name + '-debug.apk') or
      SameFileName(FileName, Name + '-release.apk') or
      SameFileName(FileName, Name + '-debug.aab') or
      SameFileName(FileName, Name + '-release.aab') or
    ```

    (I just added `.aab` variant, forgot about it earlier :) )

eugeneloza:
> Do not add -android

Hmm... I was thinking more about:

> kryftolike-1.7.471-android.apk
kryftolike-1.7.471-linux-x86_64.deb
kryftolike-1.7.471-linux-x86_64.tar.gz
kryftolike-1.7.471-win32-i386.zip
kryftolike-1.7.471-win64-x86_64.zip

I.e. for consistency reasons. E.g. we still have "linux" in Debian package.

> Keep the suffix -release when it is release

Will do!

michaliskambi:
About `-android` suffix: I see your point. Yes, this looks much more obvious with `-android` in the name. OK, you're right -> let's have `-android` in the name :)

eugeneloza:
> Keep the suffix -release when it is release.

Done in https://github.com/castle-engine/castle-engine/pull/320/commits/1fb5a6f9e7822c6a41d0e82e9a1dcd8f99542ea8 !

eugeneloza:
> SameFileName(FileName, Name + '-debug.apk')

I wonder if it won't fail now as the project version (which may not correspond to the one in Manifest) is included in the name?

michaliskambi:
> I wonder if it won't fail now as the project version (which may not correspond to the one in Manifest) is included in the name?

Ah, it will. The code in 

```
  if { avoid Android packages }
     SameFileName(FileName, Name + '-debug.apk') or
     SameFileName(FileName, Name + '-release.apk') or
     SameFileName(FileName, Name + '-debug.aab') or
     SameFileName(FileName, Name + '-release.aab') or
...
```

doesn't account for the new names at all (-android suffix, possible version number etc.). Please fix it to

```
  if { avoid Android packages }
     IsWild(FileName, Name + '*-android-debug.apk', true) or
     IsWild(FileName, Name + '*-android--release.apk', true) or
     IsWild(FileName, Name + '*-android--debug.aab', true) or
     IsWild(FileName, Name + '*-android--release.aab', true) or
...
```

and test whether it makes sense :) This affects what `castle-engine package-source` does, it should omit the APK / AAB we can generate. (And it is OK to use `*` for any version number, or no version number).

michaliskambi:
And when testing `castle-engine package-source`, note that I just pushed a critical fix to it, in https://github.com/castle-engine/castle-engine/commit/a5bd51df86d526fe06d12888921d294cdcd6f275 . It was accidentally creating by default a zip file, but with extension tar.gz.

michaliskambi (approved):
All good -- thank you and sorry for taking so long to apply!

PR #321, opened 2021-07-18
ASTC format

eugeneloza:
Adds basic support for ASTC Format:

1. Encoding using ASTCENC tool.
2. Decoding by loading an appropriate TGPUCompressedImage

Tested and working both on Desktop and on Android 11 (not 100% sure, as uncompressed images were also packaged for some reason, though I explicitly asked only ASTC, but maybe I've missed some detail)

I see there are tests for DDS images available, but not KTX, maybe it's good idea to add those also for ASTC?


tools/build-tool/code/tooltexturegeneration.pas
```diff
@@ -267,6 +267,39 @@ procedure AutoGenerateTextures(const Project: TCastleProject);
...
+    const ColorspaceOption: String);
+  var
+    ToolExe: String;
+    ToolName: String;
+  begin
+    ToolName := 'astcenc';
+    {$ifdef CPU64}
+      ToolName := 'astcenc-sse2'; // https://github.com/ARM-software/astc-encoder says this will always work on a 64-bit processor
+      //ToolName := 'astcenc-none'; // I have no idea what is it
+      {$if FPC_VERSION >= 3}
```
michaliskambi, line 281: This looks risky. We change the call to an external program (`astcenc-xxx`), depending on FPC version, and on what is reported by some FPC unit. I would not trust that decisions based on this are 100% good for external program. Better (and simpler) to *always* use the same binary.  Just call always `astcenc-sse2`, if this is guaranteed to work on x86_64.

tools/build-tool/code/tooltexturegeneration.pas
```diff
@@ -421,34 +454,36 @@ procedure AutoGenerateTextures(const Project: TCastleProject);
 
         tcETC1:             PVRTexTool(InputFile, OutputFile, C, 'ETC1');
                       // or Compressonator(InputFile, OutputFile, C, 'ETC_RGB');
-        tcASTC_4x4_RGBA:           PVRTexTool(InputFile, OutputFile, C, 'ASTC_4x4');
```
michaliskambi, line 424: For some time, I'd say to keep the PVRTexTool option (and thus, DDS format for ASTC compressed textures) possible, under `{$ifdef CASTLE_ASTC_PVR_DDS}` symbol (documented in `castleconf.conf`).

While I think that new solution is better (`.astc` file format is simpler than `.dds`, `astcenc` is open-source), but for such change I'd say to keep the old version existing. We may want to sometimes test whether it results in the same things, with the same quality, speed. (even if PVRTexTool just calls astcenc for this job -- but then we want to be able to verify this by actually running it e.g. on Unholy/Escape data.)

michaliskambi (changes requested):
Thank you! I have 2 notes.

tools/build-tool/code/tooltexturegeneration.pas
eugeneloza: I understand the concern, but the truth is that each of the next versions works gradually faster (by using some non-standard CPU tricks). This way _if the user device supports this instruction_ it's better to run a better binary. In low FPC versions we just don't have a tool to find out if it's true or not.

tools/build-tool/code/tooltexturegeneration.pas
eugeneloza: PVRTexTool is buggy in this case - and this specific call works only with new version. It would fail to process all the US assets (requiring both height and width be divisors of the NxM), just throws an exception and quits. There isn't much use in keeping it.

michaliskambi: OK, agreed.

tools/build-tool/code/tooltexturegeneration.pas
michaliskambi: If it's that big speedup (I understand 60% smaller time) than I feel convinced :)

In this case I have different request for these lines (didn't notice it earlier):

````
      {$if FPC_VERSION >= 3}
      {$if FPC_RELEASE >= 3}
...
```

only checks the FPC _major_ version and _release_. Is this really what you want? You didn't want to check the _minor_version?

Also it checks FPC_RELEASE always. So e.g. version `10.0.0` would fail this check :)

I usually find it's safest to formulate these checks in reverse, knowing that you only need to care about supported FPC versions anyway ( https://castle-engine.io/supported_compilers.php ). So if your intention was to say that it works only since FPC 3.3.x , you should instead write

```
{$ifndef VER3_0}
{$ifndef VER3_2}
...
```

michaliskambi: @eugeneloza Pinging about this :) I can change the code, just let me know what should be the proper condition. Since which FPC version do we have the necessary identifiers:

- SSE41Support
- AVX2Support 
- SSE42Support 
- POPCNTSupport 
- F16CSupport 

Bear in mind that since https://github.com/castle-engine/castle-engine/pull/350 , we'll bump the required FPC to be >= 3.2.0. So it doesn't really matter whether something is / is not available in e.g. FPC 3.0.x. We only need information whether it is in 3.2.0 (and if not, since *when* is it available).

Looking at CPU unit in FPC 3.2.0, it has AVX2Support.

In FPC 3.2.2 it has both POPCNTSupport and AVX2Support.

Where should I look for others? Are they are in CPU unit but in FPC trunk/main?

eugeneloza: Oh, I've absolutely forgotten that this PR awaits my answer, sorry. I'll look into it ASAP.

eugeneloza: @michaliskambi 

> If it's that big speedup (I understand 60% smaller time) then I feel convinced :)

https://github.com/ARM-software/astc-encoder#prebuilt-binaries

```
For x86-64 we provide, in order of increasing performance:

    astcenc-sse2 - uses SSE2
    astcenc-sse4.1 - uses SSE4.1 and POPCNT
    astcenc-avx2 - uses AVX2, SSE4.2, POPCNT, and F16C

The x86-64 SSE2 builds will work on all x86-64 machines, but it is the slowest of the three. The other two require extended CPU instruction set support which is not universally available, but each step gains ~15% more performance.
```

That is we should have +30% performance for "avx2" version.

> Since which FPC version do we have the necessary identifiers:

* astcenc-sse2: SSE2
* - Seems to be "always supported"
* astcenc-sse4.1: SSE4.1 and POPCNT
* - `SSE41Support` - TRUNK(?=3.3.1)
* - `POPCNTSupport` - FPC 3.2.2
* astcenc-avx2: AVX2, SSE4.2, POPCNT, and F16C
* - `AVX2Support` - FPC 3.0.0
* - `SSE42Support` - TRUNK(?=3.3.1)
* - `POPCNTSupport` - FPC 3.2.2
* - `F16CSupport` - TRUNK(?=3.3.1)

(I think I didn't screw the versions up - derived based on this file history https://gitlab.com/freepascal.org/fpc/source/-/blob/main/rtl/x86_64/cpu.pp). That means that we need trunk version if we want to know if it's safe to use other than SSE2 version.

As an alternative we may grab a copy of that unit in TRUNK or even of those few assembler procedures that are of interest to us - and use it in CGE before they'll make it into some stable build. WDYT?

> only checks the FPC major version and release.

Fixed! (But didn't test) in build.

However, if we copy those procedures into CGE temporarily, this fix won't be needed.

michaliskambi: > However, if we copy those procedures into CGE temporarily, this fix won't be needed.

I'd say we can keep it as you did, i.e. the feature stays disabled when compiled with older FPC versions. It is OK, i.e. the speed of this pre-processing operations is not that critical, and this way code of CGE stays clean, we don't have to remember to maintain some platform/compiler-specific pieces of assembler :)

michaliskambi (approved):

PR #326, opened 2021-07-27
Fix random crashes on Android Adreno devices 

and3md:
Running our 3d game template on all Android devices with Snapdragon (Adreno). Crashes randomly. On Medaitek, Exynos everything works ok. After investigation I found that these crashes are caused by BumpMapping (`bmSteepParallax` and `bmSteepParallaxShadowing` modes).  That's why we can't reproduce that in 2D games.

The solution is even weirder - in some cases `num_steps` can have random value. so We need add clamp:

```
  vec3 v_to_eye = normalize(castle_vertex_to_eye_in_tangent_space);
  float num_steps = mix(30.0, 10.0, v_to_eye.z);
#ifdef CASTLE_BUGGY_BUMP_MAPPING_NUM_STEPS
  num_steps = clamp(num_steps, 10.0, 30.0);
#endif
  float step = 1.0 / num_steps;
```
With this everything works OK.

I tested this fix on:
- Xiaomi Mi 10 Lite (M2002J9G), Snapdragon 765G, Adreno 620:
- Motorola Moto G8 (XT2045-2), Snapdragon 665, Adreno 610
- Motorola Moto G7 Plus (XT2081-2), Snapdragon 460, Adreno 610
- Xiaomi Redmi Note 8, Snapdragon 665, Adreno 610
- Xiaomi Redmi Note 7A Snapdragon 439, Adreno 505

I also created a repository with test projects: https://github.com/and3md/castle-adreno-tests

### DIFF
--- a/src/base/opengl/castleglversion.pas
+++ b/src/base/opengl/castleglversion.pas
@@ -65,7 +65,9 @@ type
     { Intel GPU with Intel drivers. }
     gvIntel,
     { Imagination Technologies (PowerVR) GPU, common on mobile devices. }
-    gvImaginationTechnologies
+    gvImaginationTechnologies,
+    { Qualcomm Adreno, mobile devices }
+    gvQualcomm
   );
 
   TGLVersion = class(TGenericGLVersion)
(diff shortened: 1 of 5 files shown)


michaliskambi (approved):
Thank you! All perfect, merging :)

PR #333, opened 2021-08-27
Spine FFD with curve field

Kagamma:
This PR adds 2 new features:
- New X3D node, `CubicBezierCoordinateInterpolator`. This is basically `CoordinateInterpolator` but with control points.
- We now process `curve` field while playing Spine FFD animations. This is handled by using new X3D node above.

A modified version of Spine FFD demo, which include `curve` field can be found here: https://github.com/castle-engine/demo-models/pull/8


michaliskambi:
Thank you!

Everything is perfect, in particular I see how your

- `ReadCurveControlPointsTimeOnly` is an analogy to existing `ReadCurveControlPoints` 

- and the generic `TAbstractCubicBezierArrayInterpolatorNode` is a nice solution to then easily get specialized `TCubicBezierCoordinateInterpolatorNode`.

Merging, will also merge https://github.com/castle-engine/demo-models/pull/8 right away.

michaliskambi (approved):

PR #337, opened 2021-08-31
Allows to edit TCollectionItem items in castle-editor

Kagamma:
The property editor will show selected TCollectionItem item's properties while we are editing, and point back to previous component when the TCollectionItem list form is closed.

image


tools/castle-editor/code/framedesign.pas
```diff
@@ -2299,6 +2308,55 @@ procedure TDesignFrame.PropertyEditorModified(Sender: TObject);
...
+end;
+
+procedure TDesignFrame.PropertyGridCollectionItemClose(Sender: TObject; var CloseAction: TCloseAction);
+begin
+  UpdateSelectedControl;
+end;
+
+procedure TDesignFrame.PropertyGridCollectionItemAdd(Sender: TObject);
+begin
+  (TButton(Sender).Parent.Parent as TCollectionPropertyEditorForm).actAddExecute(Sender);
```
michaliskambi, line 2338: Better use here safer `(Sender as TButton)` than `TButton(Sender)`.

Speed doesn't matter for this code, and this way we'll get nice exception in case of an error :)

tools/castle-editor/code/framedesign.pas
```diff
@@ -2647,7 +2708,38 @@ procedure TDesignFrame.UpdateSelectedControl;
       for I := 0 to SelectedCount - 1 do
         SelectionForOI.Add(Selected[I]);
       for InspectorType in TInspectorType do
+      begin
         Inspector[InspectorType].Selection := SelectionForOI;
+        if not IsCollectionItemEditFormInitialized then
```
michaliskambi, line 2713: Please move this section into a nested routine inside `TDesignFrame.UpdateSelectedControl;`, like `InitializePropertyEditorForms`. You can then move the new local variables (Row, J, Ed, Fm) there, and thus the `TDesignFrame.UpdateSelectedControl` implementation will be easier to read. Like

```pascal
procedure TDesignFrame.UpdateSelectedControl;

  procedure InitializePropertyEditorForms;
  var
    J: Integer;
    Row: TOIPropertyGridRow;
    Ed: TCollectionPropertyEditor;
    Fm: TCollectionPropertyEditorForm;
  begin
  end;

begin
  ...
end;
```

tools/castle-editor/code/framedesign.pas
```diff
@@ -2647,7 +2708,38 @@ procedure TDesignFrame.UpdateSelectedControl;
...
+        begin
+          for J := 0 to Inspector[InspectorType].RowCount - 1 do
```
michaliskambi, line 2715: Looking at the usage of `Continue` and `Break` in this loop, if effectively searches for the 1st row that satisfies `Row.Editor is TCollectionPropertyEditor`, and handles only it. So I would write it more explicitly, as 2 steps,

- find such row
- act on this row

like 

```pascal
Ed := nil;

// search for row with TCollectionPropertyEditor
for J := 0 to Inspector[InspectorType].RowCount - 1 do
begin
  Row := Inspector[InspectorType].Rows[J];
  if Row.Editor is TCollectionPropertyEditor then
  begin
    Ed := TCollectionPropertyEditor(Row.Editor);
    Break;
  end;
end;

// if row with TCollectionPropertyEditor found, act on it
if Ed <> nil then
begin
  Fm := TCollectionPropertyEditorForm(Ed.ShowCollectionEditor(nil, nil, ''));
  ..
end;
```

michaliskambi (changes requested):
Thank you!

I added some comments about the code.

As a more general note, looking at `IsCollectionItemEditFormInitialized` variable (it is set `false` only in constructor), it seems that the main logic (that does `Ed.ShowCollectionEditor`) will execute only once for the whole design lifetime. And it will execute as soon as one selects some component with `TCollectionPropertyEditor`. It seems we should instead:

- Execute `Ed.ShowCollectionEditor` when user clicks on some "..." inside the collection item row (we should not need the code to find "the first row that has TCollectionPropertyEditor`" -- because it's bad if component has multiple properties that have `TCollectionPropertyEditor`).

- Be able to execute it many times. If we do `Ed.ShowCollectionEditor`, and later user closes this form, user should be able to invoke it again.

Maybe instead of boolean `IsCollectionItemEditFormInitialized` we should instead have a boolean like `IsCollectionItemEditFormOpen`, that is reset to `false` if the collection window is closed? So that user can open a collection editor (for the same or different property) again.

Kagamma:
IsCollectionItemEditFormInitialized is necessary. CollectionPropertyEditorForm only create when users open the editing form first time and then live for the rest of the application's lifetime. Thus we need to reassign events to this form whenever we switch projects.

 I make a call to ShowCollectionEditor in order to force the application to create CollectionPropertyEditorForm beforehand, assign necessary events and then close/hide it. It has nothing to do with the "..." button on the property editor.

tools/castle-editor/code/framedesign.pas
Kagamma: Resolved in 61f89d3

tools/castle-editor/code/framedesign.pas
Kagamma: Resolved in 0362e42

tools/castle-editor/code/framedesign.pas
Kagamma: Resolved in 0362e42

michaliskambi (approved):
Sorry for delay -- merging. Your explanation of the purpose of `IsCollectionFormInitialized` (and why we have to do it only once for the form lifetime, on the first row with prop editor) makes sense. I will actually paste it into a comment into the code soon :)

Note: closing the project form before closing the property form will later make access violation if you try to close the property form (makes sense, as the instance with `PropertyGridCollectionItemClose` callback no longer exists). I'll address it soon too.

I tested that it works on https://github.com/castle-engine/castle-db-aware-controls -- I can now edit `FieldDefs` of `TDbf` component (FPC component, independent of CGE) in the CGE editor.  I also tested `Anchors` on your https://github.com/Kagamma/cge-3d-particle-emitter .

michaliskambi:
Pushed related https://github.com/castle-engine/castle-engine/commit/1a0cd6b3e78f049913862f97412f1785c90c5018

PR #338, opened 2021-08-31
Delphi Port (WIP)

and3md:
First massive changes for Delphi. Still not compiling - WIP. Most of changes on Delphi side not tested - I'm still fixing differences between FPC and Delphi to get something working. I think this PR should not break FPC but I can't give 100% guarantee. I'm not sure if it should be merged now, @michaliskambi should decide :)

## The plan
We want to compile the engine first and then improve the code. The point is to not break compilation on Free Pascal by changing many APIs in one go. And don't waste time now for polishing something that can change quickly. So currently there are a lot of defines, some of them can be removed by introducing some abstractions in the future.  

So the plan is:
1. Port most of modules and see how the sources will look like <- we are still here
2. See Castle Engine Window :)
3. Collect ideas/conclusions on how to make the code more readable (I think first version will look ugly.)
4. First code refactoring - at this point we should know want we want and how to do that ;)
5. Write a lot tests,
6. Fix issues and not working functions
7. Port missing units
8. Second code refactoring: More integrate FPC and Delphi code (unit by unit) , maybe introducing some abstractions, CAREFULLY


michaliskambi:
Cool. Let's see what Jenkins says (after we solve conflicts).

You probably wanted to invert the logic when to define  CASTLE_DEPRECATED_WINDOW_CLASSES in castleconf.inc :)

For loading symbols from DLL like zlib we can for sure find some syntax that works in both.

michaliskambi:
The new PR replaces this: https://github.com/castle-engine/castle-engine/pull/350 . 

- It seems sufficiently stable.

- It includes all relevant commits (`delphi` branch was merged into `delphi_next`). 

So all the work and commenting can continue within https://github.com/castle-engine/castle-engine/pull/350 . I'm closing this.

PR #340, opened 2021-09-06
Adds support (partially) OpenGL ES 3.0

Kagamma:
This PR adds a new OpenGL ES 3.0 header. Currently it contains only necessary APIs to set up Transform Feedback on mobile devices.
A new flag `OpenGLES3` allows to switch between OpenGLES 2.0 & OpenGLES 3.0. The flag is enabled by default at the moment.

### DIFF
--- a/packages/castle_base.lpk
+++ b/packages/castle_base.lpk
@@ -74,7 +74,7 @@ This is the same license as used by Lazarus LCL and FPC RTL.
 See https://castle-engine.io/license.php for details.
 "/>
     <Version Major="6" Minor="5"/>
-    <Files Count="935">
+    <Files Count="936">
       <Item1>
         <Filename Value="../src/3d/auto_generated_persistent_vectors/tcastlecamera_persistent_vectors.inc"/>
         <Type Value="Include"/>
@@ -3842,6 +3842,10 @@ See https://castle-engine.io/license.php for details.
         <Filename Value="../src/x3d/auto_generated_node_helpers/x3dnodes_cubicbeziercoordinateinterpolator.inc"/>
         <Type Value="Include"/>
       </Item935>
+      <Item936>
+        <Filename Value="../src/base/opengl/castlegles30.pas"/>
+        <UnitName Value="CastleGLES30"/>
+      </Item936>
     </Files>
     <RequiredPkgs Count="1">
       <Item1>
--- a/src/base/opengl/castlegles30.pas
+++ b/src/base/opengl/castlegles30.pas
@@ -1,0 +1,1964 @@
+{ OpenGL ES 3.0 API.
+
+  A modified version of OpenGL ES 2.0 headers by Trung Le (kagamma), which adds
+  support for OpenGL ES 3.0 APIs.
+  Currently it only support necessary APIs for setting up Transform Feedback.
+}
+
+{** OpenGL ES 2.0 headers
+ **
+ ** Ported/Translated for FreePascal by Benjamin 'BeRo' Rosseaux
+ ** benjamin@rosseaux.com - http://www.rosseaux.com
+ **
+ ** EGL part:
+ **
+ ** Copyright (c) 2007-2009 The Khronos Group Inc.
+ **
+ ** Permission is hereby granted, free of charge, to any person obtaining a
+ ** copy of this software and/or associated documentation files (the
+ ** "Materials"), to deal in the Materials without restriction, including
+ ** without limitation the rights to use, copy, modify, merge, publish,
+ ** distribute, sublicense, and/or sell copies of the Materials, and to
+ ** permit persons to whom the Materials are furnished to do so, subject to
+ ** the following conditions:
+ **
+ ** The above copyright notice and this permission notice shall be included
+ ** in all copies or substantial portions of the Materials.
+ **
+ ** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ ** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ ** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ ** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ ** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ ** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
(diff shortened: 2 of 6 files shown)


michaliskambi:
Thank you!

My idea for making this perfect in CGE:

1. I would avoid making entirely new `CastleGLES30` unit. As GL ES 3.0 is a superset of GL ES 2.0 (that is, they didn't remove/deprecate anything), I believe we should instead rename the current unit `CastleGLES20` to more generic `CastleGLES` and just maintain there "2.0 things + 3.0 additions" in one unit.

   The header could always try to load "3.0 additions" (function entry points), the implementation will have to check and only use them when GLES3 context is actually available (when `GLFeatures.VersionES_3`, see below).

2. It should not be activated by a compile-time switch, and it must have a fallback on GLES2.

   So we should try to initialize GLES3 context, and reliably fallback on OpenGLES2. We don't want to "cut off" old devices that only support OpenGLES2.

   So modification to `castleconf.inc` should be needed.

   In turn, modification of `castlewindow_egl.inc` should be more involved: try GLES3, and if it fails -> go with GLES2.

3. The `TGLFeatures` class should have additional field+property `VersionES_3`, somewhat similar to existing `Version_X_Y` on desktop OpenGL. It should be set in constructor like

    ```
    {$ifdef OpenGLES}
    FVersionES_3 := GLVersion.AtLeast(3, 0);
    {$endif}
    ```

    This detection works regardless of the context-creation method. So it will work whether context is initialized by `castlewindow_egl.inc` (non-iOS) or `OpenGLController.m` iOS.

4. There should be no `$ifdefs` for GLES3 in effect. It is decided at runtime, so it should be checked at runtime. So `castleglshaders.pas` should check `if GLFeatures.VersionES_3 ...` before using transform feedback, not `{$ifdef OpenGLES3}`.

    This is similar to (desktop) OpenGL usage. We prefer the version checks at runtime, to be able to work everywhere -- use new features where possible, but work also with older devices.

5. On iOS, the code for this is special: in Objective-C implementation. I see on Discord you already noticed it, indeed it is in `castle-engine/tools/build-tool/data/ios/xcode_project/cge_project_name/OpenGLController.m`.

Kagamma:
Done. Now it will try to load appropriate APIs depends on OpenGL ES version.
I didn't touch iOS code, as I don't have any device to test and don't want to commit untested code.

michaliskambi:
1. I simplified initialization a bit: just a single `GLESInitialization` routine, that always tries to initialize latest OpenGLES.

    I also removed the calls to `GLESXxxInitialization` from `TGLFeatures.Create`, as I think it is too late. When `TGLFeatures.Create` happens, we already have queried the context using `glGetString` that we loaded from GLES library. So it seems risky that we would then reload the library (potentially changing the `glGetString` implementation).

    So it seems both simpler and safer to "just" try to load the latest GLES library in `GLESInitialization`. If the GLES3 is available under different names on some systems, we'd just have to try first the names with "3", and then with "2" in `GLESInitialization`.

    From what I can tell, GLES3 libraries do not really occur under different names now. Note that on Linux, Mesa only provides GLES2 (there are `libgles2` packages in Debian, but no `libgles3`). I'm not sure whether Mesa implements GLES3.

2. I also added lines to `GLInformationString` (that is automatically output to log when application initializes) to debug the values of `VersionES_X_Y` you added. "The Unholy Society" now rocks on my Android with GLE3 context :) Thanks!

```
09-08 00:38:30.042  8074 11212 I unholy_society: -------------------- Rendering Context Initialization begin
09-08 00:38:30.042  8074 11212 I unholy_society: OpenGLES information (detected by unholy_society):
09-08 00:38:30.042  8074 11212 I unholy_society: 
09-08 00:38:30.042  8074 11212 I unholy_society: --------
09-08 00:38:30.042  8074 11212 I unholy_society: Version:
09-08 00:38:30.042  8074 11212 I unholy_society:   Version string: OpenGL ES 3.2 v1.r26p0-01eac0.###other-sha0123456789ABCDEF0###
09-08 00:38:30.042  8074 11212 I unholy_society:   Version parsed: major: 3, minor: 2, release exists: False, release: 0, vendor-specific information: "v1.r26p0-01eac0.###other-sha0123456789ABCDEF0###"
09-08 00:38:30.042  8074 11212 I unholy_society:   Vendor-specific version parsed: major: 0, minor: 0, release: 0
09-08 00:38:30.042  8074 11212 I unholy_society:   Vendor: ARM
09-08 00:38:30.042  8074 11212 I unholy_society:   Vendor type: Unknown
09-08 00:38:30.042  8074 11212 I unholy_society: 
09-08 00:38:30.042  8074 11212 I unholy_society:   Renderer: Mali-G71
09-08 00:38:30.042  8074 11212 I unholy_society:   Fglrx (ATI on Linux): False
09-08 00:38:30.042  8074 11212 I unholy_society:   Mesa: False
09-08 00:38:30.042  8074 11212 I unholy_society: 
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy glGenerateMipmap(EXT): False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy GL_LIGHT_MODEL_TWO_SIDE: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy VBO: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy shader shadow map: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy FBO rendering to multi-sampling texture: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy FBO rendering to cube map texture: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy swap buffers with non-standard glViewport: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy 32-bit depth buffer: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy GLSL gl_FrontFacing: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy GLSL read varying: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy Pure Shader Pipeline: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy Texture Size Above 2048: False
09-08 00:38:30.042  8074 11212 I unholy_society:   Buggy Shader Bump Mapping num_steps: False
09-08 00:38:30.042  8074 11212 I unholy_society: 
09-08 00:38:30.042  8074 11212 I unholy_society: ------------------------
09-08 00:38:30.042  8074 11212 I unholy_society: Versions available:
09-08 00:38:30.042  8074 11212 I unholy_society: 
09-08 00:38:30.042  8074 11212 I unholy_society:   2_0: True
09-08 00:38:30.042  8074 11212 I unholy_society:   3.0: True
09-08 00:38:30.042  8074 11212 I unholy_society:   3.1: True
09-08 00:38:30.042  8074 11212 I unholy_society:   3.2: True
09-08 00:38:30.042  8074 11212 I unholy_society: 
```

PR #341, opened 2021-09-08
Full OpenGL ES 3.0 header

Kagamma:
OpenGL ES 3.0 APIs translate directly from Khronos's C header https://www.khronos.org/registry/OpenGL/api/GLES3/gl3.h with the help of `h2pas`.


michaliskambi (approved):
Looks cool, merging -- thank you!

PR #351, opened 2021-12-11
Third party tools

eugeneloza:
This is a better version of https://github.com/castle-engine/castle-engine/pull/319 (cherry-picking one commit from there and adding ASTCENC - thus branched from `astc-format` branch).

See comments in https://github.com/castle-engine/castle-engine/pull/319#issuecomment-915971466

Now for the harder part:

* NVIDIA texture tools in latest version 3.1.6 changed their license from MIT into something completely bizarre - including 18+ age restrictions
[LICENSE.TXT](https://github.com/castle-engine/castle-engine/files/7697911/LICENSE.TXT)
Maybe it's reasonable to use older version 2.1.0 from https://github.com/floooh/oryol/tree/master/tools ?
Or maybe it was just wrongly assumed to be MIT in that repository? I believe we should double-check that.

* Compressonator CLI now weights 150 Mb as of 4.2.5150. Are we sure we want to include that much? (Compare to 2Mb for ASTCENC and 0.4 MB for nvcompress)

I've added "searching for them" in the corresponding folders, but didn't add the binaries.


michaliskambi:
ASTCENC (current state of this PR) - everything looks perfect, merging :)

I started researching and wrote a loooong answer about NVidia and Compressonator tools -- I will put it on forum instead and link from here.

michaliskambi:
I created a thread https://forum.castle-engine.io/t/texture-compression-tools-integrate-in-cge/475 with my thoughts on how to continue :)

Issue #353, opened 2021-12-14
Editor: Access violation when try drag a Button to HorizontalGroup


and3md:
Just a reminder to investigate after I finish my current task.

Steps to reproduce: 
1. Create new project with empty template
2. Open design
3. Add button
4. Add horizontal group
5. Try drag button to horizontal group -> Access violation

System: Windows 10
Current master.

michaliskambi:
Tested now quickly on Linux/x86_64, works OK, dragged button lands inside a horizontal group without any error.

With latest master, FPC 3.2.2.

![Zrzut ekranu z 2021-12-15 17-21-58](https://user-images.githubusercontent.com/1076978/146224527-044825ac-430e-450f-b9f8-4365d02f4c64.png)

eugeneloza:
Unfortunately I cannot reproduce it either. Neither on binaries downloaded from the official site just a few seconds ago, nor on my local build (with some quite obsolete CGE version). Windows 10x64.

However, from time to time I run into this bug. I couldn't find a way to reproduce it. For me it's usually after using Castle Editor for some long time and then trying to drag something around (either the last component or into the last component of some hierarchy level). It feels to have something to do with the Undo System, but I could never get neither the proper debug line nor (even unreliable) reproduction steps.

eugeneloza:
This bug jumps up from time to time, very rarely, but still does:

https://trello.com/c/YizbsBMD/110-bug-access-violation-in-castle-editor-when-drag-n-dropping-items-in-hierarchy

https://trello.com/c/hwVKI010/114-bug-access-violation-in-castle-editor-when-changing-the-name-of-the-component-in-hierarchy-and-trying-to-drag-it-by-mouse-in-edi

Also this one may be somehow related: https://trello.com/c/V6v2rBwv/75-bug-access-violation-in-castle-editor

and3md:
@michaliskambi I didn't have any editor crashes when I made platformer on Linux so I think that can be only Windows related issue. 
@eugeneloza I don't remember if I used Undo before AV. But the designs was a lot simpler than level in platformer demo. And I had two AV in one day (on Windows). I will check that more carefully and maybe I'll find a case that can be reproduced every time.

eugeneloza:
> I don't remember if I used Undo before AV

[...]

PR #354, opened 2021-12-21
Fix crash when drag'n'drop nodes in hierarchy on Windows

and3md:
This PR fixes https://github.com/castle-engine/castle-engine/issues/353

Also fixed selection bug after drop and removed hierarchy reload after drag'n'drop TCastleInterface (transforms need still do hierarchy reload).


and3md:
ValidateHierarchy implemented :)

michaliskambi (approved):
Looking excellent, thank you!

PR #356, opened 2022-01-06
Fix lazarus release candidate version parsing

eugeneloza:
Fixes https://forum.castle-engine.io/t/0rc1-is-an-invalid-integer/492/3 however, only for this typical case of Lazarus 2.2.0RC1. Also added a test to make sure parsing goes as expected.

As this fix covers all current possible versioning variants for Lazarus on Windows, however in the past they've had quite bizarre version systems:

2022-01-06-093917_1366x768_scrot

And current implementation will fail for such versioning, e.g.

* 0.9.30.2 will result in 0.9.30 version and will be seen as absolutely equal to 0.9.30.4RC3
* 1.0RC2 will throw an exception `0RC2 is invalid integer`

I'm not sure if we can "ignore" this. But anyway we now support two latest Lazarus release candidates versioning :)

### DIFF
--- a/tests/CastleEngineManifest.xml
+++ b/tests/CastleEngineManifest.xml
@@ -16,6 +16,7 @@
     <search_paths>
       <path value="code/" />
       <path value="code/common/" />
+      <path value="../tools/common-code/" />
     </search_paths>
   </compiler_options>
 </project>
--- a/tests/code/testtoolfpcversion.pas
+++ b/tests/code/testtoolfpcversion.pas
@@ -33,36 +33,45 @@ uses SysUtils,
   CastleFilesUtils, CastleURIUtils;
 
 type
+  { Descendant of TToolVersion, just to expose protected ParseVersion.
+    We actually don't need to make it public here, it's enough that TTestVersionParsing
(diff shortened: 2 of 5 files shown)


eugeneloza:
Note, that I was unable to test this feature in "real life conditions". I've made sure that the version strings like "2.2.0RC1" would parse appropriately, but there is no way to install `RC*` from FpcUpDeluxe (I've created issue there) and I didn't install the binaries directly from Lazarus.

eugeneloza:
Thanks to FpcUpDeluxe dev's help - I've got the RC version of Lazarus and everything is working as expected:
2022-01-06-161330_1366x768_scrot

michaliskambi:
Thank you! I used your idea, though I changed the way ReleaseRemark parsing works a bit. I think my approach is a bit simpler :)

I just use `NonDigits` for delimiters passed to `NextToken`, and then at the end I can just do

```
ReleaseRemark := SEnding(S, SeekPos - 1)
```

to get ReleaseRemark. It is also even more "resilient to weird things". Using NonDigit for token delimiters, it means that they really should be valid digits (or ''). 

I also extended the auto-tests, this was a great point to add them.

michaliskambi (approved):

Issue #324, opened 2021-07-25
Implement TTexturePropertiesNode.BoundaryModeXxx and MIRRORED_REPEAT mode


Free-Pascal-meets-SDL-Website:
Hi there,

well, title says it all. I wonder if the wrap mode MIRRORED_REPEAT (see https://www.web3d.org/specifications/X3Dv4Draft/ISO-IEC19775-1v4-CD/Part01/components/texturing.html#t-TextureBoundaryModes) is implemented?

On
`TextureNode.TextureProperties.BoundaryModeS := 'MIRRORED_REPEAT';`
and
`TextureNode.TextureProperties.BoundaryModeT := 'MIRRORED_REPEAT';`

 and translating the texture out of the base S,T-space, the texture just repeats but isn't mirrored.

And there doesn't seem to be any kind of error checking of sensible inputs, as "fantasy modes" are also accepted and returned.

So, that is why I'm asking. :)

Best regards
Matthias

michaliskambi:
The `TTexturePropertiesNode.BoundaryModeXxx` is right now completely ignored (which is also why we don't even make a warning in case of invalid values, in Pascal or in X3D file).

If you need it, let me know -- I'll leave this issue open :) It is rather easy to add support for it. (Note that it will override the `repeatS/T/R` settings at texture nodes.) The property will become a Pascal enumerated value then, not a String.

And we can implement "mirrored repeat". Note that some very old versions of desktop OpenGL may not support it, but it is supported on modern desktops and on all mobile devices we support (as it is present in OpenGLES 2):
- https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_mirrored_repeat.txt
- http://docs.gl/gl4/glTexParameter
- https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glTexParameter.xml

Free-Pascal-meets-SDL-Website:
Thanks for your reply.

> If you need it, let me know -- I'll leave this issue open :) It is rather easy to add support for it. (Note that it will override the `repeatS/T/R` settings at texture nodes.) 

This mode would be highly useful to implement support for flipped tiles in the converter (https://github.com/Free-Pascal-meets-SDL-Website/CastleConvertTiledMap).

> The property will become a Pascal enumerated value then, not a String.

I appreciate this :).

Free-Pascal-meets-SDL-Website:
[...]

PR #358, opened 2022-01-09
Implement boundary mode MIRRORED_REPEAT (and others)

Free-Pascal-meets-SDL-Website:
This PR implements (formally) all X3D boundary modes, which are available, including MIRRORED_REPEAT. It is not meant to be merged. Michalis, I would appreciate if you review this PR before I go on working on it. Any hint is very welcome, as always. Thanks in advance.

According to [X3D specs](https://www.web3d.org/specifications/X3Dv4Draft/ISO-IEC19775-1v4-CD/Part01/components/texturing.html#t-TextureBoundaryModes) the boundary modes are: CLAMP, CLAMP_TO_EDGE, CLAMP_TO_BOUNDARY, MIRRORED_REPEAT, REPEAT.

According to [GL specs](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glTexParameter.xml) these can be translated to GL_CLAMP_TO_EDGE, GL_MIRRORED_REPEAT, GL_REPEAT. Also, the mode GL_CLAMP is known.

This PR:
- implements a type TBoundaryMode used for all three tex. coord. axis s, t, r
- uses TBoundaryMode instead of strings to read/write boundary mode for tex. coord.
- works only for TGLImageTextureNode atm. (see attached example code)

TODO
- update x3dnodes_textureproperties.inc file to prevent conflicts (see issue #324)
- should TBoundaryMode be part of RenderOptions?
- extend boundary mode to other GL texture objects (multi tex., movie,...)
- pack all three modes in record to call in HandleTextureProperties proc. to save two arguments?
- remove repeatS/T/R properties finally + what to do with TextureRepeatToGL function?

Example code + image file (should be in data folder)
[mirrorrepeat.lpr.txt](https://github.com/castle-engine/castle-engine/files/7834434/mirrorrepeat.lpr.txt)
test512


michaliskambi:
Thank you! 

1. I did provide some notes in my answer to https://github.com/castle-engine/castle-engine/issues/324 -- basically I'd say to remake `TextureProperties` to have `auto-generate-more: true` in `Texturing.txt`, this will make your modifications using my new "better approach" to X3D nodes.

2. I'll look at the rest of your modifications ASAP. 

3. As for `CLAMP_TO_BOUNDARY`: while we should parse it (that is, allow in X3D `CLAMP_TO_BOUNDARY` string) I don't think we should really handle it. That is, our renderer should just make a warning, and behave just like `CLAMP_TO_EDGE`.  Because properly handling `CLAMP_TO_BOUNDARY` means complicating our implementation for little gain. The "boundary" is a concept from OpenGL textures (that's how it got into X3D) but I don't think it really got much real-world usage. Because:
    - It is not present in OpenGLES 3 ( https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glTexParameter.xhtml , allows only GL_CLAMP_TO_EDGE, GL_MIRRORED_REPEAT, or GL_REPEAT). Though it remains supported in OpenGL 4 ( https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexParameter.xhtml ).
    - glTF doesn't support it ( https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html , gltF only has "Repeat", "Mirrored Repeat", "Clamp to edge"). 
    - I don't know whether there's any tool that actually generates X3D models (or other asset formats) that require using such wrap mode. Note that to support it, one should really also support `borderColor` (otherwise the border is always considered black), so this makes supporting it more complicated.

4. For the same reason, "Clamp" should be just an alias for "Clamp to edge" (X3D spec doesn't say this explicitly, but I think this is the only reasonable interpretation -- matches the possibilities of OpenGLES and glTF).

Free-Pascal-meets-SDL-Website:
Thanks for your reply. I will try to update this PR according to your details here and in issue #324 asap. :-)

michaliskambi:
Note: I'm doing now a significant rearrangement of units in CGE master branch, it touches some of the files you're working on. 

The subdirectory `src/x3d/` is now called `src/scene/`, with stuff related to X3DFields and X3DNodes units in `src/scene/x3d/` . The detailed description of subdirectories is now in https://castle-engine.io/units_map .

If you'll later see a message about some conflicts with the `master` branch in CGE in this PR, don't be alarmed :) I'll take care to resolve them before merging PR (if GIT will not be smart enough to deal with them automatically).

Free-Pascal-meets-SDL-Website:
I see, thanks for the hint.

michaliskambi:
Thank you! I'll look at this ASAP.

As for:

> some example models don't look the same as before, e. g. background/image_background/background_image_partial.x3dv; texturing_advanced/tex3d_repeat_modes.x3dv; texturing_advanced/inlined_pixel_textures.x3dv => Expalantion: All of them rely on the TextureNode's repeatS/T/R fields and at the same time provide a TexturePropertyNode which according to spec. should lead to ignoring the repeat fields but use the boundary mode field of the TexturePropertyNode (which is correctly implemented now) [see 18.4.9]

Yeah, that is expected and correct. As you say, we'll just have to fix these demo files separately, to specify repeat modes inside TextureProprerties.

michaliskambi:
Sorry, I know it takes a while, but I do have this in my TODO. I'm now going for a short vacation, but I very much want to review + apply this next week.

Free-Pascal-meets-SDL-Website:
That's great! Looking forward to your review/application.

Happy vacation!

michaliskambi:
Thank you and again sorry it took so long!  In short: 

- I reviewed it, made minor adjustments, everything looks cool. I always find some minor adjustment to make :), but overall I just used your approach for where to handle this, all is great.

- I'll want to do additional tests later today (grep CGE and demo-models for TextureProperties usage and eventually fix them, as we talked earlier -- to account that now TextureProperties.boundaryModeXxx will override repeatS/T/R). I'll apply it then.

Some details:

- I rebased this PR on master, to resolve all conflicts.

- I did small fix for Delphi (needs `overload` keyword, unlike FPC)

- I added `chTexturePropertiesNode` to all TextureProperties fields changes. This is not used by CGE *for now* -- but it was on master and I do plan to use it, to better react to changes at runtime to TextureProperties fields.

- I regenerated the files using x3d-nodes-to-pascal, to make sure we make no mods to the auto-generated files. I removed the unnecessary TTexturePropertiesNode.FdMagnificationFilter/FdMinificationFilter in non-auto-generated code (x3dnodes_standard_texturing.inc) and now it plays nicely with auto-generated code :)

- Treat X3D "CLAMP" as "CLAMP_TO_EDGE", and support ancient OpenGLs without GL_CLAMP_TO_EDGE too

- Used `GLenum` instead of `TGLenum`, used `const` for `RepeatXxx` params (no big reason, it's just what we prefer to do in the rest of CGE code; admittedly even my own older code didn't always follow it, so I fixed that too :) ).

- I added to demo-models ( https://github.com/castle-engine/demo-models/ ) a quick demo in texturing_advanced/texture_boundary_modes.x3dv and it works OK.

texture_boundary_modes_0

michaliskambi:
- I upgraded various CGE code in https://github.com/castle-engine/castle-engine/ to account for the fact that TTexturePropertiesNode.BoundaryModeS/T/R now overrides texture repeatS/T/R - sprite sheets, TCastleImageTransform, glTF reading (we also support now glTF MirroredRepeat thanks to this, yay!), anisotropic example.

- I updated everything in https://github.com/castle-engine/demo-models/ to account for it too.

- I updated https://castle-engine.io/x3d_implementation_texturing.php about various TextureProperties details.

I think we're done :) Merging. Thank you!

tex3d_repeat_modes_0
texture_boundary_modes_0

michaliskambi (approved):

Free-Pascal-meets-SDL-Website:
Wow, thanks for extending, refining and integrating the PR to be successfully merged. Quite interesting, as always. I also like the acorn demo :-).

This also closes issue #324 .

PR #365, opened 2022-01-12
fixed castlewindow_winapi.inc to use widechar

HotariTobu:
I have an issue that view3dscene can't load a scene file of a path with multibyte characters because of garbling.
To improve this, I changed `TCastleWindowBase.BackendFileDialog` in `castlewindow_winapi.inc` to use `WideChar`.
In addition, I added these.

- a definition of struct `TOpenFilenameW` in `src\window\windows\castlewindow_winapi.inc`
- bindings to `comdlg32.dll` in `src\window\windows\castlewindow_winapi.inc`
- a function `PWideCharOrNil` in `src\base\castlestringutils.pas`

Former `castlewindow_winapi.inc` seemed to use the Ansi version of WinAPI (A suffix).

Environment:
Windows10 21H2 AMD64
Free Pascal Compiler version 3.2.2 [2021/05/15] for i386

### DIFF
--- a/src/window/windows/castlewindow_winapi.inc
+++ b/src/window/windows/castlewindow_winapi.inc
@@ -1423,7 +1423,7 @@ function TCastleWindowBase.BackendFileDialog(const Title: string; var FileName: 
 
 var
   Struct: TOpenFileNameW;
-  FileNameBuffer: array [0..Pred(MAX_PATH+10)] of widechar;
+  FileNameBuffer: array [0..6000] of widechar;
   {$ifndef FPC}FileNameString: String;{$endif}
   FilePath: string;
   Mode: TGLModeFrozenScreen;
@@ -1564,8 +1564,8 @@ begin
   Mode := TGLModeFrozenScreen.Create(Self);
   try
     if OpenDialog then
-      Result := GetOpenFileNameW(LPOPENFILENAME(@Struct)) else
-      Result := GetSaveFileNameW(LPOPENFILENAME(@Struct));
+      Result := GetOpenFileNameW({$ifdef FPC}@{$endif}Struct) else
+      Result := GetSaveFileNameW({$ifdef FPC}@{$endif}Struct);
   finally Mode.Free end;
 
   if Result then
--- a/src/window/windows/castlewindow_winapi.inc
+++ b/src/window/windows/castlewindow_winapi.inc
@@ -1336,11 +1336,6 @@ procedure TCastleWindowBase.CreateBackend;
 begin
 end;
 
-{$IFDEF FPC}
-function  GetOpenFileName(var OpenFile: TOpenFilenameW): Bool; stdcall; external 'comdlg32.dll' name 'GetOpenFileNameW';  // WideVersion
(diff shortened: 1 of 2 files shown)


michaliskambi:
Thank you -- indeed we had a TODO in `TCastleWindowBase.BackendFileDialog` 

```
{ TODO: Use Unicode version of WinAPI (W suffix) here. }
```

Everything else in `castlewindow_winapi.inc` uses Unicode versions now, and `TCastleWindowBase.BackendFileDialog`  should follow.

Why do you need to define `tagOFNW` and (for FPC) GetOpenFileName? It seems that FPC Windows header already includes `GetOpenFileNameW` and the related structure (`OPENFILENAME` or `OPENFILENAME_NT4`).

HotariTobu:
I don't know FPC Windows header already includes 'GetOpenFileNameW'.
I deleted the definition of struct `TOpenFilenameW` and changed the type of variable Struct from `TOpenFilename` to `TOpenFilenameW`.

michaliskambi:
Why do you still need to define 

```
function  GetOpenFileName(var OpenFile: TOpenFilenameW): Bool; stdcall; external 'comdlg32.dll' name 'GetOpenFileNameW';  // WideVersion
function  GetSaveFileName(var OpenFile: TOpenFilenameW): Bool; stdcall; external 'comdlg32.dll' name 'GetSaveFileNameW';  // WideVersion
```

? Can you use the `GetOpenFileNameW` / `GetSaveFileNameW` from FPC Windows headers?

HotariTobu:
I deleted the definitions of `GetOpenFileNameW` and `GetSaveFileNameW` and cast `Struct` into `LPOPENFILENAME` to use the FPC Windows headers' functions.

src/window/windows/castlewindow_winapi.inc
```diff
@@ -1422,8 +1422,8 @@ function TCastleWindowBase.BackendFileDialog(const Title: string; var FileName:
   end;
 
 var
-  Struct: TOpenFileName;
-  FileNameBuffer: array [0..6000] of char;
+  Struct: TOpenFileNameW;
+  FileNameBuffer: array [0..Pred(MAX_PATH+10)] of widechar;
```
michaliskambi, line 1426: I see you change 6000 into `Pred(MAX_PATH+10)`. What is the reason?

MAX_PATH is 260 (but you can have longer paths, and ideally everything works with it, if you adjust Windows registry settings --- https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd ). So our 6000 was allowing to pass longer paths.

Granted, hardcoding 6000 is ugly :) But then, `Pred(MAX_PATH+10)` means we just hardcode a smaller number, because file paths may still be longer than MAX_PATH.

michaliskambi (changes requested):
I have one last comment :)

Sorry for providing my review in small chunks, instead of just reading and submitting all my comments at once. I am a bit distracted today trying to do too many tasks at once :)

src/window/windows/castlewindow_winapi.inc
HotariTobu: I didn't know that Windows10 allowed over 256 path lengths.
Therefore, I restored the path length limitation.

michaliskambi (approved):
Thank you!

HotariTobu:
Thank you for marging my commits.

PR #367, opened 2022-01-15
Add quotation marks to XML quote

eugeneloza:
Add also quotation marks as possible symbols in project name. See https://trello.com/c/6vmn9JKj/196-bug-project-name-caption-cannot-contain-special-characters


michaliskambi (approved):

michaliskambi:
Thank you!

PR #374, opened 2022-01-25
Castle tester (running tests also on Delphi, running tests on android with FPC)

and3md:
After the first stage of handling delphi according to our plan (https://github.com/castle-engine/castle-engine/pull/338), it was time for a less spectacular task - running tests on Delphi. Because we need a tool that works in FPC and Delphi. I created castle-tester that is (almost) compatible with fpcunit - needs only few ifdefs. Now we can run our tests with FPC, Delphi compilers.

## UI
It's common CGE app: 

obraz

Because this is a CGE project also we can run tests on mobile. I tested running castle-tester on android.

## Console
Running in console is available with `--console` parameter.  With additional `--suite` parameter only one test case can be tested:

`./castle-tester --console --suite=TTestCameras`

## fpcunit 
castle-tester is compatibile with fpcunit so you can still use `test_castle_game_engine.lpi` but `compile_console.sh` now builds castle-tester

## Fixes for Delphi
This PR also fixes found bugs in delphi.

## TODOs:
- selecting tests in UI
- tester for LCL/VCL things (TCastleTester class is ready to handle that, just another app style maybe by `--vlcl` option)


michaliskambi:
Thank you!

What I did now:

- Rebased on master.
- hidden `ButtonSelectTests` since it is not yet functional, show tests count using `LabelTestsCount`
- split the code clearly into 2 testing frameworks + (independent) testcases:

  ```
  code/tester-castle/ # new testing API using CastleTester
  code/tester-fpcunit/ # helpers for old testing API using FpcUnit
  code/testcases/ # testcases that work with both testing frameworks
  ```
- I rechecked and documented how to run both Castle and FpcUnit versions in README.md, extending your docs a bit.
- Fixed the line used by Emacs to run single testcases: `// -*- compile-command: "./test_single_testcase.sh TTestXXX" -*-`
- Fixed `FileNameContainsNonAsciiCharacters` to return `false` when argument is empty
- Filled missing AssertSameValue overload for Double
- I separated modes, as I wanted to make independent "what is current platform" from "whether we use TCastleWindow or console UI".
- Tests that are not possible to run should not `Fail` -- they should go without doing anything. E.g. on mobile, tests that create new TCastleWindow just are not executed, we don't fail them, as it would mean we cannot ever have correct tests on mobile.
- Uncommented all tests -- I saw that some testcases, and `TestInternalDetectClassPNG`, were not used when `CASTLE_TESTER`. But they actually execute fine for me.

Merging :)

michaliskambi (approved):

PR #380, opened 2022-02-14
Add {$warn 5027 off} to SaveToPascalCode

eugeneloza:
Avoid 'Note: Local variable "..." is assigned but never used' in images generated by TCastleImage.SaveToPascalCode

### DIFF
--- a/src/images/castleimages.pas
+++ b/src/images/castleimages.pas
@@ -2759,6 +2759,7 @@ begin
     'function ' +ImageName+ ': ' +ClassName+ ';' +nl + nl;
 
   CodeImplementation := CodeImplementation +
+    '{$warn 5027 off}' + NL + // To avoid 'Note: Local variable "..." is assigned but never used'
     'var' + NL +
     '  F' +ImageName+ ': ' +ClassName+ ';' + NL +
     'const' + NL +


michaliskambi:
Which compiler, which version emits such note?

I just tested with FPC 3.2.0 and FPC 3.2.2 and they don't report any note about this. And rightly so, there's no local variable there :), and variable in unit implementation `Fxxx` is used in a normal fashion in a function below, like

```pascal
function Title: TRGBAlphaImage;
begin
  if FTitle = nil then
  begin
    FTitle := ....
  end;
  Result := FTitle;
end;
```

Note that if we go with it, we'll have to make sure it compiles with Delphi too, Delphi has different warnings directives.

eugeneloza:
Free Pascal Compiler version 3.3.1-9915-g6dd0a1c078 [2022/01/14] for x86_64

2022-02-14 15_18_33-Greenshot

eugeneloza:
Indeed it might be a glitch of an unstable version (as the constant is actually used)

michaliskambi:
Ah, so this is FPC unstable, and it complains about `SplashPixels` that is used in 

```pascal
Move(SplashPixels, FSplash.RawPixels^, SizeOf(SplashPixels));
```

FPC unstable makes a number of other similar notes when compiling CGE. E.g. at castlestreamutils.pas ---

```
WriteBuffer(tmp, SizeOf(tmp));
```

also is not detected as "this is using `tmp` value". As a consequence we have a series of notes:


```
castlestreamutils.pas(209,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(217,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(225,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(233,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(241,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(249,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(257,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(265,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(273,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(281,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(289,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(297,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(305,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(313,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(321,3) Note: Local variable "tmp" is assigned but never used
castlestreamutils.pas(329,3) Note: Local variable "tmp" is assigned but never used
```

It seems that FPC unstable in general doesn't consider that routines using untyped parameters (like `Move`, `TStream.WriteBuffer`) actually use this parameter. I'm tempted to call this as bug in FPC unstable -- preferably it should be fixed in FPC. Otherwise we'll have to fix a few more places in CGE.

@eugeneloza Can you maybe submit it to FPC bugtracker? :) Maybe it can be fixed on FPC side.

PR #384, opened 2022-02-20
Ignores premultiplied alpha setting when loading Spine atlas

Kagamma:
This setting is useless for us (we control blending mode via `RenderOptions`), but causes exception because it's not yet handled.


michaliskambi:
Sure, thank you!

michaliskambi (approved):

PR #390, opened 2022-05-07
Use TFingerIndex in TCastleButton for consistency

eugeneloza:
Just noticed that `TCastleButton` uses `Integer` instead of `TFingerIndex`. No harm done, but for consistency... :)


michaliskambi (approved):
Of course, thank you!

PR #403, opened 2022-09-08
Convert two ISO-8858-1 characters to UTF-8

paulgevers:
The Debian linter (lintian) notes that two source files are not UTF-8. This PR fixes that.


michaliskambi (approved):
Thank you!

PR #417, opened 2022-11-21
Fixes for Chinese Input

Freedomax:
support Chinese Input


michaliskambi (changes requested):
Thank you for the pull request! Some comments:

1. Please post a more complete description / screenshot / recording how does the Chinese input fail for you, and what / why are you fixing with this PR.

    I was doing some experiment with it long time ago (resulting in https://github.com/castle-engine/castle-engine/blob/7690d9bb9837218a6005e41f57bb9c3e9bf7c55b/src/window/windows/castlewindow_winapi.inc#L512 ) and it seemed to work. (Though I do not understand Chinese, but I could see I can input Chinese characters.)

2. Please make all the comments in English. 

    Even if the target user group of this feature is people who understand Chinese (though I understand IME can be used in more contexts), we still use English everywhere so that everyone can understand this feature, and is able to test it and maintain.

Freedomax:
Thank you for your reply.
The results of processing only Handle_WM_CHAR are shown below.

Compile under FPC3.3.1:
WM_CHAR_FPC
The console output:
WM_Char_output_FPC

Compile under Delphi11:
WM_CHAR_Delphi
The console output:
WM_Char_output_Delphi

When compile under FPC,WParam is wrong.But when compile under Delphi,the edit has no output

Handle IME API,such as ImmSetCompositionWindow,IME window can be well placed.I haven't dealt with it yet.

Freedomax:
Sorry,commit on "castlecontrols_scrollview.inc" automatically merged by github desktop.
Should I close this PR?

michaliskambi:
> Sorry,commit on "castlecontrols_scrollview.inc" automatically merged by github desktop.

When you push some commit to a branch that was used to make a PR, it is automatically added to this PR.

To do unrelated work, you should create a new branch in your fork. I see it fixes an unrelated issue, and indeed it seems it should be fixed. (Though I'm hesitate whether we should synchronize `Scroll := Translation.Y` or `Translation.Y := Scroll` in this case :) )

No need to close this PR :)

- You can just push a commit that reverts it. 

- Or you can ignore it, I can skip this commit when merging, not a problem.

Thank you for the explanation and updates to the comments in code! Everything looks good. I'll test it and merge ~tomorrow.

Freedomax:
OK,Keep commit to "castlecontrols_scrollview.inc"  ,so there is no need to send another PR.

michaliskambi:
( Sorry for not addressing this yet, I'm coming down with a cold since last weekend, and I didn't have the time to actually test it on Windows yet. I'll get to this :) )

michaliskambi:
Sorry for the delay, I have a few big tasks to finish this week. And just yesterday we merged another big upgrade to Castle Game Engine physics (joints! documented briefly on https://castle-engine.io/physics ). I will get to this PR.

Freedomax:
> Sorry for the delay, I have a few big tasks to finish this week. And just yesterday we merged another big upgrade to Castle Game Engine physics (joints! documented briefly on https://castle-engine.io/physics ). I will get to this PR.

It's nothing, it's a small matter. I know you have many more important things to do.

Freedomax:
Sorry, just after I renamed the **master** branch, this PR was automatically closed.
Now I will wait for `TCastleEdit` to be perfect before dealing with `IME`.

michaliskambi:
Oh I don't want to lose this work because of an accident -- and I wanted to sit to review and merge it (and https://github.com/castle-engine/castle-engine/issues/453 ) later this night, and review. 

I see the related branch has been removed, which means PR can no longer exist and there's no branch from which to take the commit hash. But I could get the patch from GitHub (adding `.patch` to GitHub PR URL, https://github.com/castle-engine/castle-engine/pull/417.patch ) and apply this patch (`git apply 417.patch`). I committed it to CGE branch chinese-input-fixes-417 ( https://github.com/castle-engine/castle-engine/tree/chinese-input-fixes-417 ) and will look into it later.

michaliskambi:
Merged to master! Thank you and again sorry for such delays.

1. My resolution about the "by the way" modification done in this PR to `castlecontrols_scrollview.inc`:

    - I reproduced the case when user can desynchronize `Scroll` with `FScrollArea.Translation.Y`, so I agree with the intention to synchronize them.

        That said, doing `Scroll := FScrollArea.Translation.Y` in `TCastleScrollView.Loaded` was troubling me -- it meant we actually always discard `Scroll` saved in JSON.

        And it seems arbitrary to synchronize (A) `Scroll := FScrollArea.Translation.Y` vs (B) `FScrollArea.Translation.Y := Scroll`. Your choice of A was indeed better (because CGE editor allows you to move UI without changing `Scroll`), but in the perfect world we shouldn't have to make such choice :)

    - So I thought that we shouldn't really synchronize 2 values. TCastleScrollView.Scroll property should just get/set the `FScrollArea.Translation`, without introducing any extra field that has be synchronized. Done -- this works really nice, it avoids the whole problem instead of solving it :)

2. About the IME changes:

    - I familiarized myself with the related WinAPI stuff ( https://learn.microsoft.com/en-us/windows/win32/intl/wm-ime-startcomposition , https://learn.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immgetcontext etc. ), added temporary logs when WM_IME_START/ENDCOMPOSITION occur, to understand what's happening. Overall I feel I understand the code, and it looks good.

    - I did small formatting improvements (we prefer to not write `self.`, we write `false` and `true` lowercase, we put spaces around operators `:=`).

    - I tested it (added Chinese input method to Windows 10, invoked the Windows IME window on TCastleEdit in `examples/fonts/test_local_characters/CastleEngineManifest.xml`, typed random stuff).

I'm merging this, with questions / TODO for you to follow up:

1. Handle_IME sets `Result := true` only in case of `WM_IME_COMPOSITION`. So `WM_IME_STARTCOMPOSITION, WM_IME_ENDCOMPOSITION`, while they do some work, they also allow `DefWindowProcW` to run.

   Is this correct? If yes -> please document at WM_IME_STARTCOMPOSITION, WM_IME_ENDCOMPOSITION handling.

   If not -> add `Result := true` there.

   In case of `WM_IME_COMPOSITION` you set `Result := true` which seems natural.

2. Why do we handle `WM_IME_ENDCOMPOSITION`? Isn't this supposed to hide the IME window? At least in my tests, it seems `WM_IME_ENDCOMPOSITION` only occurs when window disappears. Do we need to do anything in reaction to it? It seems the current code will reposition the window that is just being hidden.

3. I didn't test whether characters I input actually match what CGE tries to render. I rely on you to confirm it :) You will have to define a font with Chinese characters.

    See `examples/fonts/test_local_characters` that does it quite manually, by defining characters in `mk_font_units.sh` script.

    The more modern way is to define TCastleFont component in a .castle-user-interface file, and set there `LoadBasicCharacters`. So you

    - point the `TCastleFont.URL` to a font with Chinese characters,
    - put the relevant Chinese characters in `TCastleFont.LoadCharacters`
    - set this font as `TCastleEdit.CustomFont`

    ... and then the edit should display Chinese characters properly. Yes, I should update `examples/fonts/test_local_characters` to show this approach properly, that's my TODO for tomorrow :)

4. I understand the TODO "should get focused edit's position, not done" -- I see now you just take cursor position.

    If you want to finish it, you can look at `Container.Focus`. If it's non-empty, and last item is `TCastleEdit`, then the `LocalToContainerPosition` allows to get the position (in final device pixels by default).

    So I guess you can extend

    ```pascal
    GetCursorPos(pt);
    ScreenToClient(h_Wnd, pt);
    ```

    with something like this:

    ```
    if (Container.Focus.Count <> 0) and
       (Container.Focus.Last is TCastleEdit) then
      Pt := Container.Focus.LocalToContainerPosition(TVector2.Zero)
    else
    begin
      GetCursorPos(Pt);
      ScreenToClient(h_Wnd, Pt);
    end;
    ```

I will leave it for you to finish in following PRs, as you're evidently more familiar with IME and can really test it :) Hopefully my comments help.

Freedomax:
> Handle_IME sets Result := true only in case of WM_IME_COMPOSITION. So WM_IME_STARTCOMPOSITION, WM_IME_ENDCOMPOSITION, while they do some work, they also allow DefWindowProcW to run.
Is this correct? If yes -> please document at WM_IME_STARTCOMPOSITION, WM_IME_ENDCOMPOSITION handling.
If not -> add Result := true there.
In case of WM_IME_COMPOSITION you set Result := true which seems natural.

Handle `WM_IME_STARTCOMPOSITION, WM_IME_ENDCOMPOSITION`, must return false. Because CGE does not implement the IME input window by itself, return false to let the system IME handle it by itself. We only tell the system where the IME window needs to be displayed, font information, etc.

> Why do we handle WM_IME_ENDCOMPOSITION? Isn't this supposed to hide the IME window? At least in my tests, it seems WM_IME_ENDCOMPOSITION only occurs when window disappears. Do we need to do anything in reaction to it? It seems the current code will reposition the window that is just being hidden.

Usually the input of Chinese is one word or several words at a time. The IME window will be opened, and then closed multiple times. Since the IME can be developed by any company, sometimes the IME window will appear prematurely at the last closed position. So deal with `WM_IME_ENDCOMPOSITION` to avoid This case. Also, there's no harm in doing it.

> If you want to finish it, you can look at Container.Focus. If it's non-empty, and last item is TCastleEdit, then the LocalToContainerPosition allows to get the position (in final device pixels by default).

The most perfect way is to get the position of the caret, not just the position of the edit. I will check whether `TCastleEdit` has such a function later.

PR #430, opened 2022-12-27
New Flag for WMEraseBkgnd to avoid flicker

ProHolz:
published the VCL Mouse Events

### DIFF
--- a/src/delphi/vcl.castlecontrol.pas
+++ b/src/delphi/vcl.castlecontrol.pas
@@ -68,7 +68,6 @@ type
       to update Pressed when needed. }
     procedure UpdateShiftState(const Shift: TShiftState);
   private
-      FEraseBackGround : Boolean;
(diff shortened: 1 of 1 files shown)


michaliskambi (changes requested):
Thank you! I had in TODO to deal with this flicker. Indeed handling WM_ERASEBKGND is what is needed, our TCastleWindow on WinAPI also does it.

I would request one change: let's remove the `EraseBackGround` property and field, and just do it always. There doesn't seem to be much point in making this (ignoring WM_ERASEBKGND) be optional and so we don't want to support the option of `EraseBackGround = true`.

I.e. `WMEraseBkgnd` should just unconditionally do 

```pascal
procedure TCastleControl.WMEraseBkgnd(var Message: TWMEraseBkgnd);
begin
  Message.Result := 1;
end;
```

ProHolz:
For me this will work, I can change it no problem

ProHolz:
property EraseBackGround removed

michaliskambi (approved):
Thank you!

PR #432, opened 2022-12-29
Enable Quick Test GH Action running on Windows

fabiang:
GH Action for Quick Test is now working on Windows. If you don't defined `shell: pwsh` the bash script is executed with gitbash, where `%PATH%` is different on Windows machines.

### DIFF
--- a/.github/workflows/quick-test.yml
+++ b/.github/workflows/quick-test.yml
@@ -19,14 +19,10 @@ jobs:
     name: Quick Test of Tools
     strategy:
       matrix:
-        # TODO: Not including windows-latest, as it fails:
-        # ./tools/build-tool/castle-engine_compile.sh: line 43: fpc: command not found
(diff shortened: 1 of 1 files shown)


michaliskambi:
Oh, cool, so I cal request PowerShell on Windows like this. 

I like this solution, it is also consistent with GitLab CI that strongly recommends to use PowerShell in case of Windows. (I have yet to make a news post with my quick "PowerShell cheatsheet for bash programmers" that I wrote for myself :) ).

I'm approving the workflow to run, let's see what GH Actions answer for this PR.

( Note that our CGE workflows fail since 2 days because of FPC (3.3.1) upgrade in Docker, I'm investigating it, so some failures are to be expected and not related to this PR work. )

michaliskambi (approved):
Thank you, and GH Actions confirm "Quick Test" it now good on both Windows and Linux.

PR #435, opened 2023-01-17
New color picker window

and3md:
This PR implement new color picker similar to blender color popup. It shows changed values immediately, not after clicking OK.
Taking mouse pointer far from color picker window closes it automatically. 

obraz

obraz
 
obraz

obraz

obraz

It uses mbColorLib package and there is still a problem in running our tests - I think something more is needed to get them working. 

Maybe `TCastlePropertyEditorHook` should be moved to another file but it is currently related only color picker window so I left it in castlepropedits_color.inc.


michaliskambi (approved):
I noticed the need for some changes, and made them, to apply quickly :)

After some deliberation, I also reverted my decision about using `mbcolorliblaz.lpk`. Instead I created `castle_editor_components.lpk`, that includes mbColor components, and some CGE stuff (hook and property editors, forms using mbColor). For 2 reasons:

- This makes castle_components.lpk cleaner. We don't expose to users units like `formcastlecolorpicker.pas` that they are not supposed to be using. Internal stuff is neatly in `castle_editor_components.lpk`
- By having own package `castle_editor_components.lpk`, we hopefully avoid bugreports from people who manually install LPK. Although I also editor packages/README.md to discourage this practice, and encourage more using our "Register Lazarus Packages" button.

I also changed the decision about package.

PR #437, opened 2023-01-19
TPiecewiseCubicBezier-Change: Change the scope of the UpdateBezierCur…

KumurTash:
Change the scope of the UpdateBezierCurves method in TPiecewiseCubicBezier.
This will allow classes inherited from TPiecewiseCubicBezier to change the way additional points are calculated.


michaliskambi (approved):
Makes sense, applying -- thank you!

I did a small commit improving whitespace to follow our conventions :)

PR #438, opened 2023-01-24
Avoid crash when TObjectGroupLayer is empty

Freedomax:
`TObjectGroupLayer.Objects` is created in `TObjectGroupLayer.Load`, when TObjectGroupLayer is empty(very likely), `objects` will not be created ,Calling `TCastleTiledMapConverter.BuildObjectGroupLayerNode` the program will crash.
```
procedure TCastleTiledMapData.TObjectGroupLayer.Load(const Element: TDOMElement; const BaseUrl: String);
...
begin 
 ... 
  try
    while I.GetNext do
    begin
      if  LowerCase(I.Current.TagName) = 'object' then
      begin
        NewObject := TTiledObject.Create;
        NewObject.Load(I.Current, BaseUrl);
        if not Assigned(Objects) then
          Objects := TTiledObjectList.Create;
        Objects.Add(NewObject);
      end;
    end;
  finally FreeAndNil(I) end;
end;
```

### DIFF
--- a/src/scene/load/x3dloadinternaltiledmap.pas
+++ b/src/scene/load/x3dloadinternaltiledmap.pas
@@ -233,7 +233,7 @@ var
   AVector2List: TVector2List;
   I: Cardinal;
 begin
-  if not Assigned(ALayer as TCastleTiledMapData.TObjectGroupLayer).Objects then Exit;
+  if not Assigned((ALayer as TCastleTiledMapData.TObjectGroupLayer).Objects) then Exit;
 
   TiledObjectMaterial := nil;
   TiledObjectNode := nil;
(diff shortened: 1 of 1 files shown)


michaliskambi:
Thank you for the report!

I reproduced the problem easily thanks to your description, and added a test map to reproduce it. It is now part of automatic tests so will be automatically checked to make sure this bug doesn't return (test map in `tests/data/tiled-maps/desert_with_empty_objects_layer.tmx`, test code using it in `tests/code/testcases/testcastletiledmap.pas`).

I solved the actual crash a bit differently -- instead of creating `Objects` on-demand, I decided it is simpler to just create it in constructor, and then code accessing it can reliably assume it is never `nil`. I would say that creating it on-demand was in that case a premature optimization, lowering code quality (and leading to bugs) without any real benefits (I can't imagine a real-life application where this extra creation of `Objects` even for empty object layers would be any real performance/memory problem). So it falls under my guideline to avoid premature optimizations ( https://castle-engine.io/coding_conventions#_optimize_smartly_profile_optimize_where_it_matters_and_not_where_it_doesnt_think_about_smarter_algorithms_and_moving_cpu_work_to_gpu_to_get_big_benefits ).

PR #443, opened 2023-01-31
Treating window DPI correctly on macOS

janadamec1:
This fixes the issue in view3dscene running on Retina screen on Mac (https://github.com/castle-engine/view3dscene/issues/47). Basically mouse event coordinates were not scaled.

After some discussion we decided to change the view3dscene UI scaling to respect DPI and move Dpi property from TCastleWindow to TCastleContainer.

I will open the PR in view3dscene once this is merged - although the only change is one line with setting the UIScaling.


michaliskambi (approved):
Sorry for such long delay in review! Reviewed it now, perfect, thank you -- merging.

PR #445, opened 2023-02-07
Making TCastleTiledMap support rendering animations.

Freedomax:
I tested a tiledmap with about 100 animations ,FPS has not decreased,and found an animation that is out of sync with rotation applied, the reason is unknown.


Freedomax:
> I tested a tiledmap with about 100 animations ,FPS has not decreased,and found an animation that is out of sync with rotation applied, the reason is unknown.
Time is not synchronized due to loading time delay, and further calculations and settings are needed.

Freedomax:
The implementation of `TexCoordInterp` has issues, so turn it off first.

Freedomax:
The critical issue has been fixed, I tested it using a large map and the result is perfect.

Freedomax:
Hi,@Michalis,Based on your previous substantial optimization, I further optimized TCastleTiledMap using TDictionary ;). The number of created shapes now equals the number of tile sets used by the layers (excluding unused tile sets), and is completely independent of the arrangement of tiles in the map. For example, if two tile sets are used, only two shapes will be created.

Freedomax:
The original **desert_big.tmx** is no longer sufficient, so I created a pathological map.
Comparison picture before and after this PR :
test

The map :
[desert_big_twotilesets.zip](https://github.com/castle-engine/castle-engine/files/10682117/desert_big_twotilesets.zip)

Freedomax:
The optimization of tiles without animations has been completed, and now the optimization of TimeSensor and Animation will be carried out, with the following optimization targets and key values: 

TimeSensor : `CycleTime`
 Animation : `tileset` and `tileid`.

michaliskambi:
Fantastic you got the Tiled animations working, I promised to do it myself in forum thread https://forum.castle-engine.io/t/tcastletiledmap-cannot-render-animations/728 , I'm happy you tackled it first :)

1. As for playing animations, as I mentioned on https://discord.com/channels/389676745957310465/882243371584864287/1072853451085725828 , I'd say we can do it simpler (just call `Start` on the `TTimeSensorNode` instances that we want) and in `TCastleTiledMap` code (so no need for `TCastleSceneCore.PlayAllAnimations`). So instead of `TCastleSceneCore.PlayAllAnimations`, please add instead something like

    ```delphi
    procedure TCastleTiledMap.PlayAnimations;
    var
      I: Integer;
      TimeSensor: TTimeSensorNode;
    begin
      for I := 0 to FScene.AnimationsList.Count - 1 do
      begin
        TimeSensor := FScene.AnimationsList.Objects[I] as TTimeSensorNode;
        TimeSensor.Start;
      end;
    end;
    ```

    and then you can call it automatically to play animations once map is loaded, so from `TCastleTiledMap.ReloadMap`. Oh, right where you call `FScene.PlayAllAnimations` now :)

    As a bonus, I would introduce a property `AutoPlayAnimations: Boolean`, default `true`. When set to `false`, user would have to manually call `PlayAnimations`. So the call in `TCastleTiledMap.ReloadMap` would actually just do `if AutoPlayAnimations then PlayAnimations`.

2. And then, as later possible extensions, we could define more stuff based on this. Like playing only animations on a single layer, like

    ```delphi
    procedure TCastleTiledMap.PlayLayerAnimations(const LayerName: String);
    begin
      TimeSensor := FScene.AnimationTimeSensor('TimeSensor_' + LayerName);
      if TimeSensor <> nil then
        TimeSensor.Start
      else
        WritelnWarning('No layer "%s" on which we can play animations', [LayerName]);
    end;
    ```

    and/or a corresponding `StopAnimations(const ResetAnimations: Boolean)` that would call `Stop` on all time sensors, and it `ResetAnimations` even "reset" the animation state to initial (this should be doable by doing `TimeSensor.FakeTime(0, false, true)`.

    I am not yet sure if `PlayLayerAnimations` and/or `StopAnimations` are useful.

    For now, I would say to ignore the `PlayLayerAnimations` idea, but add the `StopAnimations(const ResetAnimations: Boolean)` that I described.

    And add a button to `examples/tiled/map_viewer_in_viewport/` demo to "Play Animations" , "Stop Animations" that calls the respective methods, to test them.

2. As for rendering with dictionary: Yeah, this makes perfect sense as an extension of my previous "static batching".

    Note that it will break the rendering order, assumed by some Tiled maps (e.g. `examples/tiled/map_viewer_in_viewport/data/maps/orthogonal-outside.tmx`). Though actually it was kind of broken already, e.g. `perspective_walls.tmx` in that directory fails now anyway. I added `ExperimentalAssumePerfectRenderingOrder` to play with it, but it is a hack -- it will make rendering of Tiled always OK, but it breaks depth when rendering with other objects. After your modifications, `ExperimentalAssumePerfectRenderingOrder` will not have sense anymore, but that's maybe good, it wasn't a future-proof solution anyway. I suppose the only right solution is to increase Z *for each tile*, to really guarantee proper order. But it means Z will grow to quite large values for huge maps -- well, but in 2D, maybe it should. 
    
    I want to think/experiment with how to get perfect rendering order, though for now I guess "increase Z for each tile" seems like a straightforward solution. And then your dictionary makes perfect sense, and my `ExperimentalAssumePerfectRenderingOrder` can be just removed as not really useful approach.

3. More review will come, once I get my head rested, yesterday I have finally merged a big work (shadow-volumes-new with lots of renderer upgrades) and should have finally time to review the pending PRs.

Freedomax:
I'm sorry, I am unable to complete the task of "using only one TimeSensor for each layer" as the animation time for each tile may be different and I have not found a solution. If you have any methods, please let me know. I can only use one TimeSensor for animations with the same loop time now.

Freedomax:
OK, I'll try `TimeSensor.EventTime`, seem like is a solution.

michaliskambi:
> I'm sorry, I am unable to complete the task of "using only one TimeSensor for each layer" as the animation time for each tile may be different and I have not found a solution. If you have any methods, please let me know. I can only use one TimeSensor for animations with the same loop time now.

Ah, we don't need it. I mean we can absolutely have multiple time sensors per one layer. There is no need to squeeze them into one.

The `TCastleTiledMap.PlayLayerAnimations` -- for now I'd say we don't need this method at all, and in future (if it will be useful) we can always just scan the time sensors, and e.g. run all time sensors whose names start with given layer. E.g. it can do

```
procedure TCastleTiledMap.PlayLayerAnimations(const LayerName: String);
var
  I: Integer;
  TimeSensor: TTimeSensorNode;
begin
  for I := 0 to FScene.AnimationsList.Count - 1 do
  begin
    TimeSensor := FScene.AnimationsList.Objects[I] as TTimeSensorNode;
    TimeSensor.Start;
  end;
end;
begin
  TimeSensor := FScene.AnimationTimeSensor();
  if TimeSensor <> nil then
    TimeSensor.Start
  else
    WritelnWarning('No layer "%s" on which we can play animations', [LayerName]);
end;

**Comment by Freedomax** [comment]
All optimizations are completed.

1. TimeSensor
`TTimeSensorContext = TDictionary<Cardinal, TTimeSensorNode>`

Due to the property `CycleInterval`, the number of creations is equal to the number of categories of animation cycle time in the map. If there is only one cycle time, then only one `TTimeSensorNode` will be created.



2. Animations
`TAnimationContext = TDictionary<TCastleTiledMapData.TAnimation, TAnimationNodes>`

It is a tile on a certain tile set, so the number of shapes created is equal to the number of tiles with animations in the tile set. All the same animations use the same `TCoordinateInterpolator2DNode` (obviously also triggered by the same `TTimeSensorNode`) to update in batch.
If there is only one tile with animation, then only one will be created.


After testing, this update greatly improves the animation efficiency. The test map CPU usage decreased from 2% to 1%, and GPU usage also decreased by half.

**Comment by Freedomax** [comment]
Regarding `TCastleTiledMap.PlayAnimations`, I haven't tested it successfully yet.

**Comment by Freedomax** [comment]
Adding `FScene.ProcessEvents := True;` fixed it. All tasks are basically completed, and some useful functions can be added in the future.

**Comment by Freedomax** [comment]
I just discovered a flaw. Since I processed all animations of the same tile as a batch, I neglected the fact that tiles can be flipped/rotated and these were ignored. I used the simplest method, redefining :
```
TAnimationWithFlips = record
  Animation : TCastleTiledMapData.TAnimation;
  HorizontalFlip, VerticalFlip, DiagonalFlip: Boolean;
end;

TAnimationContext = TDictionary<TAnimationWithFlips,TAnimationNodes>; 
```
so that these animations are considered a new animation. I also considered other methods but they were too complex and unnecessary. The current method is sufficient as these new animations can also be batch rendered.

**Comment by Freedomax** [comment]
I added a `CurrentZ` to procedure `RenderTile`, increasing by 0.000001 per tile to ensure the correct rendering order, even if 1000 x 1000, the Z will only increase by 1.

desert2tilesets tmx
perspective_walls tmx

File: src/scene/castlescenecore.pas [hunk 1]
```diff
@@ -437,6 +437,9 @@   TPlayAnimationParameters = class
     if the scene does not change (in a significant way) between each query.
     So you can query information like @link(LocalBoundingBox),
     @link(BoundingBox), @link(VerticesCount)... as often as you want to. }
+
+  { TCastleSceneCore }
```
michaliskambi, line 441: Do not add such comments with just class name. They are a bit useless (a person reading doc, or source code, will see the class name anyway), and would obscure the good documentation string for PasDoc. Lazarus adds them automatically, I recommend to disable it in Lazarus settings, see the screenshot in https://github.com/pasdoc/pasdoc/issues/149#issuecomment-1308837496 .

src/scene/x3d/x3dnodes_standard_time.inc
```diff
@@ -1,3 +1,4 @@
+{%MainUnit X3DNodes.pas}
```
michaliskambi, line 1: Specify filenames lowercase here, as they are strict lowercase on Unix-like systems, so please make this

```
{%MainUnit x3dnodes.pas}
```

src/scene/load/x3dloadinternaltiledmap.pas
```diff
@@ -48,6 +48,39 @@ function LoadTiledMap2d(const Stream: TStream; const BaseUrl: String): TX3DRootN
   { Convert Tiled map into X3D node. }
   TCastleTiledMapConverter = class
   strict private
+    type
```
michaliskambi, line 51: Indent types within the `type` section by 2 additional spaces, also like in https://castle-engine.io/coding_conventions#_indenting_inside_classes .

src/scene/load/x3dloadinternaltiledmap.pas
```diff
@@ -48,6 +48,39 @@ function LoadTiledMap2d(const Stream: TStream; const BaseUrl: String): TX3DRootN
...
+      Animation : TCastleTiledMapData.TAnimation;
+      HorizontalFlip, VerticalFlip, DiagonalFlip: Boolean;
+    end;
+
+    { CycleInterval in milliseconds. }
+    TTimeSensorContext = {$ifdef FPC}specialize{$endif} TDictionary<Cardinal,TTimeSensorNode>;
+
+    TAnimationContext = {$ifdef FPC}specialize{$endif} TDictionary<TAnimationWithFlips,TAnimationNodes>;
+
+    TRenderContext =class ({$ifdef FPC}specialize{$endif} TDictionary<TCastleTiledMapData.TTileset,TTilesetNodes>)
```
michaliskambi, line 72: We already have a class `TRenderContext` in CGE (in `CastleRenderContext`), so this one should be renamed, to avoid confusion :) 

In general, the word "context" in CGE is often associated with "OpenGL(ES) rendering context", it's better not to use it for something else.

I would propose: `TLayerConversion`. As the purpose of it is to encapsulate information local to `TCastleTiledMapConverter.BuildTileLayerNode` call.

So

- TRenderContext -> TLayerConversion
- TTimeSensorContext -> TLayerTimeSensors
- TAnimationContext -> TLayerAnimations

michaliskambi (changes requested):
Thank you very much for doing this, again. And sorry for doing my review in "phases", too much work lately.

1. I made some minor comments about indentation and naming below.

2. Aside from this, the high-level logic looks good, but I'd like to test it more on actual Tiled maps.

    So the additional request is: can you provide sample some sample Tiled maps using animations, to test various scenarios you account for? They don't have to be pretty :) though you can use any existing open-source assets (like from https://opengameart.org/ ).

    Commit them to `examples/tiled/map_viewer_in_viewport/data/` so that everyone can easily open them. Please also commit `desert_big_twotilesets.zip` there.

src/scene/load/x3dloadinternaltiledmap.pas
Freedomax: I dropped the idea because that would make the code too complicated.

Freedomax:
After your 5953e5705ea98ed009f42cdce5379757496bcbe5 submission the desert_big.tmx FPS dropped to 50 ,I don t know why.

Freedomax:
Thank you for your suggestions. I'll take the time to make a map with animations later.

Freedomax:
Two maps were created. The first is a small map which primarily showcases the rotation and flip features. The second is a large map with over 10,000 animations including snowfall and flowing water (rivers only, oceans are still static).
The pngs are all from opengameart.org, and the copyright information is annotated in data/maps/animations/AUTHORS.txt.

Freedomax:
> After your [5953e57](https://github.com/castle-engine/castle-engine/commit/5953e5705ea98ed009f42cdce5379757496bcbe5) submission the desert_big.tmx FPS dropped to 50 ,I don t know why.

Maybe the update of `shadow-volumes-new` has optimized the rendering of X3D, by eliminating Nodes outside the view. This explains why the FPS has decreased now, because I only created one Node that occupied the entire map, resulting in a negative optimization. :) The solution is to create a new Node when one exceeds a certain size, which will match the update of `shadow-volumes-new`.

michaliskambi:
> After your [5953e57](https://github.com/castle-engine/castle-engine/commit/5953e5705ea98ed009f42cdce5379757496bcbe5) submission the desert_big.tmx FPS dropped to 50 ,I don t know why.
> 
> Maybe the update of shadow-volumes-new has optimized the rendering of X3D, by eliminating Nodes outside the view. This explains why the FPS has decreased now, because I only created one Node that occupied the entire map, resulting in a negative optimization. :) The solution is to create a new Node when one exceeds a certain size, which will match the update of shadow-volumes-new.

But no such change happened in `shadow-volumes-new` :) Before and after `shadow-volumes-new`, we had algorithm doing frustum culling, per shape, in TCastleScene. Nothing in this regard should have changed.

I'll investigate why the slowness occurred. To be clear, can you point the exact moments (commits here) where you experienced larger speed and slower speed? I see you did a few synchronizations with master branch, do you know for sure when FPS drop occurred?

Freedomax:
> > After your [5953e57](https://github.com/castle-engine/castle-engine/commit/5953e5705ea98ed009f42cdce5379757496bcbe5) submission the desert_big.tmx FPS dropped to 50 ,I don t know why.
> > Maybe the update of shadow-volumes-new has optimized the rendering of X3D, by eliminating Nodes outside the view. This explains why the FPS has decreased now, because I only created one Node that occupied the entire map, resulting in a negative optimization. :) The solution is to create a new Node when one exceeds a certain size, which will match the update of shadow-volumes-new.
> 
> But no such change happened in `shadow-volumes-new` :) Before and after `shadow-volumes-new`, we had algorithm doing frustum culling, per shape, in TCastleScene. Nothing in this regard should have changed.
> 
> I'll investigate why the slowness occurred. To be clear, can you point the exact moments (commits here) where you experienced larger speed and slower speed? I see you did a few synchronizations with master branch, do you know for sure when FPS drop occurred?

This issue is not related to my PR. You can test it with the current **master** branch and you will find that the FPS of "desert_big_twotilesets.tmx" is higher than "desert_big.tmx" (with default settings and not scaled). The file of "desert_big_twotilesets.tmx" can be downloaded in my previous reply.

Freedomax:
I just partitioned the map with a 2000*2000 pixel grid, creating a new node across regions, and now the FPS is back to 60, just as I suspected.

Freedomax:
Founded, 513a2cf1d1f04f3479013e14af7d99791560194b.
After this commit,FPS goes down.

michaliskambi:
> Founded, https://github.com/castle-engine/castle-engine/commit/513a2cf1d1f04f3479013e14af7d99791560194b.
> After this commit,FPS goes down.

Thank you! Fantastic, thanks for drilling down and finding this. 

Yeah, I know exactly why FPS drops there, and it is my fault :), this change (from `shadow-volumes-new`) indeed makes some cases of rendering much slower -- in particular it will hurt the situation when you have a `TQuadSetNode` (non-indexed) with lots of quads. And this is *exactly* the situation that occurs when generating Tiled map, it is exactly using big TQuadSetNode.

And the good news -- I can and should fix it. That is, bring back `TQuadSetNode` speed as it was. We need to convert `TQuadSetNode` into 1 primitive with lots of triangles. (Instead, that old code, that just got activated, would use a lot of triangle fan primitives.)

It is excellent that you found this, and I'll definitely will want to fix it before merging this PR -- so please don't do any changes to address it on Tiled side, this problem should be fixed deeper in CGE rendering code, I'll do it :)

Also, this meant that Tiled rendering on OpenGLES, before `shadow-volumes-new`, was already quite slow. Because OpenGLES was since ~always using that unoptimal approach. So my fix will

- bring FPS on desktop (OpenGL) back to what it was
- make FPS on mobile (OpenGLES) much better

Freedomax:
A new map.
beach

michaliskambi:
@Freedomax The speed regression I introduced in https://github.com/castle-engine/castle-engine/commit/513a2cf1d1f04f3479013e14af7d99791560194b is now fixed. Please merge CGE master to this PR, you should see speed of everything back to normal.

And as a bonus, it is all efficient on mobile (OpenGLES) too, which was slow before, for a long time.

I also added a statistics showing "Draw Calls" to view3dscene, and CGE editor (press F8 to see FPS, draw calls and other info). This way such regression would be easily visible in the statistics.

Admittedly my current system is a bit too fast, all Tiled maps are crazy fast, before and after :) But at least I can look at "Draw calls", now simple maps have again 1 draw call, as expected, not thousands.

Freedomax:
> I just discovered a flaw. Since I processed all animations of the same tile as a batch, I neglected the fact that tiles can be flipped/rotated and these were ignored. I used the simplest method, redefining :
> 
> ```
> TAnimationWithFlips = record
>   Animation : TCastleTiledMapData.TAnimation;
>   HorizontalFlip, VerticalFlip, DiagonalFlip: Boolean;
> end;
> 
> TAnimationContext = TDictionary<TAnimationWithFlips,TAnimationNodes>; 
> ```
> 
> so that these animations are considered a new animation. I also considered other methods but they were too complex and unnecessary. The current method is sufficient as these new animations can also be batch rendered.

I implemented it in the same way as before.

`TLayerAnimations = TDictionary<TCastleTiledMapData.TAnimation,TAnimationNodes>;`

 Regardless of flipping/rotating, the same animations are currently on the same node. After testing, the efficiency of "snowworld_big.tmx" increased by about 12%, but it took a little longer to load.

Freedomax:
The `SetURL` of `TCastleTiledMap` may trigger an exception in the design mode and the editor cannot open the design file. This issue also exists in the master branch. It is fixed in [01574b7](https://github.com/castle-engine/castle-engine/pull/445/commits/01574b72645c9cfda44b60a9eeef65ff234b48e6).

michaliskambi (approved):
Thank you! I did some small changes:

- minor code style stuff (lowercase `true`, `false` etc.)
- tested `FixTileset`, decided to keep (nice idea, alternative to `SmoothScalingSafeBorder`), but added more docs about it and renamed to `ForceTilesetSpacing`
- remade `AnimationsCount` -> boolean `HasAnimations` (as we don't expose any API to actually access animations by indexes from TCastleTiledMap)
- I also removed my `ExperimentalAssumePerfectRenderingOrder`; it was ugly and is thankfully no longer needed after your changes).

Thank you again for great additions to Tiled! I plan to make a news post about it soon.

michaliskambi:
I have also extended https://castle-engine.io/tiled_maps to mention support for animations and `ForceTilesetSpacing`.

Thanks to this PR, we have 4 things:

- Animations (all animations play automatically, you can also explicitly start and stop them)
- Fixed seams at certain zooms
- Added ForceTilesetSpacing (useful to avoid artifacts with SmoothScaling=true)
- Proper Z-order of tiles in all situations

I will make a bigger news post about `TCastleTiledMap` and https://castle-engine.io/tiled_maps, including the features done with this PR, around this weekend. My last news post about Tiled only mentioned support for Tiled in TCastleScene ( https://castle-engine.io/wp/2023/01/22/big-improvements-and-optimizations-for-tiled-maps-handling-in-tcastlescene/ ), we have something *much* better now in `TCastleTiledMap`, thanks also to your work.

Note: I absolutely want to give you credit, in the news post, in https://castle-engine.io/credits etc. How should I call you? Just _Freedomax_ (GitHub handle)? Or do you want me to use a real first + last name, or some other nick?

Freedomax:
> Note: I absolutely want to give you credit, in the news post, in https://castle-engine.io/credits etc. How should I call you? Just Freedomax (GitHub handle)? Or do you want me to use a real first + last name, or some other nick?

@michaliskambi ,about the credit: using Freedomax (GitHub handle) is sufficient. I don't want to use my real name, and I use very casual names on different online platforms because I don't want others to be able to find me through social engineering. (Because there may be danger associated with a particular statement I make.)

About `ForceTilesetSpacing`,I think in the future a URL can be generated for each repaired Tileset, and then a cache can be used so that the map switch does not need to regenerate the Tileset. Then the user can choose to delete these caches at some point to avoid memory consumption.

Issue #450, opened 2023-02-17
About the URL property editor


Freedomax:
The default folder of "Browse" should point to the folder of the current `URL` (if the `URL` folder does not exist, it should point to the project "data" directory), not determined by the operating system.
The current method will waste some time virtually.

Freedomax:
This is a small issue, I created a PR https://github.com/castle-engine/castle-engine/pull/451.

PR #451, opened 2023-02-17
Set InitialDir when TCastleOpen3DDialog execute

Freedomax:
see #450


michaliskambi (approved):
Makes sense. 

We also explicitly adjust `InitialDir` of dialogs in a few places, as default value of it is not very useful.

And in `castlewindow_lcl.inc` we even do

```
      Dialog.InitialDir := SysToUTF8(ExtractFilePath(ExpandedFileName));
      Dialog.FileName := SysToUTF8(ExtractFileName(ExpandedFileName));
```

... so, what you propose in this PR is consistent. We should adjust `InitialDir`. Merging, thank you!

PR #455, opened 2023-02-22
Custom Scrollbar

Freedomax:
https://user-images.githubusercontent.com/26040454/220613561-7f506dda-60c0-4c2e-b37c-413d5d26dcd8.mp4

Although TCastleTheme can also set the image of Scrollbar, it needs to use code, which is a bit troublesome.
With this PR, You can easily customize Scrollbar, `SpaceForButton` is used to leave space for the upper and lower buttons, and you can add button sub-controls yourself.
`DrawFrameUnderButton` determines whether to draw the Frame under the button. If the Frame has a border, it is recommended to set it to false, and if the frame is flat, it is recommended to set it to True. Both can be customized.


michaliskambi:
Cool, this is indeed something I wanted to do too -- i.e. expose all images to be configurable per-control on each TCastleUserInterface.

( And in the more distant future, I'd like TCastleTheme to also be configurable from editor, but this can wait. )

Notes:

1. About the preferred API: I'd like to keep it as simple as possible, so

    - without a boolean property like `CustomScrollbar`
    - without `Custom` prefix for image subcomponents like `CustomScrollbarFrame` and `CustomScrollbarSlider`

    I know our TCastleButton has such `CustomXxx` properties, and `CustomBackground` boolean, and I'm unhappy about it :) I will at some point want to improve that API (in a backward-compatible way of course, so old properties would still be read from JSON etc.). See https://github.com/castle-engine/castle-engine/blob/master/src/ui/castlecontrols_button.inc

    Our TCastleAbstractSlider may serve as a better example here, https://github.com/castle-engine/castle-engine/blob/master/src/ui/castlecontrols_sliders.inc . It exposes images like this:

    ```pascal
    { Background image that fills the entire slider area. }
    property Background: TCastleImagePersistent read FBackground;

    { Thumb image indicates the current value,
      like (like @link(TCastleFloatSlider.Value) or @link(TCastleIntegerSlider.Value)). }
    property Thumb: TCastleImagePersistent read FThumb;
    ```

    and there is no boolean property to activate using these images. Instead, the code only checks if they are `Empty`, and draws like

    ```pascal
    procedure TCastleAbstractSlider.Render;
    var
      BackgroundFinal: TCastleImagePersistent;
    begin
      inherited;

      if FBackground.Empty then
        BackgroundFinal := Theme.ImagesPersistent[tiSlider]
      else
        BackgroundFinal := FBackground;

      BackgroundFinal.DrawUiBegin(UIScale);
      BackgroundFinal.Draw(RenderRect);
      BackgroundFinal.DrawUiEnd;
    end;
    ```

    This also makes implementation more straightforward -- the code just chooses one or another TCastleImagePersistent instance (from theme or not), and the rest of drawing code stays the same.  

    I didn't look closely is it easily possible in this case -- so I'm just noting here, please take a look whether this simpler approach (without any boolean property to activate custom images) is possible. 

2. A minor note about the code style of "setter" methods for properies: see https://castle-engine.io/coding_conventions#_start_a_property_setter_with_if_fxxx_value_then . So instead of


    ```pascal
    procedure TCastleScrollViewCustom.SetSpaceForButton(const AValue: Single);
    begin
      if FSpaceForButton=AValue then Exit;

      FSpaceForButton:=AValue;
      VisibleChange([chRender]);
    end;
    ```

    we prefer inverse check for `FSpaceForButton <> AValue`, and put spaces around comparison/assignment operators:

    ```pascal
    procedure TCastleScrollViewCustom.SetSpaceForButton(const AValue: Single);
    begin
      if FSpaceForButton <> AValue then
      begin
        FSpaceForButton := AValue;
        VisibleChange([chRender]);
      end;
    end;
    ```

3. Thank you for the PR and again sorry for delay in applying your PRs. I admittedly had too much busy time lately, and this week I'm sick and trying to rest. But whenever I get back up to speed, my most important TODO is to apply your PRs, in particular Tiled improvements (animations, no seams). So, again, thank you, and keep up the good work, and excuse my delays in applying :)

Freedomax:
The Boolean attribute `CustomScrollbar` has been deleted, and all "Custom" prefixes have been removed. Currently, whether use **Custom** is enabled depends on whether the `ScrollbarSlider` is empty.
Removed `DrawFrameUnderButton` since it's not possible to render in `RenderOverChildren`.

michaliskambi:
Thank you! I played with this, added small code changes, merging :)

michaliskambi (approved):

PR #456, opened 2023-02-23
Fix #453

### LINKED ISSUE #453: Editor CPU usage is too high.
The CPU usage is only about 1% when running the game, but the CPU usage of the editor is always about 9%. What is the cause?

Freedomax:
```
class procedure TCastleApplicationIdle.ApplicationIdle(Sender: TObject; var Done: Boolean);
...
begin
  ...
  { Call DoUpdate for all TCastleControl instances. }
  for I := 0 to ControlsList.Count - 1 do
  begin
    C := ControlsList[I] as TCastleControl;
    C.DoUpdate;
  end;
  ...
end; 
```
In `C.DoUpdate`, `Invalidate` will be called, and `InvalidateRect` winAPI will be called on windows. The `ApplicationIdle` method will be called **thousands of times** per second (about 6000 times on my computer), exhausting the CPU.
I only used a simple Timer dedicated to handling `Invalidate`, and the cpu usage of the editor is close to 0% after testing.

### DIFF
--- a/src/lcl/castlecontrol.pas
+++ b/src/lcl/castlecontrol.pas
@@ -23,7 +23,7 @@ interface
 
 uses
   Classes, SysUtils,
-  StdCtrls, OpenGLContext, Controls, Forms, LCLVersion, LCLType,CustomTimer,
+  StdCtrls, OpenGLContext, Controls, Forms, LCLVersion, LCLType, CustomTimer,
   CastleRectangles, CastleVectors, CastleKeysMouse, CastleUtils, CastleTimeUtils,
   CastleUIControls, CastleRenderOptions,
   CastleImages, CastleGLVersion, CastleLCLUtils,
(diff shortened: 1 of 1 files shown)


michaliskambi:
Thanks, using timer for `TCastleControl` was indeed my TODO. 

I'll try later today, I want to do the remake a bit differently

- I want to make it more like TCastleControl on FMX and VCL, so that in long run TCastleControl on LCL, VCL and FMX is consistent,
- Get rid of using application "idle handler" in LCL completely (because the way we use it, always returning "Done := false", makes problems for other applications that could want to use LCL idle handler),
- No hardcoding of number "60". If anything, we use ApplicationProperties.LimitFps, though it should be used more consistently with current code in TCastleWindow and TCastleControl.

Anyhow, this confirms that our LCL idle handling is the culprit, thank you!

Freedomax:
new

I'm happy with the current CPU usage. Since you're going to handle it in a better way, I close this PR.

PR #457, opened 2023-02-26
Fix "Handle_IME"

Freedomax:
GIF 2023-02-26 08-32-32

Updated code to get caret position of `TCastleEdit`.
Replaced `if vhIMC <> 0 then ...` with `if vhIMC = 0 then Exit;`.Because the nesting of "begin end" is a bit too much.


Freedomax:
`TextHeight('中Wy')` calling `TTextureFontData.Glyph` will throw a warning:
"Font is missing glyph for character 中 (Unicode number 20013)"

There are several ways to deal with it, I haven't dealt with this warning yet because I'm not sure my way is good .@michaliskambi ,I think you can deal with it yourself.

michaliskambi:
Thank you! 

I added a small commit:

- Add additional blank lines between nested routines

- Added as comments 2 of your notes from https://github.com/castle-engine/castle-engine/pull/417 -- to keep this documented in code.

The modification to `src/window/windows/castlewindow_winapi.inc` is cool for me now.

The modification to `TCastleAbstractFont.Measure` -- I will want to experiment, and just expose dedicated properties to `MeasureRowHeight`, `MeasureRowHeightBase`, `MeasureDescend`. We need a general solution here, that also works e.g. for fonts that only have digits. Our current solution to this is "you can override Measure method" but this sucks for designing text in editor. And we need a way to customize all these measurements (unless I'll manage to remove `RowHeightBase` and just make it always `RowHeight - Descend`).

So, I'll experiment with approach for `TCastleAbstractFont.Measure` that I'm satisfied with, and then merge it.

Freedomax:
The current `GetCaretPos` is problematic, and many details have been ignored.
If I want to be precise, I need to re-write many code in `TCastleEdit.Render`, which is unacceptable. Because CGE lacks an interface like `TTextService`, I need to wait for it to improve it in the future.

michaliskambi:
> If I want to be precise, I need to re-write many code in TCastleEdit.Render, which is unacceptable. Because CGE lacks an interface like TTextService, I need to wait for it to improve it in the future.

Yeah, our text editing should become at some point more feature-rich, with TCastleMemo, and "real" handling of caret and selection common to (at least) TCastleEdit and TCastleMemo. I would like to attack it at some point (but probably after 7.0 release), starting by TCastleMemo, and common code for both TCastleEdit and TCastleMemo.

For now this is good -- I mean we just have TCastleEdit, and it has the simplest possible caret just always at the end.

michaliskambi (approved):
Merging, thank you!

As for modification how is font measured: I did a few improvements to it in CGE, you can now modify `MeasureHeight` and `MeasureCapHeight` properties to influence this comfortably (from code or from editor). 

Zrzut ekranu z 2023-03-06 04-48-07

Check out new examples/fonts/test_local_characters/ .

PR #459, opened 2023-03-01
Fix #458

### LINKED ISSUE #458: After setting "Culling", "Visible" is not correct
https://user-images.githubusercontent.com/26040454/221815877-065b243f-9435-41f6-89a5-2fab5057967e.mp4

Add 100 buttons to ScrollArea, and all buttons set `Culling := True`;
After resizing the window, many buttons are not displayed correctly.

[test-scrollarea-0.1-src.zip](https://github.com/castle-engine/castle-engine/files/10848846/test-scrollarea-0.1-src.zip)

Freedomax:
see #458.


michaliskambi:
I added a comment explaining the reason. Thank you for catching this!

PR #460, opened 2023-03-04
declare RemoveRoute

Freedomax:
This function is used to synchronize between different sprites.
Delete the original `TTimeSensor` route, and add the main sprite's `TTimeSensor` route.
I have tested it with examples, no problem.


michaliskambi (approved):
Looks good and makes sense, thank you!

PR #466, opened 2023-03-17
Fix custom editor template

Freedomax:
"custom_editor_template/castle_editor.lpi" is broken and cannot compile editor with custom controls. I fixed it and tested it.


michaliskambi:
This file is autogenerated by bash script `tools/build-tool/data/custom_editor_template_rebuild.sh`, that indeed required update after recent addition of another package as dependency of CGE editor. So I fixed the script, run it, and the result should be equivalent to your fix :)

Thank you!

PR #469, opened 2023-03-24
Fix `LoadTilesetFromURL`

Freedomax:
The path read by LoadTilesetFromURL is wrong, we should use a path relative to .tsx, not .tmx.
Here is the original function:
```
  { TSX file loading. }
  procedure LoadTilesetFromURL;
  var
    Doc: TXMLDocument;
  begin
    Doc := URLReadXML(URL);
    try
      Check(LowerCase(Doc.DocumentElement.TagName) = 'tileset',
        'Root element of TSX file must be <tileset>');
      Load(Doc.DocumentElement, BaseUrl);
    finally
      FreeAndNil(Doc);
    end;
  end;   
```
I fixed it with `Load(Doc.DocumentElement, URL);`
I also created a test map, `/examples/tiled/map_viewer_in_viewport/data/maps/test_tileset_path/desert_use_parent_dir_tsx.tmx`.

And avoid confusion, renamed private var `BaseUrl` -> `FBaseUrl`.


michaliskambi:
Thank you for the fix and test-case, great!

michaliskambi (approved):

PR #477, opened 2023-05-07
Avoid misleading log that config was saved when actually nothing was saved

eugeneloza:
Explicitly inform the user that "nothing was saved because there were no changes"

### DIFF
--- a/src/files/castlexmlcfginternal.pas
+++ b/src/files/castlexmlcfginternal.pas
@@ -91,7 +91,10 @@ type
     constructor Create(AOwner: TComponent); override;
     destructor Destroy; override;
     procedure Clear;
-    procedure Flush;    // Writes the XML file
+    { Writes the config to XML file
+      returns true if something was written
+      and false if there were no changes in config }
+    function Flush : Boolean;    // Writes the XML file
     function  GetValue(const APath, ADefault: String): String; overload;
     function  GetValue(const APath: String; ADefault: Integer): Integer; overload;
     function  GetValue(const APath: String; ADefault: Boolean): Boolean; overload;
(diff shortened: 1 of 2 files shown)


eugeneloza:
The check failed because of 
```
+ repository_cleanup . --remove-unversioned
repository_cleanup: Reverting GIT working copy in: .
Updated 3 paths from the index
repository_cleanup: Cleaning unversioned files.
repository_cleanup: Checking status is empty now.
repository_cleanup: Non-empty status, failed (not cleaned the repo as requested).
 M src/vampyre_imaginglib/src/Readme.md
 M tests/delphi_tests/README.md
 M tools/build-tool/data/README.md
```
on MacOS?

michaliskambi:
Thank you, merged!

I'm looking into what happened on macOS (I saw the previous PR also failed on macOS because of it). Definitely it's not related to this PR changes.

michaliskambi:
> As for Jenkins fail: I'm looking into what happened on macOS (I saw the previous PR also failed on macOS because of it). Definitely it's not related to this PR changes.

Indeed it was not related, and master should build OK..

These 3 files

```
 M src/vampyre_imaginglib/src/Readme.md
 M tests/delphi_tests/README.md
 M tools/build-tool/data/README.md
```

had for some reason Windows-like (CRLF) line endings even on Unix system, even though our `.gitattributes` specifies explicitly `*.md text eol=auto` (maybe they have been committed+pushed before this `.gitattributes` line, didn't investigate further now). It doesn't really seem to cause any issue on Linux, but on that macOS with Jenkins slave (maybe due to older GIT version there, 2.36) these 3 files were constantly marked as "modified".

I changed their line endings in master in https://github.com/castle-engine/castle-engine/commit/3832b0799fb91be720c112bf923b6f18668f8030 . This should make future forks/branches checkout OK on that macOS.

PR #490, opened 2023-05-15
Delphi without recompiling CGE every time

peardox:
Highly experimental ATM and it's only alterations to directory structure and config files (i.e. comments are hard to implement - althugh HTML-style commenting prob works - will expand + comment if it works)

ATM if you install this PR and add $(BDSCOMMONDIR)\CGE\$(Platform)\$(Config) to ...

Tools->Options->Language->Delphi->Win32 in the Library Path and Debug DCU Path entries then all libraries + units are accessable in a Delphi-native way without access to source (OK, you'd usually want that too - not what I'm working on here)

You will notice that I've got a Delphi10.4+ dir - which is for what it sounds like and is version automatic.
Delphi's < 10.4 need a very minor morification to the DPROJ + DPK to make them work as {$LIBSUFFIX AUTO} started with 10.4

You will notice that the project is now called caslte-engine280.bpl (or some other number  depends on Delphi version) - a Delphi specific convention.

This version was created with 11.3 CE which makes for a re-distributable CE version without source as long as the above paths are added to registry

Next I'm going to do the 10.3 and 10.2 versions of the dproj+dpk and test them. I already know exactly what needs altering but need to install those VMs to verify those versions also work.

A big detraction is that at present this is a Win32 ONLY solution. I know how to make it Win64 as well (and any other OS when we have them) so after < 10.4 I'll look a Win64


peardox:
Superceeded by delphi64 branch - notes still correct though

PR #491, opened 2023-05-15
Delphi 10.2, 10.3, 10.4 + 11 Win32 no source required

peardox:
This is an expansion of #490 that adds compiler specific support

They all require $(BDSCOMMONDIR)\\CGE\\$(Platform)\\$(Config) setting to become source optional

{$LIBSUFFIX AUTO} was added in 10.4 but I noticed that https://github.com/skia4delphi/skia4delphi defines 10.4 without using Auto. It then struck me that if someone has 10.4.0 or 10.4.1 it is known that the LIBSUFFIX feature was buggy - obviously this only affects a very small potential number of users (likely zero) but to be safe it semed advisable to specify 270 rather than leave it to Auto

The Delphi11 version uses Auto and as such is forwards-compatible for future versions (like Delphi 12, 13 etc) - should really be called Delphi11+ I guess

These are still Win32 only and both 10.2.3 and 10.3.3 as well as the initial 11.3CE have been tested with my Steam test (external Steam inclusion - not in PR)

All tested under both Debug and Release without any hint of where the CGE source lives.

Actually this seems more suited to a GetIt package - possibly some variant of the original is also a meaningful inclusion?

It should be possible to use these if multiple Delphi versions exist on the system - each Delphi has it's own libraries directory in  e.g. C:\Users\Public\Documents\Embarcadero\Studio\22.0 - this package targets that directory and outputs it's DCUs into a CGE folder - which is what $(BDSCOMMONDIR)\\CGE\\$(Platform)\\$(Config)  expands to - BDSCOMMONDIR is set by each Delphi Version at runtime and so is compiler-dependant

If you wanna test it out your 10.2.3 will be fine (using the Delphi10.2 project of course)


peardox:
Superceeded by delphi64 branch - notes still correct though

PR #492, opened 2023-05-15
Proper Delphi packaging with split RTL + Design-Time

peardox:
That went far better than I imaghined (OK, a few hiccups along the way but it now does what the title says)

Setting Search Path + Debug DCU Path for any desired platform to $(BDSCOMMONDIR)\\CGE\\$(Platform)\\$(Config) as detailed in #490 + #491. There will be some setting that also allows Source Access from within the IDE (Browse Path at a guess - will check)

All CGE compatible Delphi versions (inc CE) are supported. For people who are still on the 10.4.2 CE this will also work fine.

I've only defined Win32 + Win64 as available for the RTP as those are the only ones I know for a fact work. If any other platforms are supported then they can be easily added without code changes (as long as the code is good of course) - Just add some other platform to the castle-engine project/ The FMX.TCastleControl + VCL.TCastleControl are Design-Time only and thus (obviously) should only have win32 debug/release options

To use set a win32 and build all. Install the Design-Time packages. Switch to Win64 and build again (Design-time can be skipped of course, they're already done). If Linuyx / Mac are potentially working add the platform and build - dead simple.

Last step is to manually add $(BDSCOMMONDIR)\\CGE\\$(Platform)\\$(Config) in applicable platforms - each needs doing for Delphi to find the units.


src/delphi/vcl.castlecontrol.reg.pas
```diff
@@ -0,0 +1,45 @@
...
+
+  "Castle Game Engine" is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  ----------------------------------------------------------------------------
+}
+
+{ Control with OpenGL context on a Delphi VCL form. }
+unit vcl.castlecontrol.reg;
```
michaliskambi, line 17: Make unit names use `PascalCase` with dots, so `Vcl.CastleControl.Reg`.

Moreover, to make the unit suitable for possible other design-time stuff, please call it just `Vcl.CastleControl.Design`.

src/delphi/vcl.castlecontrol.reg.pas
```diff
@@ -0,0 +1,45 @@
...
+  included in this distribution, for details about the copyright.
+
+  "Castle Game Engine" is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  ----------------------------------------------------------------------------
+}
+
+{ Control with OpenGL context on a Delphi VCL form. }
```
michaliskambi, line 16: Please change this comment to 

```
{ Design-time functionality for the Castle Game Engine on Delphi VCL form. }
```

Reasons:

- The idea is to emphasize already here that this unit is only for design-time stuff.

- Also instead of "OpenGL context" we prefer more generic terms (in case of this unit "Castle Game Engine on Delphi VCL form" seems OK), as eventually we'll have other renderers.

src/delphi/vcl.castlecontrol.reg.pas
```diff
@@ -0,0 +1,45 @@
...
+{$I castleconf.inc}
+
+interface
+
+uses SysUtils, Classes, Vcl.Controls, Vcl.ExtCtrls, Types,  WinApi.Messages,
+  CastleGLVersion, CastleGLUtils,
+  CastleInternalContextBase, CastleInternalContextWgl, CastleInternalContainer,
+  CastleVectors, CastleKeysMouse, VCL.CastleControl;
+
+{ Designtime Only }
```
michaliskambi, line 28: No need for this comment `{ Designtime Only }`. It should already be said at unit header, as per previous review comment.

Same thing for `{ Designtime Only }` in implementation.

src/delphi/vcl.castlecontrol.pas
```diff
@@ -14,7 +14,7 @@
 }
 
 { Control with OpenGL context on a Delphi VCL form. }
-unit Vcl.CastleControl;
+unit vcl.castlecontrol;
```
michaliskambi, line 17: Do not make the unit name lowercase. Please keep it `PascalCase`, so `Vcl.CastleControl`.

src/delphi/vcl.castlecontrol.pas
```diff
@@ -109,21 +109,25 @@       TContainer = class(TCastleContainerEasy)
     property OnMouseMove;
     property OnMouseWheel;
   end;
-
+{ Designtime Only }
```
michaliskambi, line 112: Remove the sections that you moved to other (design-time only) unit, do not only comment them out.

src/delphi/fmx.castlecontrol.reg.pas
```diff
@@ -0,0 +1,47 @@
...
+  included in this distribution, for details about the copyright.
+
+  "Castle Game Engine" is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  ----------------------------------------------------------------------------
+}
+
+{ Control with OpenGL context on a Delphi FMX form. }
```
michaliskambi, line 16: Please apply here analogous comments as for VCL counterpart. So

```delphi
{ Design-time functionality for the Castle Game Engine on Delphi FMX form. }
unit Fmx.CastleControl.Design;
```

and remove further `{ Designtime Only }` comments.

src/delphi/fmx.castlecontrol.pas
```diff
@@ -14,7 +14,7 @@
 }
 
 { Control with OpenGL context on a Delphi FMX form. }
-unit Fmx.CastleControl;
+unit fmx.castlecontrol;
```
michaliskambi, line 17: Please apply similar comments as for VCL counterpart, so keep unit name `PascalCase`, so `Fmx.CastleControl`. And just remove (not only comment-only) the design-time pieces moved.

packages/delphi/notes.txt
```diff
@@ -0,0 +1,21 @@
+$(BDSCOMMONDIR)\Bpl
```
michaliskambi, line 1: I'm afraid these notes are not really decipherable, even when I know some comments about what you're doing. Please remove them from this PR. And / or reformat the useful bits and place in packages/delphi/README.md in Markdown format.

michaliskambi (changes requested):
Thank for very much! I pointed out some minor things below in the review. Let me add here some *big* ones :) 

1. This approach creates a number of new package files, for each Delphi version (10.2, 10.3, 10.4, 11+).

   That will be really hard maintain, as:

    - each change to units or directories will require to be carefully duplicated for all Delphi versions. 
    
    - And as most people will just use Delphi 11.x at this point, it will be difficult to test. I could set up CI (Jenkins) to do this.... but this is also significant extra work if we need to maintain a big collection of Delphis across many years in the future. 
    
    - Moreover, adding each new Delphi version (as we spoke in other issues, it seems we can reasonably make CGE compatible with >= XE7) would make this more and more painful, as I guess each new Delphi would require new directory in this case.

   So I would definitely want to go back to only 1 set of files, that can "just work" on all Delphi versions we support. Our current design-time packages already work for all these Delphi versions, from 10.2 up to the latest Delphi 11.x. Can you explain why it cannot work after the split between design-time and run-time?

    You mentioned the `{$LIBSUFFIX AUTO}` but I do not understand how is this critical. Each Delphi version has, or can have, a separate directory for BPL? So we don't need a suffix with Delphi version added to each BPL?

    I know that many Delphi libraries have Delphi-version-specific packages, but I honestly hoped to avoid it with CGE. Also looking at our dependencies :

    - Vampyre has only 1 set of packages: https://github.com/galfar/imaginglib/tree/master/Packages . 
    
    - Delphinus has only 2 set of packages, for >= XE 6 and before. But I know that Delphinus is not a good example here, as it only has design-time packages.

2. What is the relation of this PR to https://github.com/castle-engine/castle-engine/pull/490 and https://github.com/castle-engine/castle-engine/pull/491 ? From a brief look, it is a next version, and so the 2 previous PRs can be closed? If yes please close them.

peardox:
After extensive exploration it turns out that only one 'template' Delphi package can be used to derive the specific ones

Take any Delphi package and all that differs is the library versioning - in the gpk this is libsuffix and the drpoj has dllsuffix. These should be set according to Delphi version in use at build - 10.2 = 250 .... 11.x = 280

ATM you ignore versions and just output castle_engine while the Delphi 'correct naming' would be e.g. castle_engine250 - the compiler will actually look for both versions with the numbered one taking precedence.

Not using version numbers can result in finding the lib for another version of Delphi (it's a bit of a pain in the behind but this is what the compiler looks for)

michaliskambi:
> ATM you ignore versions and just output castle_engine while the Delphi 'correct naming' would be e.g. castle_engine250 - the compiler will actually look for both versions with the numbered one taking precedence.
> 
> Not using version numbers can result in finding the lib for another version of Delphi (it's a bit of a pain in the behind but this is what the compiler looks for)

What is the user experience when compiler finds the wrong BPL version? 

E.g. I compiled package for Delphi 10.2, without any special suffix. Now I run Delphi 10.3, but it picks up the BPL compiled for 10.2. Naturally it will not work (due to at least RTL changes between these Delphis). What does Delphi display? A. Does it show some message like _"Linked with wrong RTL version"_ or something? B. Or is it worse, does it crash at some future unspecified point with some "Access Violation"?

Can you make a screenshot?

peardox:
Can't remember exact wording but essentially yoou'll get a compile-time error complaining about wrong lib version. Similar to trying to mix x86 + x64 will do. I have seen this when testing other packages without a version suffix. The error message is not very user friendly (like all compile errors).

If you look at the delphi distribution you'll find they call all theirs e.g. rtl280 - just look in e.g. C:\Program Files (x86)\Embarcadero\Studio\22.0\bin or redist - everything has a suffix...

michaliskambi:
> Can't remember exact wording but essentially yoou'll get a compile-time error complaining about wrong lib version. Similar to trying to mix x86 + x64 will do. I have seen this when testing other packages without a version suffix. The error message is not very user friendly (like all compile errors).

Can you make a screenshot with the exact message that appears? When trying to load BPL from e.g. Delphi 10.2 into Delphi 10.3?

To be clear, I agree that Delphi-version-specific suffix would be good. But I want to judge how critical this is. If the eventual error message is understandable, and in consequence people with multiple Delphi version on the same machine will just have to occasionally recompile CGE -- this is acceptable. I do not expect many people with old Delphi versions (esp. when there's Delphi 11 CE for free, *and* FPC / Lazarus for free) and even they will unlikely actively use multiple Delphi versions. I imagine most people just use 1 Delphi version regularly and most people plan to upgrade eventually, if their current version is far behind.

Having one package files is better than maintaining multiple Delphi-specific package versions, even if we would be able to generate them by a script. Much less maintenance pain.

peardox:
This will take me several hurs to do properly so I'll report back later

michaliskambi:
Thank you for https://github.com/castle-engine/castle-engine/issues/499 .

While the error message isn't clear (it doesn't say that BPL loading failed due to incompatible Delphi version, and that the solution is just to rebuild the package) but it is at least a simple message, i.e. Delphi doesn't crash at some unspecified moment trying to use incompatible BPL -- it just refuses to load it.

I light of this, my thoughts from previous post stand. We don't want to introduce Delphi-version-specific subdirectories in packages/delphi/ . Too much complication to maintain, even if the subdirectories would be generated by a shell script. The people who switch between Delphi versions will just see the error message from https://github.com/castle-engine/castle-engine/issues/499 .

PR #501, opened 2023-05-31
Cge getit test

peardox:
This is a simulation of what GetIt should be doing...

This includes packages/getit which is a modified split delphi

PLEASE read the README.md as there is a lot of manual setup involved that GetIt should be automating (which of course we have to do manually to test). Any skipped steps WILL cause problems so do them carefully AND in order.

I have tested on a properly cleaned out Delphi 10.2.3 (this is not the same as a clean 10.2.3) and it worked compiling and running the https://github.com/peardox/CGEFormTest tests perfectly Win32/64 Debug/Release

10.2.3 actually had BDSCatalogRepository set as a User Documents subfolder - so check the value (11.x should be as the README.md tho)

There are other things we'll want GetIt to do such as adding the BrowsingPath

It should be noted that this version is NOT suitable for Engine development - it's an end-user package


michaliskambi:
Note that this PR is somewhat obsolete in light of new CGE packages organization ( https://castle-engine.io/wp/2024/02/23/new-delphi-packages-organization-and-tools-castle-game-engine-menu-in-delphi-ide/ , https://castle-engine.io/delphi_packages ). I'm also unsure whether this reorganization is actually required by GetIt, as the title and description suggests -- GetIt packages in Delphi seem to essentially do whatever the authors want, some (like FMXLinux) also just add sources to "Library Path" that are recompiled with each project. 

That said, I'm keeping this PR open, because we will likely one day change our recommended approach to follow this, and reuse compilation results of CGE packages. It is already required by C++ Builder. It avoids the need to recompile CGE in each new project, which is useful.

So the concept of this PR will be applied to CGE at some point.

PR #508, opened 2023-07-02
Updates for Android - better sound support, target API lvl 33, compile API lvl 19,  Admob 22.1

and3md:
## Sound problems
On a new Android 13 device, I noticed sound issues in [my game](https://play.google.com/store/apps/details?id=com.digitalkarabela.cge.colorpick). I had to update OpenAL with Oboe support to fix that (newest version without Oboe not working) 

So I made new OpenAL build using official sources with tag 1.23.1 with Oboe 1.7.0. The build procedure was quite complicated so I wrote I article about that on [my site](https://digitalkarabela.com/how-to-compile-openal-library-for-android-with-oboe-library/).

I made build with shared c++ stl library.  But I didn't know how and where to add these c++ libraries (they are form android NDK, maybe gradle can add it automatically with some option?)

Why I do shared build? Because Google recommends that  - https://developer.android.com/ndk/guides/cpp-support#static_runtimes . If you prefer static build LMK.

## Target API LVL 
I also updated API lvl to 33. That lvl will be soon requirement for Google Play. So I thing we should test everything with this level.

## Admob
It turned out that our version of Admob can stop working in [30 June 2023](https://developers.google.com/admob/android/deprecation) so I updated it to latest release 22.1 (API changed a lot so I rewrite our module).

## Compile API LVL

Updated to 19 because new Admob needs that.

## Testing 
Changes tested on four phones (From Android 6.0 to 13)


michaliskambi (changes requested):
Thank you! All the proposed changes look good, my comments are only about libraries :)

1.

> I made build with shared c++ stl library. But I didn't know how and where to add these c++ libraries (they are form android NDK, maybe gradle can add it automatically with some option?)

  This is indeed important to check. If we don't need to maintain c++ shared library, that's one less headache :)

2. We should have rebuilding openal reproducible with a repo. (Because experience shows these libs have to be updated occasionally.)

    The OpenAL version from current master is in https://github.com/castle-engine/android-openal . It's a fork of https://repo.or.cz/w/openal-soft/android.git (which is already a fork of OpenAL Soft).

    You should either update that repo, or create a new one with fork of https://github.com/kcat/openal-soft (e.g. to https://github.com/castle-engine/android-openal-2 , and we can later archive previous repo). Then add there makefiles / scripts to automate the process described on make build as much as possible (including CGE-specific stuff, if need be -- it will be a fork of OpenAL Soft we don't need to merge to upstream, just maintain it to keep being able to recompile OpenAL for CGE). The goal is to have it as easy to recompile as

    - https://github.com/castle-engine/android-openal
    - https://github.com/castle-engine/android-freetype
    - https://github.com/castle-engine/android-tremolo
    - https://github.com/castle-engine/libpng-android

and3md:
> This is indeed important to check. If we don't need to maintain c++ library (shared or not), but NDK provides something ready, that's one less headache to maintain :)

I was hoping you knew, so I haven't searched yet :)

tools/build-tool/data/android/integrated/app/build.gradle
```diff
@@ -10,13 +10,21 @@ android {
         versionCode ${VERSION_CODE}
         versionName "${VERSION}"
 
+        ndkVersion "22.1.7171670"
```
michaliskambi, line 13: Do we need to specify this exact version? Seems very specific.

tools/build-tool/code/toolandroidmerging.pas
```diff
@@ -239,8 +239,19 @@ procedure MergeBuildGradle(const Source, Destination: string;
   DestinationContents: string;
   Doc: TXMLDocument;
 
+  function MakeIndent(const Indent: Integer): String;
```
michaliskambi, line 242: Use standard DupeString from StrUtils.

michaliskambi (changes requested):

tools/build-tool/data/android/integrated/app/build.gradle
and3md: By default gradle used 21 but I just wanted use the same I precompiled libraries before (to make easier testing). So we don't need it. But we should add ability to choose it in our manifest file (as far as I know we don't support that now). That should help someone with close sourced c++ libraries that was compiled with specified ndk version to use the same version with our libraries.

and3md: https://developer.android.com/ndk/guides/cpp-support#one_stl_per_app

michaliskambi:
I've done small improvements, in particular:

1. We now let Android Gradle plugin to download Oboe, so we don't have to distribute Oboe sources. We follow https://developer.android.com/games/sdk/oboe , https://developer.android.com/games/sdk/oboe/ . OpenAL's CMakeLists.txt already contains proper lines to require linking with Oboe in this case (`find_package` and extending `target_link_libraries`) when `OBOE_SOURCE` is not defined but `ALSOFT_REQUIRE_OBOE` is on.

2. I fixed building in a directory with spaces, like in `/home/xxx/Castle Game Engine Projects/xxx/`. Submitted to upstream too, as https://github.com/kcat/openal-soft/issues/889 .

michaliskambi (approved):

PR #534, opened 2023-09-20
Fixed sorting of layers in TiledLayersPropertyEditor.

and3md:
By simply adding space when there is only one digit in number.
Before:

obraz

After:
obraz


michaliskambi (approved):
Oh that's excellent solution, thanks. It's trivial to implement and it does the job. 

I see you noticed in my comments I've fought with this issue in the past without success :), I didn't think of this.

PR #536, opened 2023-09-30
Add more file-format properties to TCastleTiledMapData.TLayer & TTile, add new Exists-property to show hide layer during runtime

Dennis1000:
Added some more properties to the TCastleTiledMapData.TLayer and TCastleTiledMapData.TTile class.
Prefixed the "Class" and "Type" property with "&" because of compiler keywords.

Added a Exists-property to TCastleTiledMapData.TLayer to show/hide the layer during runtime.
Deprecated therefore the TCastleTiledMap.Layers array.


michaliskambi:
Thank you very much for PR and sorry for a big delay in reviewing + applying! Don't let this stop you from contributing more, please -- I'm just overloaded, but I'll get to it eventually. As we talked on Discord, this is a good and desired change, I want to test and apply it.

I'm back from ITDevCon conference last week ( https://github.com/castle-engine/conference-itdevcon-2023/ ) where I naturally talked about Castle Game Engine. I hope to catch up with my backlog of tasks soon :)

Dennis1000:
This is just the short (aka fastest, non-conflicting) version of the layer & tile update. I really would like to move/map the layer and the tile to the TCastleTiledMap class instead of the underlying CastleTiledMapData property. But this would be a breaking change, as we've talked lately.

michaliskambi:
Sorry for taking so long time to review! I finally merged a big work last week, and I'm catching up with stuff.

I pushed small updates to this PR and I'm merging it. Thanks a lot - for the work, and for the patience :)

> I really would like to move/map the layer and the tile to the TCastleTiledMap class instead of the underlying CastleTiledMapData property. But this would be a breaking change, as we've talked lately.

I fully agree.

Admittedly I wanted to "start small", with small modification, to only change `TCastleTiledMapData.TLayer` but I'm not happy with my decision now. (*From what I recall* it was my decision -- my memory is foggy about exact details we discussed, which of course only confirms that I took way too long to deal with this PR.) In the end, having read-write `Exists` in `TCastleTiledMapData.TLayer`, alongside read-only `Visible` in the same class, that actually exposes the same value, is not optimal API.

I added some TODOs to code, to summarize our plans, and I described our plans for better Tiled API on https://castle-engine.io/roadmap#tiled_layers . I also want to deal soon with https://castle-engine.io/roadmap#component_list first, which should be a good basis for this.

PR #557, opened 2023-12-19
Build Tool Changes for VSCode Extension

and3md:
Some new options for better integration with coming soon VSCode extension:
- output-environment fpc-exe - returns path to fpc used by build tool (for e.g. to find the bundled one)
- output executable-name - returns program name
- compile-run - compile and run in one step


tools/build-tool/castle-engine.dpr
```diff
@@ -179,9 +180,16 @@ begin
             'output' +NL+
             '    Output some project information (from the manifest).' + NL +
             '    Next parameter determines the information:' + NL +
+            '      executable-name' + NL +
             '      version' + NL +
             '      version-code' + NL +
             NL+
+            'output-environment' +NL+
+            '    Output some environment informations.' + NL +
```
michaliskambi, line 188: `Output some environment informations.`
->
`Output some environment information (independent of any project).`

tools/build-tool/castle-engine.dpr
```diff
@@ -179,9 +180,16 @@ begin
...
+            '    Next parameter determines the information:' + NL +
+            '      fpc-exe' + NL +
+            '      ???' + NL +
```
michaliskambi, line 191: Remove `???`

tools/build-tool/castle-engine.dpr
```diff
@@ -414,6 +427,18 @@ begin
           Project.DoRun(Target, OS, CPU, RestOfParameters);
         finally FreeAndNil(RestOfParameters) end;
       end else
+      if Command = 'compile-run' then
+      begin
+        Project.DoCompile(OverrideCompiler, Target, OS, CPU, Mode, CompilerExtraOptions);
+        RestOfParameters := TCastleStringList.Create;
```
michaliskambi, line 433: This is copy-paste of logic done for `run`, better extract it to a function `CreateRunParameters` to be reused both when 

`if Command = 'run' then` 

and when

`if Command = 'compile-run' the`

Like

```
function CreateRunParameters: TCastleStringList;
begin
  Result := TCastleStringList.Create;
  Result.Text := Parameters.Text;
  Result.Delete(0); // remove our own name
  Assert(SameText(Result[0], 'run') or SameText(Result[0], 'compile-run'));
  Result.Delete(0); // remove "run" or "compile-run"
end;
```

tools/common-code/toolcompilerinfo.pas
```diff
@@ -41,6 +41,10 @@ function FindExeLazarus(const ExeName: String): String;
   and LazarusCustomPath. }
 function PathExtendForFpcLazarus(const PathList: String): String;
 
+{ Returns compiler info to std output
+  Keys: fpc-exe - path to the compiler executable }
+procedure DoOutputEnvironment(const OutputKey: String);
```
michaliskambi, line 46:
```
{ Returns compiler info to std output
  Keys: fpc-exe - path to the compiler executable }
```

-> 

```
{ Write environment information to the standard output. 
  See "castle-engine --help" for valid arguments after "castle-engine output-environment",
  these are values allowed for OutputKey. }
```

michaliskambi (changes requested):
Thank you! I made a quick review -- sorry for terseness, I did it a bit fast, if anything is unclear let me know :)

and3md:
I didn't expect there will be so many small issues. I was confident that I had checked the changes earlier and I opened the pull request thoughtlessly, without verifying the content. Sorry about that. I was too focused on the vscode extension itself.

and3md:
Notice please that `output executable-name` returns the name without an extension on Windows. It comes from the manifest, and I assumed you want this behavior (I added the '.exe' on the plugin side). If not, please let me know.

michaliskambi:
Thanks, all changes are good.

Additional 2 notes:

1. I see in https://github.com/castle-engine/castle-engine/pull/557/commits/923eb9e38fa9eabd786ccad720b8c1cc6bd6c0d7 you needed to add `13: ;`. That's because you have the entry 

    ```
    (Short: #0 ; Long: 'output-environment'; Argument: oaRequired),
    ```

    in the `Options` array.

    But it should not needed. You should revert https://github.com/castle-engine/castle-engine/pull/557/commits/923eb9e38fa9eabd786ccad720b8c1cc6bd6c0d7  and remove the `(Short: #0 ; Long: 'output-environment'; Argument: oaRequired),` entry. That would only be useful to support a long option like `castle-engine --output-environment=something some-command`, but that is not what you want to achieve -- you want the `output-environment` to be the command. For this, you *should not* put `output-environment` in the `Options` array.

2. As for:

    > Notice please that output executable-name returns the name without an extension on Windows. It comes from the manifest, and I assumed you want this behavior (I added the '.exe' on the plugin side). If not, please let me know.

    This is good, i.e. `executable-name` should match behavior of `executable_name` in manifest, which we document explicitly in https://castle-engine.io/project_manifest as 

    > executable_name: Binary name (without platform-specific extension, like .exe on Windows). Optional, by default equal to just name.

    So it is good how it is, i.e. `castle-engine executable-name` should not have extension, for consistency. This way it is also independent from the current target OS. 

    If it would make life easier for the VS code plugin, we could introduce another option to return this with extension, just make sure to name it differently, like 

    - `executable-name-with-extension` and return `ChangeFileExt(ExecutableName, ExeExtensionOS(OS))` 
    - or `full-executable-name` (with absolute path) and return `Path + ChangeFileExt(ExecutableName, ExeExtensionOS(OS))`.
    
    I am fine with all, or one, or zero of the additions above, whatever will make it easier for VS Code plugin here :)

and3md:
Note 1:
Oh During reading the sources I assumed that all options must be in this array.  Reverted and removed from Options array.

Note 2:
> This is good, i.e. executable-name should match behavior of executable_name in manifest, which we document explicitly in https://castle-engine.io/project_manifest 
I am fine with all, or one, or zero of the additions above, whatever will make it easier for VS Code plugin here

I don't currently need the executable-name version with the extension, so maybe it's not worth introducing it at this point.

michaliskambi:
Thank you, merged!

PR #564, opened 2024-01-29
An option to get free pascal standard units paths from build tool

and3md:
Added an option to get free pascal standard units paths used by vscode extension when freepascal installation does not have fpc.cfg.


michaliskambi (changes requested):
Thank you!

I only have a request to change 

- `FpcStandardUnitsPaths` -> `FpcStandardUnitsPath` , 
- `fpc-standard-units-paths` -> `fpc-standard-units-path` 
 
That is, without `s` at the end, as this is one path, and the `FindExeFpcCompiler` parameter is also called like that.

michaliskambi (approved):

PR #566, opened 2024-02-01
Update library interface with new function to set navigation input shortcuts

janadamec1:
Changes to (deprecated) library:

1. new exported function to set InputShortcuts to Navigation
2. Fixed crash bug when reopening the scene using library interface (nilling MainScene pointer after close)
3. Better exception reporting - function name added

I've tested the functionality inside my app.

This pull request also fixes compilation of C++ example applications using the library interface (cpp_winapi_library_tester, qt_library_tester). I've also added some Visual Studio folders (used in cpp_winapi_library_tester) to gitignore to have the clean output.

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,10 @@ xcuserdata
 /.jenkins
 AndroidSigningProperties.txt
 
+# Visual Studio stuff
+.vs/
+*.vcxproj.user
+
 # cge build tool stuff
 castle-engine-output
 # Already included by *.res above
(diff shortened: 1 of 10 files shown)


michaliskambi (approved):

michaliskambi:
Thank you very much! I noticed your work in the branch yesterday. All perfect, no comments, applying :)

In particular thanks for the `if Navigation.Camera <> nil then` -> `if Navigation.Camera = nil then` fix in `castlescenecore.pas`, that was a dumb mistake from my part. It seems I did it during https://github.com/castle-engine/castle-engine/commit/8145ad28f2b72587d944f8d3066e151ed83bf3b2 , and I mistakenly used `<>` instead of `=` and never checked whether it works. 

Note that we have a a number of automatic tests in CGE in `tests` subdirectory, if you want to add a test making a `AddViewpointFromNavigation` call (to make sure that such simple mistake will never again occur in the future), you're welcome!

PR #580, opened 2024-03-05
Font texture cache fix (hack) to solve OpenGL texture resources freeing when first TCastleText is released

and3md:
Now when I know what the problem is, I'm surprised I didn't come up with the solution sooner :D
As we anticipated, the issue was in the cache, but in a different manner. After meticulously examining the `TX3DFontTexturesCache` code many times, I was convinced that the error lies elsewhere.

Each `TCastleScene` releases OpenGL resources when is destroyed. In that case cached, shared font texture is also removed on renderer side. That's why we don't have a crash but only visual garbage in other `TCastleText` instances. 

When `TCastleText` is destroyed it destroys also internal TCastleScene that in `TCastleTransform.Destroy()` calls `GLContextClose()`. That calls `TGLShape(Shape).GLContextClose;` Next `ShapesGLContextClose()` is called:

```
procedure TGLShape.GLContextClose;

  if TexturesPrepared then
  begin
    TexturesPrepared := false;
    EnumerateTextures({$ifdef FPC}@{$endif} UnprepareTexture); <--- here is the problem
  end;
```
and next:

```
function TGLShape.UnprepareTexture(Shape: TShape; Texture: TAbstractTextureNode): Pointer;
begin
  Result := nil; // let EnumerateTextures to enumerate all textures
  TTextureResources.Unprepare(Texture);
end;
```

The simplest fix is a little hacky but I believe that when OpenGLContext is really closed that code will tidy up GL resources
```
TAbstractTextureNode.SetInternalRendererResource(const Value: TInternalRendererResource); mamy:

      ApplicationProperties.OnGLContextCloseObject.Add(
        {$ifdef FPC}@{$endif} GLContextCloseEvent);
```
and we can call `InternalRendererResourceFree;` when font texture is destroyed in any other cases.

Font textures are `TPixelTextureNode` but I needed to make them stand out somehow so I created  `TInternalCachedPixelTextureNode` for that case:

```
function TGLShape.UnprepareTexture(Shape: TShape; Texture: TAbstractTextureNode): Pointer;
begin
  Result := nil; // let EnumerateTextures to enumerate all textures
  if not (Texture is TInternalCachedPixelTextureNode) then
    TTextureResources.Unprepare(Texture);
end;
```
I know that is a hack because when we will want to cache other things we will get the same problem again. But more clean fix needs an archtecture change (maybe we need some render resource manager) or maybe you will know better solution that I don't notice.

### DIFF
--- a/src/scene/castlesceneinternalshape.pas
+++ b/src/scene/castlesceneinternalshape.pas
@@ -212,7 +212,21 @@ end;
 function TGLShape.UnprepareTexture(Shape: TShape; Texture: TAbstractTextureNode): Pointer;
 begin
   Result := nil; // let EnumerateTextures to enumerate all textures
-  if not (Texture is TInternalCachedPixelTextureNode) then
+
+  { Do not call TTextureResources.Unprepare
+    (which would do Texture.InternalRendererResourceFree in the end)
+    when Texture is TInternalReusedPixelTextureNode,
+    since this texture may be reused by other scenes.
+
+    Testcase:
+    - create in editor viewport, add TCastleText
+    - add another TCastleText
+    - remove one (any one) TCastleText --
+      the other TCastleText display will be broken
+      if condition "if not (Texture is TInternalReusedPixelTextureNode) then"
+      below gets removed.
+  }
+  if not (Texture is TInternalReusedPixelTextureNode) then
     TTextureResources.Unprepare(Texture);
 end;
 
--- a/src/scene/castlesceneinternalshape.pas
+++ b/src/scene/castlesceneinternalshape.pas
@@ -212,7 +212,8 @@ end;
 function TGLShape.UnprepareTexture(Shape: TShape; Texture: TAbstractTextureNode): Pointer;
 begin
   Result := nil; // let EnumerateTextures to enumerate all textures
-  TTextureResources.Unprepare(Texture);
+  if not (Texture is TInternalCachedPixelTextureNode) then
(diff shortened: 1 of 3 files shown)


michaliskambi:
While initially the solution looked to us both as "seems like a hack" (we add an exceptional behavior just for the faulty case) , after some thinking I decided to "bless it" as "just the right solution" :)

Reason: 

- FontTexture is indeed exceptional -- it is X3D node, enumerated by `EnumerateTextures`, but it is *shared* by multiple TCastleScene instances. Such node sharing is not allowed in general (each different TCastleScene instance must have different X3D nodes, and any resource sharing like OpenGL texture id is done on "deeper" level). So the font textures are an exception here.

- A more invasive fix could even be to not return `FontTexture` with `TShape.EnumerateTextures`. But then, *some* usage depends on it (e.g. we need to prepare font texture with shape, to render font OK). So it would require a whole new approach to have "nodes that are used by this shape, and must be prepared, but may be shared with other TCastleScene". We deliberately wanted to avoid this code complication, that's why engine has for a long time a rule "all TCastleScene instances must have different X3D nodes"... but our font texture breaks this rule.
 
- So I say special exception for not freeing too early by `Unprepare` is just justified and right solution, and it seems complete.

I simplified the PR a tiny bit (no need to call `C.FontTexture.InternalRendererResourceFree`, it will actually happen automatically when freeing the node already). Applying. Thank you!

PR #581, opened 2024-03-12
Fix memory leaks in TCastleDownload and proposal for faster and safe download thread freeing

and3md:
There was two memory leaks: 
1. `FOwnsContents` was not set to `true`;
2. In case of the approach with unsafe `Thread.FreeOnTerminate := true;` the critical section was not freed

But I also created new a little funny solution that is safe, fast and enabled by default now. When `CASTLE_SAFE_DOWNLOAD_INTERRUPTION_CLEANING_THREAD` is defined. We create another cleaning thread that is `FreeOnTerminate = true` and waits for download thread termination. If you like it you can delete all old solutions and critical section freeing from `TFpHttpClientJob.Destroy;`

```
  if (ParentThread <> nil) and (ParentThread.FreeOnTerminate) then
    FreeAndNil(SynchronizationCS);
```

### DIFF
--- a/src/files/castledownload_asynchronous.inc
+++ b/src/files/castledownload_asynchronous.inc
@@ -197,7 +197,7 @@ type
       and still keep the stream reference.
       It is your responsibility then to keep and free the @link(Contents)
       stream whenever you want. }
-    property OwnsContents: boolean read FOwnsContents write FOwnsContents;
+    property OwnsContents: boolean read FOwnsContents write FOwnsContents default true;
 
     { How many bytes were downloaded.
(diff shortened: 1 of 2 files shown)


michaliskambi:
A thread observing another thread seems excessive, in the light that thread can free itself by FreeOnTerminate :) I added `OwnsSynchronizationCS` that results in simpler PR but achieves the same, if I think right. It also fixes mem leaks (tested on Linux, with castle_download and asynchronous_downloader, aborting downloads in progress in latter case).

PR #584, opened 2024-03-19
Castle download support for httpbody/httpheaders for Android

phomm:


phomm:
@michaliskambi , hi, please review :)
Not sure about printing stack trace to log, could be removed if too verbose/insecure
And as always, suggest corrections

michaliskambi:
Thank you for updating the PR, I see it and this is much appreciated. Sorry it's taking me so long, I will get to it -- I'll test and apply :)

michaliskambi:
Thanks again for being patient. I have just created the perfect Castle Game Engine application to test this PR -- see https://github.com/michaliskambi/castle-openai :) I will announce it around the weekend. Once I merge your PR, this application should automatically work also on Android.

michaliskambi:
Thank you again! Tested, reviewed and merging. https://github.com/castle-engine/castle-openai rocks on Android thanks to you! 

I pushed now tiny improvements:

1. I updated the comment _"TODO: We may change this behavior in the future, to make it such that..."_ to better explain what I imagine for the future.  

   I saw that you made a fix to previous invalid comment (it was _"HttpPostData will automatically override HttpPostData"_...), but I wanted to outline a different future :)

2. Removed printing HTTP headers to log (one can debug them anyway by `Messaging.Log := true` from program, and I was afraid that showing them by default may be too risky -- HTTP headers may expose some passwords / API keys to anyone looking at log).

3. As for

  > Also not sure if packing/unpacking headers is fine enough, maybe need to apply some more strict rules and correlate to other platforms

  Good point -- for now I just added a check that provided header key/value just doesn't contain ":". This way, if this is ever a real problem, user will get a clear exception saying what's wrong.

I noticed you made other improvements to the Java side (comment _"TODO: we need to consider a way to deal with binary data"_, more obvious code thanks to `connection.connect()` and `connection.getResponseCode()` on own lines, using `connection.getErrorStream()` ) and I like them all, thank you. I don't really have an opinion whether the change to display `PrintStackTrace` is too verbose or not -- maybe I'll make an opinion once I see it :), for now I leave it as you did, no complaints from me.

phomm:
Thanks, night-superhero 😅
Regarding comments - fully agree, bc you have better vision for future.
Regarding logging request - totally agree, it is a security thing, I think I had idea to make it configurable, but your method is really better.
Regarding headers validation - explicit one is way better than naive, and I didn't elaborate this, unfortunately, and I will pay attention in future! 
And I agree, let's test and see if StackTrace has value

AND overall Thank you for reviewing and merging, I know you have such a wide scope, but working stoutly for years !

PR #589, opened 2024-03-29
Effekseer service on Android (precompiled libraries)

PTz0uAH:
just got Effekseer for Android working so here my non-official integrated service for the build-tool.. now the Example from Kagamma (https://github.com/Kagamma/cge-effekseer) works OOTB :-)


tools/build-tool/data/android/integrated-services/effekseer/README.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+  https://github.com/Kagamma/cge-effekseer
+
+Available binaries
+https://github.com/Kagamma/cge-effekseer/releases/download/1.61a/effekseer-binaries-161a.zip
+
+Effekseer for CGE is based on the official libeffekseer code: https://github.com/effekseer/Effekseer
+and has its own fork at: https://github.com/castle-engine/Effekseer
+
+To build Effekseer for CGE please read the instructions at: https://github.com/Kagamma/cge-effekseer
```
michaliskambi, line 43: 1. We need to extend these instructions, to make them trivial to repeat and allow everyone to recompile all SO files for Android. The reason is that SO libraries need to be recompiled whenever we do some changes in Android integration, bumping platform versions, minimal versions, NDK version -- so it has to be as-straightforward-as-possible to update these libraries.

   Please see how https://github.com/castle-engine/castle-engine/tree/master/tools/build-tool/data/android/integrated-services/freetype describes it:

    > Update the library files (SO) like this:
    > git clone https://github.com/castle-engine/android-freetype
    > cd android-freetype/
    > make build

    This is accompanied by https://github.com/castle-engine/android-freetype , our freetype fork, and https://github.com/castle-engine/android-freetype/blob/master/Makefile that automates this process: builds and copies the libraries.

2. We need libraries also for x86 and x86_64 Android versions, useful in particular with Android emulators that often have x86 / x86_64 CPU (to make them run fast on regular PCs). See all 4 platforms of freetype: https://github.com/castle-engine/castle-engine/tree/master/tools/build-tool/data/android/integrated-services/freetype/app/src/main/jniLibs

michaliskambi (changes requested):
Thank you for contributing! 

I'm adding some comments below about things that need addressing before we can merge this. The most important one is _"We need to extend these instructions, to make them trivial to repeat and allow everyone to recompile all SO files for Android..."_. Let me know if my guides there are not enough, I can extend about what is required and how to go about this.

tools/build-tool/data/android/integrated-services/effekseer/README.adoc
PTz0uAH: I prepped an android dev environment (WSL Debian 12) and the Freetype build procedure is now clear to me..

It produced the binaries for the 4 Android targets in no time..

Compiling Effekseer is somewhat a different league..

It looks a bit more complex.. and the build process differs hugely from the Freetype-build-procedure..

but I'll give it a shot..

PTz0uAH: Some progress on building libeffekseer on WSL/Debian12..

Yesterday I installed cmake, cloned https://github.com/castle-engine/Effekseer followed the steps from Effekseer- HowToBuild.. created the `Makefile` with `cmake -S . -B .` and performed an oldschool `make` (all) which resulted in a wrapper/libeffekseer.so binary.. I guess the same result would be achieved with a `cmake --build .`

Next step.. crosscompiling for Android should be a "piece of cake" right?.. (as the Android sdk functions well with the freetype build procedure).. hmm.. missing some cmake files needed for crossbuilding..

Clearly now I have to create 4 `cmake toolchain files` for the 4 Android targets.. or maybe 1 to suit all..

Have to do some studying now and take notes on behalf of a short android-libeffekseer-cge-howto-help file..  and figure out how cmake works with the Android sdk..

I'll get back to you when that particular information has been processed/ assimilated.. lets hope there aren't any caveats or oddities..

michaliskambi: Thank you!!

PR #608, opened 2024-05-20
Windows Drop File Code Checkpoint, Using WM_DROPFILES Message (not OLE).

coderextreme:
Since this is my first contribution, I'd thought I'd give people a chance to review my code before writing an example against it (I have a partial one that brings up a blank window and registers for an event, but since OnDropFiles is not assigned, DoDropFiles does not do anything (You can check this in the debugger that DoDropFiles is called).  This has not been called with several files yet, but the code could possibly work. I don't know about 0 indexing versus 1 indexing into the FileNames array.  I tried 1 indexing.

I couldn't seem to get Any WideChar stuff working yet.  AFAIK, DoDropFiles takes an array of AnsiString.

I am not expecting this pull request to get merged yet, but I would like the code reviewed (It's quite short), and possibly make sure stuff still compiles on other systems.

I don't expect any gotchas in this code except perhaps I need to add a 1 to a FileNameLength here and there.


src/window/castlewindow.pas
```diff
@@ -957,6 +957,14 @@   TCastleWindow = class(TComponent)
     procedure UpdateRequirements;
   protected
     procedure DoUpdate; virtual;
+    { procedures for dropping files }
```
michaliskambi, line 960: Move these lines into `src/window/windows/castlewindow_winapi.inc`, place them between

```
{$ifdef read_window_interface}
...
{$endif read_application_interface}
````

This will achieve the same effect (the methods will be part of `TCastleWindow` class), but will keep nicely everything in `src/window/windows/castlewindow_winapi.inc`. So you will not need to surround it with `{$ifdef CASTLE_WINDOW_WINAPI}...{$endif}`.

File: src/window/windows/castlewindow_winapi.inc [hunk 2]
```diff
@@ -199,6 +199,60 @@ const
   VK_OEM_8 = $DF;
   VK_OEM_102 = $E2;
 
+function TCastleWindow.WMDropFiles(var DropH: HDROP) : LRESULT;
+
```
michaliskambi, line 203: Remove empty line.

src/window/windows/castlewindow_winapi.inc
```diff
@@ -199,6 +199,60 @@ const
...
+  
+var
+  DroppedFileCount: Integer;  // number of files dropped
+  FileNameLength: Integer;    // length of a dropped file name
+  FileName: string;           // a dropped file name
+  I: Integer;                 // loops thru all dropped files
+  // DropPoint: TPoint;       // point where files dropped
+  FileNames: array of AnsiString;
+begin
+  inherited;
```
michaliskambi, line 212: Remove `inherited` from this -- there is no `WMDropFiles` implemented in ancestor (and this is not even virtual method), so `inherited` would do nothing (and could confuse devs looking at it :) ).

src/window/windows/castlewindow_winapi.inc
```diff
@@ -199,6 +199,60 @@ const
...
+      // set the length of the buffer
+      SetLength(FileName, FileNameLength);
+      // get the filename
+      DragQueryFile(DropH, I, PChar(FileName), FileNameLength + 1);
+      // set the index in the filename array to the file name
+      FileNames[I] := AnsiString(FileName);
+    end;
+    // DropPoint is where files were dropped
+    //DragQueryPoint(DropH, DropPoint);
+    Self.DoDropFiles(FileNames);
```
michaliskambi, line 232: Do not write explicit `Self.`, it is not necessary in Pascal to qualify method calls with this. So we generally not write `Self.Xxx`, just `Xxx`.

src/window/windows/castlewindow_winapi.inc
```diff
@@ -326,6 +380,8 @@ begin
     SetPixelFormat_WGLChoose also get here (as they use our
     Application.wndClassName). They don't set Get/SetWindowLongPtr, so Window here
     will be @nil in this case. }
+  if uMsg = WM_DROPFILES then
```
michaliskambi, line 383: You added this to `WndProc`. This will work, but there's an easier and better way: add handling of `WM_DROPFILES` to `TCastleWindow.WindowProc`. There's a big `case uMsg of` there, where you can add handling of `WM_DROPFILES`. 

This is better, as we already handle most message in `TCastleWindow.WindowProc`, and we check is `Window` valid along the way.

So, please move implementation of this `TCastleWindow.WindowProc`.

michaliskambi (changes requested):
Thank you very much! Reviewed, I have small notes -- see below. Only minor stuff, I think the biggest change will be to move handling to `TCastleWindow.WindowProc`.

michaliskambi:
Note that GitHub Actions detected failure at Delphi compilation -- `Error: E2138 Invalid message parameter list`.

You can leave investigating + solving it on me later. If you'll make it work with FPC, I'll later fix Delphi, no problem :)

michaliskambi:
To answer some things in PR description:

> I don't know about 0 indexing versus 1 indexing into the FileNames array. I tried 1 indexing.

In Pascal, strings (`string`, `AnsiString`, `WideString`, `UnicodeString` etc.) are indexed from 1. Everything else is indexed from 0. In particular, `array of String` is indexed from 0.

> I couldn't seem to get Any WideChar stuff working yet. AFAIK, DoDropFiles takes an array of AnsiString.

In case of FPC (open-source compiler that you use), this is true.

The full picture is a bit more complicated as we also support Delphi, where `String` equals `UnicodeString` and all characters are 16-bit (it matches most the "wide strings" as understood by WinAPI, even better than `WideString`). This is documented on https://castle-engine.io/coding_conventions#strings_unicode . As I mentioned in previous comment, you can ignore Delphi compatibility, I'm prepared to later investigate / fix this on my side  :)

michaliskambi:
Question: You mention a few times (in the title of this too) that it's "not OLE". If you have please post here some links and arguments why we'd like to go with a different implementation, using OLE? Because I quite like to current approach (WM_DROPFILES), this is simple and we like simple code :)

src/window/castlewindow.pas
coderextreme: Okay, i will try that. Not having the $endif match the $ifdef is confusing

coderextreme: I made the

{$ifdef read_window_interface}

match the $endif.  I saw one other place that had this.
{$endif read_window_interface}

coderextreme: I don't see any problem with replacing read_application_interface with read_window_interface. Resolved, I hope.

src/window/windows/castlewindow_winapi.inc
coderextreme: Ok.  This should be complete.

coderextreme:
I couldn't test fully due to MimeType type changes blocking simple loads of files without a mimeType.  I will look into creating a stream, but just loading a file without a stream (just a URL and a MimeType, or just a URL might be simpler).

michaliskambi:
> I couldn't test fully due to MimeType type changes blocking simple loads of files without a mimeType. I will look into creating a stream, but just loading a file without a stream (just a URL and a MimeType, or just a URL might be simpler).

As we discuss on Discord, this is likely a bug on your side -- your string handling adds a trailing `#0`.

There are no "MimeType changes" lately that change the relevant logic -- `LoadNode(Url)` detects file type based on URL (through extension->MIME type mapping), and it fails if it doesn't match any known type. 

And again, please keep the discussion in one place. It is now in 4 places (one here, 3 on Discord), which means I have to answer 4 times to keep everyone who's reading aware of what's going on :) Let's continue on `#general` on Discord and post here conclusions. In the future, please keep talking about PR contained to the GitHub comments about this PR.

coderextreme:
Pascal doesn't like the NUL at the end of the Windows string so we chop it off!

coderextreme:
Yes, the extra space is for Windows API to fill in a NUL, otherwise it will
chop off ‘d’ from ‘.x3d’. I tried several combinations, but settled on
trimming the NUL after Windows finished with the string.  This works.

We’ll probably want to discuss trying to figure out how to reclaim the
memory for the file names and array.  It could consume a lot with a lot of
files dropped .  After DoDropFiles finishes in WMDropFiles, there’s an
opportunity to free up the space for the filenames and the array.  I don’t
know how to do that yet.

Also, the client needs to be informed in documentation that that
OnDropFiles gives them should not be freed, but we don’t have much control
from the calling side.

John

coderextreme:
Here’s my test code previously mentioned.

I’ll work on freeing up the array of strings allocated in WMDropFiles in WMDropFiles.

coderextreme:
> Here’s my test code previously mentioned.
> 
> https://github.com/coderextreme/draganddraop
> 
> I’ll work on freeing up the array of strings allocated in WMDropFiles in WMDropFiles. My first guess is to use SetLength and use a length of zero. I also don’t know what calling SetLength on an array does. More stuff to look up

okay, Calling SetLength(var, 0) seems to be the way to go for both arrays and strings.

I am now realizing I am probably overwriting the file name when I call SetLength for a second time.  I will try  more two file drops to make sure things are ok.

I know the scene will get overwritten in the demo.  I am trying to follow the interface.

coderextreme:
All set, except I don't know how to free up memory held by PFileName.  AIRC, I can't use SetLength.  So, a puzzle.

michaliskambi:
You don't need to explicitly free memory allocated for Pascal `string` or dynamic `array of string`. Please remove the lines 

```
 for I := 0 to Pred(DroppedFileCount) do
    begin
      SetLength(FileNames[I], 0);
    end;
    SetLength(FileNames, 0);
```

, they are not useful and somewhat confusing :) 

The Pascal `string` (whether `AnsiString` or `UnicodeString`, depending on compiler, https://castle-engine.io/coding_conventions#strings_unicode ) is ref-counted, has copy-on-write, and in effect the allocated memory is automatically managed and you don't need to worry about it.

The Pascal dynamic array, like `array of xxx`, is also automatically managed (though without copy-on-write). So you also don't need to worry about it.

Test your application for memory leaks following https://castle-engine.io/memory_leaks to see if you have real problem with memory leaking and where.

coderextreme:
We discuss how to add

```
Window.RegisterDropTarget;
```
to applications in a portable manner.

Thanks!

coderextreme:
Seems like things have settled down.  I've got another demo working with the 3D viewer project template, even dragging and dropping and playing the animation works. I'm still working out how to organize things.  If you have any further ideas on portability, let me know.  I haven't checked for memory leaks yet.

michaliskambi:
TODOs for me :

0. Review, apply :)

1. Decide do we want to make 

    ```delphi
    procedure UnregisterDropTarget;
    procedure RegisterDropTarget;
    ```
    
    available on all platforms, and just doing nothing on non-Windows?  I think yes.

2. Test and fix Delphi compilation / working ( Delphi `string` is something else, 16-bit `UnicodeString`, https://castle-engine.io/coding_conventions#strings_unicode ; it usually doesn't matter... but it matters for code like this, where WinAPI assumes certain string encoding ). 

    Hm, we may want to just use `UnicodeString` for both FPC and Delphi actually, and use the WinAPI functions with `W` suffix, to correctly get filenames with local characters (e.g. Polish "ogonki" encoded). This is also what we did with clipboard, with TWinApiClipboard -- just using `UnicodeString` for both FPC and Delphi was simplest to solve all the issues. To call user `DoDropFiles`, we'd just convert them using `Utf16ToString`.

3. There's a nice doc from John, incorporate it as needed to API docs, https://coderextreme.net/HowTo.html .

4. Use this in castle-model-viewer.

PR #612, opened 2024-05-28
Base terrain editing tools for Castle Engine Editor

and3md:
This PR add  basic TCastleTerrain edit suport for our editor:

obraz

With it you can:
- raise terrain 
- lower terrain
- level terrain
- change heightmap size
- save height map
- save height map with other name

Raise/lower/level terrain operations supports many brushes shapes like cone, pyramid, square, circle, ring etc.
You can set many parameters like strength, size, rotation etc.

See the video:

[terrain_edit.webm](https://github.com/castle-engine/castle-engine/assets/18555708/556a12df-1183-4074-a425-6edbbd84ef18)

Most of the code is in `TCastleTerrainEditMode` class.

There are two way to modify terrain height. The faster one is not working in TCastleControl so it is comneted in `AlterTerrain` function.

It can be uncomented using this example code:
https://github.com/and3md/castle-engine-terrain-editor-example


and3md:
I don't know why, but the video doesn't work for me in Firefox. Works on Chrome-based browsers

PR #616, opened 2024-06-28
Add Priority to TCastlePlayingSound and TCastlePlayingSoundSource.

### LINKED ISSUE #615: TCastlePlayingSound and TCastlePlayingSoundSource are missing the Priority property
While TCastlePlayingSound and TCastlePlayingSoundSource both have Pitch and Volume properties, the Priority property is missing.
That Priority would be multiplied by the TCastleSound.Priority to calculate the "final priority".

Which means without Priority you are bound to the TCastleSound.Priority - thus you can't share that single object between different "priority"-groups (like the same sound for Player and NPCs - but with different priority).

Dennis1000:
Add the missing Priority-Property to TCastlePlayingSound and TCastlePlayingSoundSource.
Resolves #615

### DIFF
--- a/src/audio/castlesoundengine_engine.inc
+++ b/src/audio/castlesoundengine_engine.inc
@@ -568,7 +568,9 @@ begin
       Result.Loop    := Parameters.Loop;
       Result.Pitch   := Parameters.Pitch;
       Result.Volume  := Parameters.Volume;
-      Result.Priority := Parameters.Priority;
+      { No need to set Result.Priority,
+        TSoundAllocator.AllocateSound have already done this. }
+      Assert(Result.Priority = Parameters.Priority);
       {$warnings off} // using unportable symbols knowingly
       Result.MinGain := Parameters.MinGain;
       Result.MaxGain := Parameters.MaxGain;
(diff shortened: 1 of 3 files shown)


michaliskambi (approved):
Thank you! Merging. 

I did one meaningful change in your code in https://github.com/castle-engine/castle-engine/pull/616/commits/66f0ff20884d9a2edac6279e35feb3c3ebd5d3c5 . The reason was: I got curious *why* did you had to set `Result.Priority := Parameters.Priority`  in `TSoundEngine.PlaySound`, did we maybe had a bug previously and we didn't save priority?

It looked like a bug on 1st sight -- we set a lot of `Result` fields, but not `Result.Priority`. But upon investigation it turns out that the old code was correct, but in non-obvious way: `AllocateSound(Parameters.Priority)` actually already made sure the `Result.Priority = Parameters.Priority`.

So I removed `Result.Priority := Parameters.Priority` assignment, but added `Assert(Result.Priority = Parameters.Priority);`, so that the code is obvious now :)

Dennis1000:
You're right. Wasn't that obvious to me - my change there looked good to me :-)

Issue #598, opened 2024-04-03
Failed to detect VSCode on MacOS


caryyu:
# Ticket

https://discord.com/channels/389676745957310465/1105477501330931712/1225052612337078332

# Description

Currently, the detection only works on the env-var, however, if it can support `/Application/xxx` detection, that should make things easier. 

![image](https://github.com/castle-engine/castle-engine/assets/4496641/5e18f1db-3a67-4705-b224-982262d4ec7f)

michaliskambi:
Fixed by @janadamec1 in https://github.com/castle-engine/castle-engine/pull/622 - thank you!

PR #622, opened 2024-08-12
Autodetect VS Code on macOS for the Castle Editor

janadamec1:
This will add auto-detection of VS Code on macOS. By default, VS Code is in Applications, and to call it with parameters in RunCommandNoWait, it is OK to use the executable from within the Visual Studio Code.app (the executable is called Electron).

This will fix the Issue #598


michaliskambi (approved):
Thank you! 

I changed the code to use a constant (`MacVSCodePath = '/Applications/Visual Studio Code.app/Contents/MacOS/Electron'`) -- more obvious that it's the same path, and eliminates the possibility of someone making a typo in one path :)

I'll let GitHub Actions make checks and auto-merge it.

PR #623, opened 2024-08-15
Add scene saving function to deprecated library interface

janadamec1:
It is handy to have the function to save scene to file (X3D, STL) in the (deprecated) library interface.

I've also added command to force modern rendering (GLFeatures.RequestCapabilities := rcForceModern;) for macOS, which is exactly on par how CastleControl and Cocoa backend work on macOS after requesting OpenGL CoreProfile.

Thanks.

### DIFF
--- a/src/deprecated_library/castleengine.h
+++ b/src/deprecated_library/castleengine.h
@@ -1,5 +1,5 @@
 /*
-  Copyright 2013-2023 Jan Adamec, Michalis Kamburelis.
+  Copyright 2013-2024 Jan Adamec, Michalis Kamburelis.
 
   This file is part of "Castle Game Engine".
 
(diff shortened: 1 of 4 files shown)


michaliskambi:
> I've also added command to force modern rendering (GLFeatures.RequestCapabilities := rcForceModern;) for macOS, which is exactly on par how CastleControl and Cocoa backend work on macOS after requesting OpenGL CoreProfile.

When is this useful? What do you want to achieve with this?

To explain, `TGLFeatures.RequestCapabilities := rcForceModern` only influences how the context is created by CGE routines. And actually our TCastleWindow on Cocoa ignores it, always creating `NSOpenGLProfileVersion3_2Core`.

The `TGLFeatures.RequestCapabilities` does not change how CGE functions. When we get a modern context, we used only modern context features.

So, setting `TGLFeatures.RequestCapabilities` for `src/deprecated_library/castleengine.lpr` doesn't seem useful -- it really doesn't do anything. In case of using `src/deprecated_library/castleengine.lpr`, it is the responsibility of the application using the library to initialize the OpenGL(ES) context, and prefer a modern context, and force "core" context on macOS.

janadamec1:
Sorry, I misunderstood what this variable means. I will remove this part of code.

michaliskambi:
> Sorry, I misunderstood what this variable means. I will remove this part of code.

Thank you and merging then!

PR #624, opened 2024-08-16
Cocoa backend fixes

janadamec1:
This PR implements clipboard on macOS (for example Ctrl+C didn't work in Castle Model Viewer on macOS), and also fixes the issue with retina monitor scaling with the current macOS SDK.

The code also reflects to dragging window to another screen (notification was not implemented before) and detects size changes by overriding the setFrameSize instead of NSViewGlobalFrameDidChangeNotification, which came far too often.


michaliskambi:
Thank you enormously for pursuing this -- in particular the monitor scaling was an important issue!

I checked the code, I checked that it works without regressions on my non-Retina monitor (plugged to remote macOS in MacStadium), I checked that clipboard now works. All great.

PR #625, opened 2024-08-16
Fix triangulation output when the shape is not CCW.

janadamec1:
When saving scene to STL (which uses triangulation), all triangles are expected to be CCW, but when the shape is not CCW, exported STL file is incorrect. See the test scene attached
[3DScene.x3d.txt](https://github.com/user-attachments/files/16635677/3DScene.x3d.txt)

To reproduce: open the attached file (after removing .txt file name suffix) in current Castle Model Viewer, and try to save it as STL. The IndexedFaceSet with ccw set to false is not exported correctly. After the fix, it is OK.

BTW, this may also fix a problem with shadow volumes I am getting on some scenes (I can see shadows only from some shapes). I will re-test shadows on my scenes after this change is merged into the master branch.


michaliskambi:
Thank you! Applied the fix, albeit with one important modification:

The `[Local]Triangulate` methods get a parameter `FrontFaceAlwaysCcw`. The default is `false` which means the behavior is as before. The value `true` changes the behavior to what you proposed, i.e. "front is always CCW".

I also documented why it matters at `[Local]Triangulate` API.

Reason for my change: I'm a coward :), basically. 

That is: I looked at CGE code doing `[Local]Triangulate`, and in the past I didn't think about this fact (that front may be CCW or CW). There *may* be some undocumented assumption somewhere that `[Local]Triangulate` behave as before. I didn't *find* such place, though. But there may be.

I actually found 2 more places where your new behavior ("always CCW") is desired - one-sided TCastleMeshCollider, and test_triangulate example. So, in total there are at least 3 reasons for it (STL output, one-sided TCastleMeshCollider, test_triangulate example). 

Maybe the shadows are the 4th reason (feel free to test, you will need to modify CGE -- toggle it at this call, https://github.com/castle-engine/castle-engine/blob/1f75fc95efa5751dbb75d768a79607d3a64722c0/src/scene/castleshapeinternalshadowvolumes.pas#L240 ). I think toggling this to `true` may make sense there, in case you have 2 shapes, one with front CW, the other with front CCW. In order to detect them as 2-manifold correctly the code may need the same order, so `FrontFaceAlwaysCcw` = `true` may be justified. So, you may be right, but I'm still scared of just changing `[Local]Triangulate` for all uses. So if shadows need it, then let's pass `FrontFaceAlwaysCcw` = `true` in that call on https://github.com/castle-engine/castle-engine/blob/1f75fc95efa5751dbb75d768a79607d3a64722c0/src/scene/castleshapeinternalshadowvolumes.pas#L240 , along the comment explaining why it matters.

PR #628, opened 2024-08-27
Unicode menu items

janadamec1:
This fixes the issue 100 in Castle Model Viewer (https://github.com/castle-engine/castle-model-viewer/issues/100), where UTF8 strings (e.g. viewpoints names, file names) with non-ASCII characters did not show correctly in menu items. 

Fixes:
1. WinAPI menu items (for viewpoint names and recent file names)
2. Cocoa menu items (for viewpoint names and recent file names)
3. Cocoa window title (for currently opened file name)

Not sure if NSString.stringWithUTF8String leaks or not, seems similar to NSSTR inline function (which calls CFStringMakeConstantString). I did a test in Activity monitor, did not see anything, so it should be probably fine or not so important.


michaliskambi (approved):
Thank you!

Issue #596, opened 2024-04-03
CGE editor as universal binary on macOS -- with both x86_64 and Aarch64 (Arm64) versions


michaliskambi:
And it should build either universal binaries, or at least for current platform -- so on M1 it will build binary for Aarch64.

janadamec1:
Implementation is being done in GIT branch "build-tool_macosarm64"

michaliskambi:
I just [announced a lot of macOS improvements](https://castle-engine.io/wp/2026/02/02/macos-improvements-work-flawlessly-on-apple-silicon-aarch64-bundled-with-fpc-signed-and-notarized-cocoa-mouse-positioning-fixes-to-ondropfiles-universal-binary-support/), in particular we have engine version that "just works" out of the box on Apple Silicon (macOS/Aarch64), with everything (editor, build tool, FPC) provided for _macOS/Aarch64_.

This is *not* realized by making CGE editor as universal binary, as this issue was saying to do. Instead, we just have separate engine downloads, one for _macOS/Aarch64_, one for _macOS/x86_64_. Of course either of them can be used to make universal binaries by users (they only need to setup FPC with cross-compiler to the other platform).

...And I think this existing solution, while it's not "editor is universal binary in engine macOS downloads", is good enough :) I'm closing this issue and dropping the idea of "editor as a universal binary on macOS", because:

- Current solution works :) Users just need to choose their editor version when [downloading the engine](https://castle-engine.io/download), we just show 2 possible engine downloads for mac. 

- Making a single download with universal binary means a bit more work. For one hand, we'd need to extend our [castle-fpc](https://github.com/castle-engine/castle-fpc) to provide cross-compilers in both directions. Practically, it's also a problem for me -- I don't have access to _macOS/x86_64_ for testing anymore, only _macOS/Aarch64_.

- ...and it may be an unnecessary work, as most Apple users are now on Silicon architecture from what I hear when talking to people, and Apple is explicit that _macOS/Intel_ is going away (macOS versions after [Tahoe](https://en.wikipedia.org/wiki/MacOS_Tahoe) are not going to have Intel support). I think this means that putting more work to account for macOS/Intel, when it's going away, is not worth it :) We allow users to build their own applications as universal binaries, and this may be enough.

PR #629, opened 2024-08-31
Build tool can produce universal binaries on macOS (containing x86_64 and arm64 slices)

janadamec1:
This PR implements issue #596 
It adds two possibilities to make a universal binary on macOS.

At first, we wanted to have this optional, as making two slices in the binary doubles the compile time and file size. So when nothing is specified, the BuildTool works as before, making a binary according to other parameters or current platform.

I used FPC cross-compiler provided to download at the [Lazarus SourceForge site](https://sourceforge.net/projects/lazarus/files/Lazarus%20macOS%20x86-64/Lazarus%203.4/).

### 1. Command line argument for target
This PR adds new target 'macos'.
`castle-engine simple-compile --target=macos ....`

When specified, the build tool makes the universal binary. So in case we have one '.sh' compile script to make classic Linux, (Windows?,) and simple macOS binary, there needs to be a special command for "universal" macOS binary, probably causing to have a dedicated script file for this case.

Used and **tested in src/deprecated_library**/castleengine_compile_macos.sh. 

### 2. CastleEngineManifest.XML option
In case the project is defined using CastleEngineManifest.XML there could be a simpler way. So we added a `mac_universal_binary` (default: false) argument for <project> tag, next to existing `mac_app_bundle` (default: true).

When specified, build tool makes the universal binary on macOS. And you can keep only one simple compile script for all platforms.
**Tested in view3dscene**: after adding `mac_universal_binary="true"` into <project>, calling view3dscene/compile.sh makes the universal binary on macOS for Castle Model Viewer.

### Implementation details
We basically need to call compilation twice with parameters needed to produce both slices of universal binary. The process is shown in src/deprecated_library/castleengine_compile_macos_intelarm64.sh, and is implemented in file tools/build-tool/code/toolmacos.pas 

1. compile x86_64 slice. Apple Notarization requires at least SDK 10.9, so we need to set MACOSX_DEPLOYMENT_TARGET for fpc in this slice.
2. rename output binary file, adding .x86_64 filename suffix
3. compile arm64 slice.
4. rename output binary file, adding . aarch64 filename suffix
5. call lipo command to connect both slices together to form the resulting universal binary

To pass MACOSX_DEPLOYMENT_TARGET to the compiler, we have to add two member variables to TCompilerOptions, and pass them in RunCommandIndirPassthrough that already expect these parameters. So this looks pretty straightforward.
CompilerOptions.OverrideEnvironmentName := 'MACOSX_DEPLOYMENT_TARGET';
CompilerOptions.OverrideEnvironmentValue := '10.9.0';

To get the binary file name after compiling the slice was more challenging. I had to add variable `LinkerOutputFile` to TCompilerOptions, that stores the name of the compiled result, and caller can read its value after `Compile(...)` procedure is called. To get the actual filename, I capture the fpc output, and search for line staring with word "Linking". I tried other ways, but only this one got the correct results.

### Possible code improvements
I am not entirely happy with my solution to pass variable `LinkerOutputFile`. Other TCompilerOptions are input parameters for the compiler, while this one is output parameter. However adding a new class only for output parameters makes thing very complicated and affects lots of functions. So, after all, the current solution seems easy to use and read. Maybe we could rename member variable `LinkerOutputFile` to clearly show this is for output


michaliskambi:
Thank you, I will look to review + apply this soon! As we talked in private, this is a feature I very much want, as it makes total sense for macOS users now.

Note: I see GitHub Actions check failed for this PR (you most likely got an email from GitHub about it) -- you can ignore it in this case, that's my fault. They failed because yesterday I made an incompatible change to CGE TCastleViewport that castle-model-viewer relies on -- I should not do this, but I thought it is acceptable because it was supposed to be only very temporary. It is already solved, and next GitHub Actions check shall go OK.

janadamec1:
From failing test on macOS machine without ppcarm64 installed (I suppose it is because deprecated_library has mac_universal_binary="true"), it seems I will have to add a check whether ppcarm is installed...

janadamec1:
As an addition to changes in build-tool, I've also fixed the issues in iOS Xcode project template.

For example, now we always add Launch Storyboard, as it is required by iOS13, else the apps does not cover the entire device screen. So I've added default_launch_images/DefaultLaunchImageCentered.png, and force it when there is no `<storyboard>` defined in CastleManifest.xml. I'm not sure if the placement is OK, since this is iOS only and the code is in ToolProject.pas.

**TODO:** Anyway, this change should be reflected in [the manual too](https://castle-engine.io/project_manifest#_launch_images_for_now_only_for_ios) - Unless the project targets iOS13+, projects need to include both `<launch_images>` and `<storyboard>`.

janadamec1:
BTW tests are failing on EDownloadError: Error when downloading
"castle-data:/default_launch_images/DefaultLaunchImageCentered.png,
because we need the updated build-tool to handle castle-data:/ files
correctly. See the commit
https://github.com/castle-engine/castle-engine/pull/629/commits/30d71a465946c7e4bf096a814396596d66d1678f

At least I hope this is the case. There is an old build-tool that
tests template. Because exactly this error should be solved in the
above commit.

michaliskambi:
> BTW tests are failing on EDownloadError: Error when downloading
> "castle-data:/default_launch_images/DefaultLaunchImageCentered.png,
> because we need the updated build-tool to handle castle-data:/ files
> correctly. See the commit
> https://github.com/castle-engine/castle-engine/commit/30d71a465946c7e4bf096a814396596d66d1678f
> 
> At least I hope this is the case. There is an old build-tool that
> tests template. Because exactly this error should be solved in the
> above commit.

The tests are always performed with the build tool compiled from the latest code -- when running the testsuite, we don't even have the previous build tool available. So, the error in the test suite is something we should fix. 

I didn't analyze the details now, just posting some pointers:

- https://github.com/castle-engine/castle-engine/actions/runs/12119339308/job/33785706869?pr=629 indeed tells that `Failed: Error when downloading "castle-data:/default_launch_images/DefaultLaunchImageCentered.png"`. 

- It happens when running the testsuite of the build tool, it tests whether some build tool code (like macro expansion) works OK. 

- I can reproduce the fail on a Linux, by checking out the branch with this PR, then

```
# rebuild the build tool
cd tools/build-tool/
./castle-engine_compile.sh  # calls fpc to make castle-engine
mv castle-engine ../../bin/ # place castle-engine on $PATH -- your way of doing this may vary

cd tests/ # effectively cd <castle-engine-path>/tools/build-tool/tests/
# run the build tool testsuite, which is also a regular CGE project we build and run using the build tool:)
castle-engine compile
castle-engine run
```

- On a quick glance, your code in `AddMacrosIOS` assumes that `castle-data` is the data of the build tool, but our test suite breaks it. This would say it's an error in the test suite -- maybe `tools/build-tool/tests/test_cge_build_tool.lpr` needs to adjust the data path, just like `AdjustApplicationData` does from https://github.com/castle-engine/castle-engine/blob/2048f5789d679661cbaf4a2024420a675adccad9/tools/build-tool/castle_engine.dpr#L325 ? It may be enough to just copy-paste `AdjustApplicationData` into `test_cge_build_tool.lpr` :) But I didn't test!

michaliskambi:
Sorry for taking ages to review this! Below is just my quick initial review, with some TODOs that I also for me (not only for you) to recheck. Sorry for raw and possibly incorrect thoughts from my side, but I know you'd like to eventually have this PR applied (and me too!) so I'm posting this "raw review". Sorry in advance if I just wrote something invalid, discussion is welcome as always :)

1. Do we need both `mac_universal_binary` in manifest _and_ new `--target=macos`? Their interaction is not obvious to user (and the interaction with `build_using_lazbuild` is not obvious) and it seems one of them should be enough. My initial thought is to remove `mac_universal_binary` from manifest. Let this be decided by just `--target=macos`.  Just like `--target=android` builds for all Android CPUs.

2. Test and make sure everything, including `package` and `run`, still works with `--target=macos`.

3. Test how does this affect our ability to make universal macOS build of the editor, which is done by `lazbuild` compilation. It's OK if this is unrelated, I just need to understand.

4. In CompileMacOS there are 2x cases of `try .. except` that catch all exceptions without capturing any details about the exception they catch. They should A. at least catch exception like `except on E: Exception do...` and show it's class + message by `ExceptMessage(E)` and show it using `WritelnWarning`, not `Writeln`. B. Or, possibly, they should just let the exception break the flow -- universal binary cannot be created.

    My initial thought is to go for B, simpler and makes sense -- we failed to make a binary with 2 architetures, but needs to be tested.

5. It adds LSSupportsOpeningDocumentsInPlace, which I need to understand the consequences.

janadamec1:
> 1. Do we need both `mac_universal_binary` in manifest _and_ new `--target=macos`? Their interaction is not obvious to user (and the interaction with `build_using_lazbuild` is not obvious) and it seems one of them should be enough. My initial thought is to remove `mac_universal_binary` from manifest. Let this be decided by just `--target=macos`.  Just like `--target=android` builds for all Android CPUs.

I started with `--target=macos`. But I think the manifest option is super handy, as then there could be just one build-tool command line (or script) for all platforms, with no need to have special parameter on macOS. For me, the manifest option is much better way to create Universal binary on macOS, than to remember the command line option next time I build the project for macOS.

BTW I always forget the command line arguments for build tool, and I need to look it up in docs. Having this option in manifest saves some frustration.

> 3. Test how does this affect our ability to make universal macOS build of the editor, which is done by `lazbuild` compilation. It's OK if this is unrelated, I just need to understand.

I have to admit, I tested everything just with fpc, not LazBuild. So compiling Castle Model Viewer works (just by adding the manifest option), I will test the CastleEditor too.

> 4. In CompileMacOS there are 2x cases of `try .. except` that catch all exceptions without capturing any details about the exception they catch. They should A. at least catch exception like `except on E: Exception do...` and show it's class + message by `ExceptMessage(E)` and show it using `WritelnWarning`, not `Writeln`. B. Or, possibly, they should just let the exception break the flow -- universal binary cannot be created.
>    My initial thought is to go for B, simpler and makes sense -- we failed to make a binary with 2 architetures, but needs to be tested.

I made some improvements to report the exception (case A). This is for cases the compile script wants to build the universal binary (using manifest option), but the machine does not have the arm64 compiler installed. Should be stop the build and force user to either install the missing compiler, or remove the manifest option in our code?

I understand this might be confusing. I made this primarily because there is no arm64 on your build server, and thus the project did not compile at all, and Checks did not result with success.

> 5. It adds LSSupportsOpeningDocumentsInPlace, which I need to understand the consequences.

Basically, when this is not present, user has hard time opening the files on iOS in our app (Castle Model Viewer Mobile). iOS has not very intuitive way to open it, maybe to suggest the file will be copied to the app, not opened directly.

When adding this option, it is safe to use the file for reading, but one cannot write into it, or remember the path. For this, the app should use bookmarks for the URL. If you don't like it to have this always on, maybe we could make an option to manifest.

michaliskambi:
> But I think the manifest option is super handy, as then there could be just one build-tool command line (or script) for all platforms, with no need to have special parameter on macOS. For me, the manifest option is much better way to create Universal binary on macOS, than to remember the command line option next time I build the project for macOS.

To address "easier to remember", we already have a feature to set the default platform to build: https://castle-engine.io/web#platform_default (documented now as part of "web", but its actually a generally useful feature). We added it around January for the web :) So:

- You can set given platform (which can be OS/CPU combination, or "target" like `targetWeb`, `targetAndroid`, (after this PR) `targetMacOS`) as default for the project. 

    You can do this by GUI, in editor "Run → Platforms → …​" and then "Run → Platforms → Save Chosen Platform As Default (in Project Local Settings)".

    Or you can do this by just creating a special JSON file in the project root, see https://castle-engine.io/web#platform_default for example. The format is guaranteed.

- And then all usage, from editor, or from command-line, like `castle-engine compile`, `castle-engine run` ... uses the given platform.

    So it's a _bit_ like your `mac_universal_binary` option does, but more generic. It just changes the default platform used to build (which remains the "current OS/CPU") and it can be overridden by command-line options on the user side.

- And the `CastleProjectLocalSettings.json` is deliberately a separate, unversioned (in default `.gitignore`) file. So this remains a user-specific and project-specific setting, not recoded in `CastleEngineManifest.xml`.

If you want to just make, by default, a universal binary, you can use this. It doesn't just change "what it means to build for macOS", it changes the default platform.

And for building from scripts, I'd say you should just pass the desired platform explicitly. That's what we do in our CI scripts always too. So we have simple rule:

- If you don't specify a `--os` / `--cpu` / `--target` option, we use current OS/CPU (and this can be overridden by `CastleProjectLocalSettings.json` option, but for CI this file should not be present).

- Otherwise, we follow literally whatever you specify in `--os` / `--cpu` / `--target` option.

Hm, I guess I convinced myself a bit more to remove the `mac_universal_binary` from the manifest. Let's leave this decision (what OS/CPU combination to build) to existing mechanisms. If someone wants to make universal binary, this should be indicated by `--target=macos` (or `CastleProjectLocalSettings.json`).

> I made some improvements to report the exception (case A). This is for cases the compile script wants to build the universal binary (using manifest option), but the machine does not have the arm64 compiler installed. Should be stop the build and force user to either install the missing compiler, or remove the manifest option in our code?

Let's stop the build, force user to install missing compiler. So, I guess -> let the exception just propagate.

This makes sense, if I (in light of previous paragraph) prefer to also remove `mac_universal_binary` from the manifest. If someone requested `--target=macos`, but we cannot do this (because one compiler is not available) -> let's clearly report the error and fail. This is also good for CI, where you don't want to silently make a "less capable" build without all platforms. 

And by default, since by default we use current OS/CPU, this should not cause issues. If someone has compiler for current OS/CPU, then it will just work.

> LSSupportsOpeningDocumentsInPlace...
> If you don't like it to have this always on, maybe we could make an option to manifest.

Thank you! Hm, I don't have an opinion about this yet :) I will need even more information / testing. We change it for all applications, and 

- typical games don't open external files at all (just load their internal data) -- can this flag have any effect on them?
- for utilities (like Castle Model Viewer Mobile) that open files, is there any drawback? You mention "one cannot write into it, or remember the path" -- but I understand that without this option, we have a temporary file available, so actually also writing into it would be pointless (even if allowed)?

janadamec1:
OK, I will remove the manifest option, which will also enable failing when arm64 compiler not found.
I will also skip LazBuild variant.

Re LSSupportsOpeningDocumentsInPlace, when app does not associate any file types, this option is not relevant. So if you like, I might set to true only when there are some filetypes to associate.

michaliskambi:
> Re LSSupportsOpeningDocumentsInPlace, when app does not associate any file types, this option is not relevant. So if you like, I might set to true only when there are some filetypes to associate.

Yes, exactly, let's do this -- use LSSupportsOpeningDocumentsInPlace only if `<associate_document_types>` is non-empty in the manifest. I think this practically removes all my worries, because the "impact" of the change is then only on apps using `<associate_document_types>` and not on most "regular" games. So _just in case it does something_ (e.g. implies some extra iOS permission or such), it will not affect most CGE games.

Oh, I see you committed this already, cool.

I think that my points 1-5 from https://github.com/castle-engine/castle-engine/pull/629#issuecomment-2881006006 are now practically solved, thank you very much!

New thing I noticed:

6. Can we solve calculating the `Options.LinkerOutputFile` in a different way than by reading the FPC output and relying on `Linking: xxx<newline>` format? It's not just theoretical, I think this message _can_ look different if one translates the FPC output (which can be done in `~/.fpc.cfg`, so it also affects compilation through build tool), see screenshot (grepping for `exec_i_linking` in FPC translation files).

    If the answer is "no, we cannot" then I guess we can live with it, but let's then make a clear error when condition `if LinkerOutputBinaryPos <> 0 then` fails.

image

janadamec1:
BTW, just re-tested this branch in real life: still works with castle-model-viewer-mobile (works great on iPhone and iPad), deprecated library (universal macos library). TODO: castle-model-viewer (universal macos exe).

I will check point 6 tomorrow, however I spent several hours getting the LinkerOutputName better way last year, with no luck. Will try again.

janadamec1:
I tried finding another way to get the `Options.LinkerOutputFile`, but failed again. There could be a chance if ppaslink.sh file was not automatically deleted by fpc, as there is a ld command line inside, and the output file is after -o.

I thought this may be a minor issue as there is a similar code a few lines above searching FpcOutput for 'Linking ', see 'Fatal: Internal error'.

BTW I also have issues building the larger projects with arm64. Linker fails with some internal errors. It worked before (tested with view3dscene, but it fails now). It may be probably related to Command Line Tools 16.3 issue, discussed in [Lazarus GitLab](https://gitlab.com/freepascal.org/lazarus/lazarus/-/issues/41570). No problem with deprecated library though, so our process works.

michaliskambi (approved):

michaliskambi:
And removed the `build-tool_macosarm64` branch now after merging.

Thank you!!!

PR #631, opened 2024-08-31
Fixed Qt OpenGLWidget framebuffers handling in our Qt library example

janadamec1:
This also fixes most issues with invalid Qt framebuffers passed to CGE on all platforms.


michaliskambi (approved):
Looks good, thank you!

PR #632, opened 2024-09-02
Add function to get Castle Engine version to library interface

janadamec1:
This is just simple adding of one more function library interface.
Thanks a lot for your support.


michaliskambi (approved):
Thank you!

PR #633, opened 2024-09-17
Implemented file filters in file-open/save dialogs Cocoa backend

janadamec1:
Implemented file filters in file-open/save dialogs Cocoa backed is a simple way, without combo box for now. But at least this gives automatic file extension (when user does not give any) in SaveDialog and visual file list filtering.
Lazarus can add file filter combo (as commented out in our code), but the code is very complex.

I suggest to keep this simple solution and wait until NSSavePanel.allowedContentTypes becomes available in FPC and implement file filter is using UTIs.

### DIFF
--- a/src/window/castlewindow_cocoa_filedialog.inc
+++ b/src/window/castlewindow_cocoa_filedialog.inc
@@ -190,9 +190,10 @@ begin
     begin
       FileFilterPattern := Filter.Patterns[J];
       if FileFilterPattern = '*' then
-         AllowOtherFileTypes := true;
-      if FileFilterPattern.Length > 2 then
-         fileTypes.addObject(NSSTR(FileFilterPattern.Substring(2)));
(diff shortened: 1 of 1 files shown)


michaliskambi (approved):
Thank you! Merging.

And I agree with your reasoning in the description, this is nice (simple/maintainable and reliable) solution that we can have until allowedContentTypes is available. The Lazarus solution (commented out `CreateAccessoryView`) is quite complex, if we can get away without it -> good :)

I committed a small code improvement in https://github.com/castle-engine/castle-engine/pull/633/commits/fec074696168d3827ea2e6260de0179090d913e5 , please see there for reasons.

I tested how it looks on actual macOS with castle-model-viewer, this is great, I saw other applications on macOS doing it like this too.

Zrzut ekranu z 2024-09-17 15-16-49
Zrzut ekranu z 2024-09-17 15-17-23

Issue #634, opened 2024-09-17
DPI in examples/mobile/simple_3d_demo does not match an actual DPI on iPhone


janadamec1:
Touch controls on iPhone are too tiny when running examples/mobile/simple_3d_demo on iOS.

![MobileSimpleGame](https://github.com/user-attachments/assets/6e47b4df-f674-4ead-b7b4-f62b0442182f)

The screenshot comes from iPhoneXS

michaliskambi:
Thanks, good catch!

To explain what happens and what we recommend:

This example, as most other examples, isn't supposed to follow the DPI of the device.  We generally do not advise to use `usDpiScale`e, because it is quite complicated -- one needs to account then that text wraps at various widths on various devices. And most custom game UIs are custom and just adjusting them to the screen size by scaling everything is simpler and standard for games. 

So we  advise to use `usEncloseReferenceSize`  (default for new projects, and this is used by most examples including above) or `usEncloseReferenceSizeAutoOrientation` (even better, but requires to check design in 2 versions, so not default).

Indeed `usEncloseReferenceSize` has a problem that stuff in portrait mode has different size than in landscape, and it may be too small -- esp. if something was designed looking at landscape, but then it is displayed in portrait. `usEncloseReferenceSizeAutoOrientation` is better for this. See https://castle-engine.io/apidoc/html/CastleUIControls.html#TUIScaling and https://castle-engine.io/wp/2024/08/02/mobile-android-improvements-safe-borders-new-ui-scaling-accounting-for-auto-rotation-nicer-better-gestures-with-tcastleexaminenavigation-modern-android-photo-services-improved-docs/ about `usEncloseReferenceSizeAutoOrientation`. One can also just disable portrait mode, which isn't unreasonable for games.

Long story short, this example should:

- use EncloseReferenceSizeAutoOrientation in data/CastleSettings.xml
- or, if above requires too much work, force landscape in CastleEngineManifest.xml .

PRs are welcome if you want to experiment with this, or I'll look into it soon :)

janadamec1:
OK, setting EncloseReferenceSizeAutoOrientation in data/CastleSettings.xml solves this. #635 

Still I think the round Touch control in the bottom right corner should be DPI scaled, is it is crucial UI control for user. Still, this is a simple demo, so lets keep it simple.

michaliskambi:
> Still I think the round Touch control in the bottom right corner should be DPI scaled, is it is crucial UI control for user. 

[...]

PR #635, opened 2024-09-17
Setting UIScaling to EncloseReferenceSizeAutoOrientation in simple_3d_demo

janadamec1:
... works much better on iOS
As discussed in issue #634 . This PR solves that issue.


michaliskambi (approved):
Thank you!

PR #638, opened 2024-10-04
Adds SynEdit and SynEditDsgn as required packages for the editor

Kagamma:
The title says it all :)


michaliskambi:
Thank you, merging!

Note: This adds the `SynEdit*` as dependencies only useful for custom components, used by custom project-specific editor builds. But, yeah, this is exactly what we want in this case. In the future, likely CGE editor itself (and/or some components in `castle_editor_components.lpk` will require SynEdit too, but then we'll add this dependency for more places).

michaliskambi (approved):

PR #644, opened 2024-11-25
Set WalkNavigation MouseDragMode to rotate in deprecated library when automatic touch interface is set

janadamec1:
Related to removing ControlMouseDragMode in commit https://github.com/castle-engine/castle-engine/commit/6ff0c154548e361d166a4a9616c99bd0485d00cf

We need to address this in the deprecated library. The AutoNavigationViewport used there, so it seems the sufficient solution is to set WalkNavigation MouseDragMode when CGE_SetUserInterface(true) is called to set the automatic touch interface.

### DIFF
--- a/examples/deprecated_library/cpp_winapi_library_tester/main.cpp
+++ b/examples/deprecated_library/cpp_winapi_library_tester/main.cpp
@@ -382,7 +382,7 @@ bool Init()
         CGE_Initialize(applicationConfigDirectory);
         CGE_Open(ecgeofLog, g_windowWidth, g_windowHeight, 96);
         CGE_SetLibraryCallbackProc(OpenGlLibraryCallback);
-        CGE_SetUserInterface(true);
+        CGE_SetAutoTouchInterface(false);
         //CGE_LoadSceneFromFile("c:\\projects\\humanoid_stand.wrl");
         ShowOpenFileDialog();
         return true;
--- a/examples/deprecated_library/lazarus_library_tester/cge_dynlib_tester_form.pas
+++ b/examples/deprecated_library/lazarus_library_tester/cge_dynlib_tester_form.pas
@@ -144,7 +144,7 @@ begin
   CGE_Initialize(PCChar(PChar(GetAppConfigDir(false))));
   CGE_Open(ecgeofLog, OpenGLControl1.Width, OpenGLControl1.Height, 96);
   CGE_SetLibraryCallbackProc(@OpenGlLibraryCallback);
-  CGE_SetUserInterface(true);
+  CGE_SetAutoTouchInterface(false);
   sFile := 'data/bridge_level/bridge_final.x3dv';
   CGE_LoadSceneFromFile(@sFile[1]);
   UpdateUIAfterOpen;
--- a/examples/deprecated_library/qt_library_tester/glwidget.cpp
+++ b/examples/deprecated_library/qt_library_tester/glwidget.cpp
@@ -118,7 +118,7 @@ void GLWidget::initializeGL()
     QString configDir = QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation);
     CGE_Initialize(configDir.toUtf8());
     CGE_Open(ecgeofLog, width()*dPixRatio, height()*dPixRatio, logicalDpiY());
-    CGE_SetUserInterface(false);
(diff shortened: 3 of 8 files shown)


michaliskambi (changes requested):
It seems more future-proof to introduce another routine, like `CGE_SetMouseDragMode(...)`, to control the `Viewport.InternalWalkNavigation.MouseDragMode` . This could take an enum that allows to use all 3 values of [TMouseDragMode](https://castle-engine.io/apidoc/html/CastleCameras.html#TMouseDragMode).

In effect, just like for CGE Pascal API the ` TCastleTouchNavigation`  and `TCastleWalkNavigation.MouseDragMode` are independent, also in the library the `CGE_SetUserInterface(...)` and `CGE_SetMouseDragMode(...)` would be independent.

michaliskambi:
The new version means that setting `TouchNavigation.AutoWalkTouchInterface` and `Viewport.InternalWalkNavigation.MouseDragMode` happen together. This means you hardcode the logic "which `AutoWalkTouchInterface` value best corresponds to which `MouseDragMode`" (when you do `if (nValue = 2 {ecgetciCtlWalkDragRotate}) ... then`). So this means that library C API (unlike Pascal API) would "combine" the 2 things which we want to keep independent: 

- what touch controls are present (the 0/1/2 gizmos in viewport corners) 

- and what happens when you drag on the viewport.

I say to make it all independent in library C API, just like it is in Pascal API. And rename stuff in C API to be as consistent with Pascal API as possible. Advantages: 

1. This will give the C library user maximum flexibility. For the same reason as we made them independent in Pascal API, they should be independent in library C API.

2. It will make the library code more straightforward, so easier to maintain. One C library "concept" should control the corresponding one Pascal API concept. The enums, routines exposed by library should match as closely as possible the Pascal API, to keep things simple.

Disadvantage: I realize this means we break C API backward-compatibility. I would say we should do this -- we need to have things consistent to maintain the library easily. You are likely the only user of this API right now, so I hope you can adjust your application. 

So

1. One thing should control `TouchNavigation.AutoWalkTouchInterface` (and nothing more).

    Maybe this should be `CGE_SetVariableInt(ecgevarWalkTouchCtl)` or maybe separate API call `CGE_SetAutoWalkTouchInterface`. I admit I prefer more new `CGE_SetAutoWalkTouchInterface`, not extending the `CGE_SetVariableInt`, but I'm fine either way.

    And maybe rename `ecgevarWalkTouchCtl` to `ecgevarAutoWalkTouchInterface`. Make it obvious what C API concept corresponds to what Pascal API concept.
   
    The allowed enums should correspond, in names, to allowed enums of Pascal. We have this in Pascal:

    ```
    TTouchInterface = (
      tiNone,
      tiWalk,
      tiWalkRotate,
      tiFlyWalk,
      tiPan
    );
    ```

    So I would say to rename the C enums from

    ```
    enum ECgeTouchCtlInterface
    {
      ecgetciNone              = 0,
      ecgetciCtlWalkCtlRotate  = 1,
      ecgetciCtlWalkDragRotate = 2,
      etciCtlFlyCtlWalkDragRotate = 3,
      etciCtlPanXYDragRotate   = 4,
    };
    ````
 
    to something like this ->

    ```
    enum ECgeTouchInterface
    {
      ecgetiNone       = 0,
      ecgetiWalk       = 1,
      ecgetiWalkRotate = 2,
      ecgetiFlyWalk    = 3,
      ecgetiPan        = 4,
    };
    ```
    
2. Introduce an indepenedent way to influence `Viewport.InternalWalkNavigation.MouseDragMode`. Either `CGE_SetMouseDragMode` or `CGE_SetVariableInt(ecgevarMouseDragMode)`.

4. And keep, as we have now, an independent way to set boolean `TouchNavigation.AutoTouchInterface`. This is now called `CGE_SetUserInterface`, I would be OK to rename it (since we rename things anyway) to `CGE_SetAutoTouchInterface`.

I realize that I propose a bigger, breaking change than you. I still think it's a good direction -- we break things today, but we have a smooth way forward, the library code is easy to maintain because it is obvious "what maps to what". The resulting library code should be straightforward, if C API things correspond in obvious way to Pascal API.

janadamec1:
OK, I hope, everything is done

michaliskambi (approved):
Thank you -- this looks perfect for me now :) Merging!

PR #645, opened 2024-11-28
Updates to deprecated library iOS example

janadamec1:
Example updated to work with the recent changes in the library interface. Also added full support for iPhone (not only iPad) - namely added LaunchScreen, and UIMenus instead of UIPopoverController usage.


michaliskambi (approved):
Thank you and sorry for delay!

PR #646, opened 2025-01-14
MacOS Compatibility fixes

peardox:
These are some very minor additions to bring CGE back towards MACOS compatibility
Note that I used castleutils_os_specific_unix.inc for Mac (should be OK)
Not 100% about castleutils_filenames.inc so kindly check it

### DIFF
--- a/src/base/castleutils.pas
+++ b/src/base/castleutils.pas
@@ -75,6 +75,7 @@ interface
 uses
   {$ifdef MSWINDOWS} Windows, {$ifndef FPC} ShlObj, {$endif} {$endif}
   {$ifdef UNIX} {$ifdef FPC} BaseUnix, Unix, Dl, {$else} Posix.Unistd, {$endif} {$endif}
+  {$ifdef MACOS} {$ifdef FPC} BaseUnix, Unix, Dl, {$else} Posix.Unistd, {$endif} {$endif}
   {$ifndef FPC} Classes, {$endif}
   Variants, SysUtils, Math, Generics.Collections;
 
@@ -90,6 +91,7 @@ uses
 {$I castleutils_primitive_lists.inc}
 {$I castleutils_program_exit.inc}
 {$ifdef UNIX}      {$I castleutils_os_specific_unix.inc}    {$endif}
+{$ifdef MACOS}     {$I castleutils_os_specific_unix.inc}    {$endif}
 {$ifdef MSWINDOWS} {$I castleutils_os_specific_windows.inc} {$endif}
 {$I castleutils_math.inc}
 {$I castleutils_filenames.inc}
@@ -119,6 +121,7 @@ implementation
   For Lazarus package this would prevent maintaining single .lpk file,
   see ../packages/README. }
 {$ifdef UNIX}      {$I castleutils_os_specific_unix.inc}    {$endif}
+{$ifdef MACOS}     {$I castleutils_os_specific_unix.inc}    {$endif}
 {$ifdef MSWINDOWS} {$I castleutils_os_specific_windows.inc} {$endif}
 
 {$I castleutils_pointers.inc}
--- a/src/base/castleutils_delphi_compatibility.inc
+++ b/src/base/castleutils_delphi_compatibility.inc
@@ -225,6 +225,17 @@ begin
       '.config' + PathDelim +
       ApplicationName + PathDelim;
 {$endif}
+
+{$ifdef MACOS}
+begin
+  // TODO: simplest implementation hardcoded now, look how FPC implements it
+  if Global then
+    Result := '/etc/' + ApplicationName + PathDelim
+  else
+    Result := IncludeTrailingPathDelimiter(GetEnvironmentVariable('HOME')) +
+      '.config' + PathDelim +
+      ApplicationName + PathDelim;
+{$endif}
 end;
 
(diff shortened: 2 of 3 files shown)


michaliskambi:
Can you clarify what is the purpose of this change, what is missing from current CGE?

You say "bring CGE back towards MACOS compatibility", but we never lost macOS compatibility (with FPC) and we never had macOS compatibility (with Delphi). There's nothing to "get back to" :)

To be clear, Castle Game Engine compiles and works on macOS, see https://castle-engine.io/macos , for a long time. We use `DARWIN` (`UNIX` is also defined), not `MACOS`, define, as this is defined by FPC.  We compile correctly (with supported compilers), this is checked by GitHub Actions after every commit. With FPC. We don't support Delphi + macOS right now.

Do you try to port CGE to macOS with Delphi (which defines `MACOS` symbol that I see you use)? But then 

- It should be done a bit differently. Ideally we should not add `{$ifdef MACOS}` special cases in all places. Instead we should neutralize Delphi/FPC differences in `castleconf.inc` , making Delphi define also `UNIX` and `DARWIN` when compiling CGE units on macOS.

- This is incorrect: `{$ifdef MACOS} {$ifdef FPC} BaseUnix, Unix, Dl, {$else} Posix.Unistd, {$endif} {$endif}`. FPC with macOS is already handled, FPC doesn't define MACOS, and ideally no change here will be needed at all if we define `UNIX` for Delphi+macOS .

- And, you have to know that adjusting the lower-level units is just a start :) The most difficult work will be adjusting our FMX integration to initialize OpenGL(ES) context on macOS, which will require macOS-specific code.

peardox:
Sorry Michalis you are, of course, correct and I should have done a FPC test as well (I'll set that up shortly)

The specific purpose of the push is to allow Steam without the rest of CGE to compile cross-platform in Delphi. This is actually beneficial when developing supporting routines in general and non-graphical ones in particular.

My test app basically uses as little of CGE as possible while also using CastleSteam. The aim here is to use unaltered CGE source files with a few dependancies (e.g. castlelog, castleapplicationproperties etc) overridden to make everything compile.

The problem is that some parts of CGE won't compile in Delphi with this in place - the bits in the PR of course.

So dump this PR and I'll check it against FPC and modify to suit then re-PR when it works

michaliskambi:
> The specific purpose of the push is to allow Steam without the rest of CGE to compile cross-platform in Delphi. This is actually beneficial when developing supporting routines in general and non-graphical ones in particular.
> 
> My test app basically uses as little of CGE as possible while also using CastleSteam. The aim here is to use unaltered CGE source files with a few dependancies (e.g. castlelog, castleapplicationproperties etc) overridden to make everything compile.

Thanks for explaining! OK, makes sense and this is cool for me. I mean, 

- before we get the big goal _"full Castle Game Engine works with Delphi+macOS, including TCastleWindow and (FMX) TCastleControl"_,

- ...it makes sense to first only port a few base units, with the goal _"get CastleSteam with all its dependencies working with Delphi+macOS"_. This is a nice concrete goal and practically useful :)

> The problem is that some parts of CGE won't compile in Delphi with this in place - the bits in the PR of course.
> So dump this PR and I'll check it against FPC and modify to suit then re-PR when it works

OK, I'm closing this PR then for now.

peardox:
I tidied up my code and at least on Windows it works nicely without any patches

https://github.com/peardox/PasSteam (the Laz version is a TODO)
Note the dll + steam_api.txt (480) need adding (crappy build deployment Delphi Win)

There is one thing that confuses me in that SteamFMX is showing an FPS counter - I never put one in though. Only happens under Win11 (a Win10 VM is fine)

michaliskambi:
> There is one thing that confuses me in that SteamFMX is showing an FPS counter - I never put one in though. Only happens under Win11 (a Win10 VM is fine)

Possibly you activated the Steam feature to show FPS over the application. ( https://davidwalsh.name/demo/SteamSettingsFPS.png )

peardox:
Ahh - yep - I have that turned on in Steam ATM - thx

Only shows in FMX, presumably cos that uses some Canvas that's OpenGL? I never use anything that triggers any graphics at all...

michaliskambi:
> Only shows in FMX, presumably cos that uses some Canvas that's OpenGL? I never use anything that triggers any graphics at all...

Steam indeed has some detection "when to display overlay", though I don't know their exact criteria. (Our CGE API of TCastleStream allows to check it, https://castle-engine.io/apidoc/html/CastleSteam.TCastleSteam.html#OverlayEnabled )

peardox:
Errm - stupid question - wouldn't ...
{$ifdef MACOS}{$define DARWIN}{$endif}
... be a suitable work-around in Delphi?
I mean they're mutually exclusive AFAIK

michaliskambi:
> Errm - stupid question - wouldn't ...
> {$ifdef MACOS}{$define DARWIN}{$endif}
> ... be a suitable work-around in Delphi?
> I mean they're mutually exclusive AFAIK

Yes, that's exactly what I meant when I said 

>  Instead we should neutralize Delphi/FPC differences in castleconf.inc , making Delphi define also UNIX and DARWIN when compiling CGE units on macOS.

in earlier post in this thread, https://github.com/castle-engine/castle-engine/pull/646#issuecomment-2591183062  .

PR #647, opened 2025-01-16
More CGE + FPC friendly

peardox:
Here we go again...:)

Owing to the location of some changes it was necessary to use ....
{$ifndef FPC}{$ifdef MACOS} ....... {$endif}{$endif}
Several times.
The worst one was DGetAppConfigDir in castleutils_delphi_compatibility.inc which breaks compilation in MACOS (not catered for yet)

I've added compatibility\delphi-only\delphisteam.pas - this should be fine (and it's MACOS only anyway) - it just copies fsome stuff from elsewhere that Mac misses under Dephi

I've recompiled CGE and tested both Lazarus/FPC and Delphi - with the steam demo anyway

The latest version of https://github.com/peardox/PasSteam works properly on all platforms but Laz compilation is a mystery - it keeps complaining about shlobj - dunno why, it should theoretically almost work (might need some attention in steamapp.lcl.pas)

Third time lucky (this is the second :))

### DIFF
--- a/src/base/castledynlib.pas
+++ b/src/base/castledynlib.pas
@@ -268,12 +268,13 @@ begin
     { On macOS, search for dynamic libraries in the bundle too.
       This fallback makes sense for libpng, libvorbisfile, libsteam_api...
       It seems that for everything, so just do it always. }
-    {$ifdef DARWIN}
+    {$if defined(DARWIN)}
     if (Handle = InvalidDynLibHandle) and (BundlePath <> '') then
       Handle := LoadLibrary(PChar(BundlePath + 'Contents/MacOS/' + AName));
     {$elseif not defined(FPC) and defined(MACOS)}
     if (Handle = InvalidDynLibHandle) and (BundlePath <> '') then
       Handle := LoadLibrary(PChar(BundlePath + 'Contents/MacOS/' + AName));
+    {$else}
     {$endif}
   end;
 


peardox:
Lot of fails - eek, what I done wrong now (should have no effect)

peardox:
Messed up an ifdef - actully compiles undef Lazarus now so will hopefully have a cross-platform steam soon (without CGE)

peardox:
Well, everything passed - third time was the charm :)

michaliskambi:
Hm, I'm afraid this is not how I think this should be done.

As explained in https://github.com/castle-engine/castle-engine/pull/646 , 

> Instead we should neutralize Delphi/FPC differences in castleconf.inc , making Delphi define also UNIX and DARWIN when compiling CGE units on macOS.

So we 

- should not introduce special clauses under `{$ifdef MACOS}`. 
- I also do not see the need for special `DelphiSteam`. 
 
Let's instead: 

- define `{$ifdef MACOS}{$define DARWIN}{$endif}` as you say in https://github.com/castle-engine/castle-engine/pull/646#issuecomment-2595257631 , 
-  and adjust all the current macOS code (currently running only with FPC) to also work with Delphi. 
- This also means we should use `BundlePath` already implemented in `CastleFilesUtils`. Do not create a copy of it, do not make a new unit with this copy in `DelphiSteam`.

michaliskambi:
So modifications like this

```
    {$if defined(DARWIN)} // macOS
    'libsteam_api.dylib'
    {$elseif not defined(FPC) and defined(MACOS)}
    'libsteam_api.dylib'
```

or 

```
    {$if defined(DARWIN)}
    if (Handle = InvalidDynLibHandle) and (BundlePath <> '') then
      Handle := LoadLibrary(PChar(BundlePath + 'Contents/MacOS/' + AName));
    {$elseif not defined(FPC) and defined(MACOS)}
    if (Handle = InvalidDynLibHandle) and (BundlePath <> '') then
      Handle := LoadLibrary(PChar(BundlePath + 'Contents/MacOS/' + AName));
    {$else}
```

... should be unnecessary. The existing code already handles macOS by looking at `DARWIN`, and this should be enough, no need to spell everything twice :)

michaliskambi:
P.S. I'm closing this, looking from your previous PR -- I'm guessing you will want to make a new PR.

Feel free to reopen, if you want to work on above suggestions within this PR.

michaliskambi:
P.S. Please also write better subjects... Title "More CGE + FPC friendly" is really wrong and confusing for this PR. CGE + FPC work perfectly. You wanted in this PR to make CGE code more friendly to Delphi on macOS. 

A title like "make unit XXX compile with Delphi on macOS" would be much better.

peardox:
Yeah, that was the commit -m text - didn't spot it'd filtered thru to the PR until too late :(

peardox:
The problem is that Delphi has zero Mac support in some areas

For example if you try building castle_window for ARM64 in Delphi you'll get....

> [dccosxarm64 Error] castleutils_filenames.inc(24): E2029 Expression expected but ';' found
> [dccosxarm64 Error] castleutils_filenames.inc(180): E2029 Expression expected but ';' found
> [dccosxarm64 Error] castleutils.pas(130): E2003 Undeclared identifier: 'InitializationOSSpecific'
> [dccosxarm64 Error] castleutils.pas(174): E2003 Undeclared identifier: 'FinalizationOSSpecific'
> [dccosxarm64 Fatal Error] castlevectors.pas(730): F2063 Could not compile used unit 'castleutils.pas'

Examining these issues you find that Mac is completely ignored e.g. from castleutils_filenames.inc

```
const
  { Root dir name. Empty if not applicable to this OS. }
  RootDir = {$ifdef UNIX} '/' {$endif}
            {$ifdef MSWINDOWS} '' {$endif} ;

```
Note - no consideration of DARWIN... (there are quite a few of these cases)

The purpose is not to build CGE for Mac just simply to compile a non-functional version - which at present you can't do

michaliskambi:
> Note - no consideration of DARWIN... (there are quite a few of these cases)

Which is *exactly* why I wrote in that comment some time ago to define both `DARWIN` and `UNIX`:

> Instead we should neutralize Delphi/FPC differences in castleconf.inc , making Delphi define also UNIX and DARWIN when compiling CGE units on macOS.

peardox:
I've managed to remove all MACOS conditionals form and other junk my version - yipee

Now I only need ...
castlesteam.pas
castleinternalsteamapi.pas
castledynlib.pas

And I can modify them to update to 1.61 (which I've got working)

PR #653, opened 2025-01-30
[Steam] Sanity testing pre steam input for FPC compatability check

peardox:
This branch has one breaking modification. Changed Achievements from TStringList to TObjectList. This does, however, need more attention anyway as it is desirable to access Achievements by a String Key.

It is possibly better to handle Icons on-request (I just tested on a "game" with almost 10k Achievements which was enlightening)

There are new API entries in castleinternalsteamapi - all have been tested and work. There is a sole entry for SteamInput (it's next) merely to test the hook.

I noticed that TSteamBool is used here and there. The callback with a TUserAchievementIconFetched uses a different Boolean (32 bit) so TCallbackBool added to set them apart. Bit of a headache debugging that one as a result...


peardox:
Hmm - was expecting it to sail thru checks...

peardox:
Aborted - trying again....

PR #654, opened 2025-01-31
[steam] Test Steam 1.61 additions

peardox:
This branch has one breaking modification. Changed Achievements from TStringList to TObjectList. This does, however, need more attention anyway as it is desirable to access Achievements by a String Key.

It is possibly better to handle Icons on-request (I just tested on a "game" with almost 10k Achievements which was enlightening)

There are new API entries in castleinternalsteamapi - all have been tested and work. There is a sole entry for SteamInput (it's next) merely to test the hook.

I noticed that TSteamBool is used here and there. The callback with a TUserAchievementIconFetched uses a different Boolean (32 bit) so TCallbackBool added to set them apart. Bit of a headache debugging that one as a result...

### DIFF
--- a/src/base/castledynlib.pas
+++ b/src/base/castledynlib.pas
@@ -23,7 +23,7 @@ interface
 uses SysUtils
   {$ifdef FPC}
     { With FPC, use cross-platform DynLibs unit. }
-    {$ifndef WASI}, DynLibs{$endif}
+    , DynLibs
   {$else}
     { With Delphi, use Windows functions directly.
       On non-Windows, Delphi SysUtils defines compatible functions
@@ -37,12 +37,7 @@ type
 
 const
   { Invalid TDynLibHandle value (meaning : LoadLibrary failed) }
-  InvalidDynLibHandle: TDynLibHandle =
-    {$if defined(FPC) and not defined(WASI)}
-    DynLibs.NilHandle
-    {$else}
-    0 // used with Delphi or FPC+WebAssembly
-    {$endif};
+  InvalidDynLibHandle: TDynLibHandle = {$ifdef FPC} DynLibs.NilHandle {$else} 0 {$endif};
 
 type
   { }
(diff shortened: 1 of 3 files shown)


peardox:
Rats - FPC issues...

michaliskambi:
Please don't open and close PRs like this ( https://github.com/castle-engine/castle-engine/pull/653 , https://github.com/castle-engine/castle-engine/pull/654 , and now this PR ). You are making a number of notifications to everyone watching this repo, and it is a non-standard usage of PRs. Pull requests are not supposed to be "one shot", where you either make a perfect PR on the first try, that passes all the tests, or resign and close the PR. Don't close each PR just because GitHub Actions check fails, only to open another PR again.

Instead, basically, open a PR and work on it, until you think it's good. 

- Create a fork of CGE (and maybe a branch) and test it as much as you can,
- then open a PR, 
- mark it as a "draft" if you want to indicate that it's not ready for review/merge yet, 
- and then work on it. This means pushing more commits to the same branch in the same fork that you used to make the PR. The PR will automatically show new commits, and they will be checked by GitHub Actions.

New commits to PR will also make a notification to people watching, but this is standard (and everyone should be able to handle it, but unsubscribing from PR if they are not interested), people expect this.

Looking at PR contents:

- In `src/base/castledynlib.pas` , it seems your changes didn't merge properly with latest changes in CGE master. Make sure your diff (visible in "Files Changed") contain only things you wanted to change, revert any other changes.
- Writing `AchievementIdAnsi := AnsiString(AchievementId)` -- we don't want an explicit typecast in this case, variable names already make it obvious. Leave it as `AchievementIdAnsi := AchievementId`, as it was.
- While `USE_TESTING_API` makes perfect sense for testing, we will want to remove and simplify it all (to just use new API) before merging.

peardox:
Thanks for the comments and pointers. All fully understood.

michaliskambi:
P.S. It seems I posted this comment in wrong PR, I wanted to post it in latest, https://github.com/castle-engine/castle-engine/pull/655 :)

peardox:
Yeah, no problem. I understood it was about all three.

I have some questions regarding your comments (to educate me) - I'll do those in PM Discord I guess to keep Git clean.

I'll kill this branch after I've reviewed the logs -they were what I was after. I'm writing this specific service for use both outside and inside CGE so wanted to check how my altertions affected the CGE side of things.

Issue #656, opened 2025-02-07
[Steam] Minor Shutdown Issue


peardox:
This is rather trivial and NOT documented as a "do this" by Steam. Things work without this, they just work better with it.

I'm not suggesting this is needed, merely pointing it out (also I've not given it really thorough testing yet)

There exists what I'd guess is a resource leak in Steam. At startup it is required to get a handle to Steam Pipe...

`    SteamPipeHandle := SteamAPI_GetHSteamPipe();
`

This handle is however never released even though a mechanism exists to do so in SteamAPI_ISteamClient_BReleaseSteamPipe. Adding this to the current Destructor for Steam requires the definition of SteamAPI_ISteamClient_BReleaseSteamPipe of course. After this is done we get...

```
destructor TCastleSteam.Destroy;
begin
  FreeAndNil(FAchievements);
  if Enabled then
    begin
      SteamAPI_ISteamClient_BReleaseSteamPipe(SteamClient, SteamPipeHandle);
      SteamAPI_Shutdown();
    end;
  if ApplicationProperties(false) <> nil then
    ApplicationProperties(false).OnUpdate.Remove({$ifdef FPC}@{$endif} Update);
  inherited;
end;
```

I imagine that after Shutdown eventually the pipe goes stale and Steam discards it then on restart will create another.
One reason that you possibly wouldn't want to do this is if SteamAPI_RestartAppIfNecessary is called the pipe would be re-used (situation dependant - could flag it I guess).

The proof of this can be seen with the current Steam Demo (if this was added)...

[Without releasing pipe]
The effect of NOT releasing the pipe is subtle. Run the demo then close it and change the ID in steam_appid.txt - the demo will get confused and no Achievements will be displayed. If you _shut down_ steam then restart it the demo will pick up the Achievements will be displayed as expected.

[Release pipe on close]
The demo should now not need Steam restarting in order to switch AppIDs as can be illustrated by changing the ID in steam_appid.txt - there is no confusion.

michaliskambi:
Thank you, this seems reasonable to do (even if just for the sake of "closing clean", and if it has actual positive effect on user experience -> all the better). I'll test the described solution (probably on upcoming Monday) and do this. If you want to create a PR proposing this, I would also be happy to just test it and apply :)

peardox:
Steam say that...

```
ISteamClient Interface

Interface to create a new steam instance or to connect to an existing steam instance, whether it's in a different process or is local.

[...]

PR #657, opened 2025-02-09
[steam] Add SteamAPI_ISteamClient_BReleaseSteamPipe 

peardox:
As discussed in #656 

Small changes, add definition for SteamAPI_ISteamClient_BReleaseSteamPipe 
Alter TCastleSteam.Destroy to call it in order to release the pipe (theoretically a resource leak)
No wrapping yet as it's only of any real use on Shutdown ATM (may be exposed later if required)

### DIFF
--- a/src/services/steam/castleinternalsteamapi.pas
+++ b/src/services/steam/castleinternalsteamapi.pas
@@ -227,6 +227,7 @@ var
   // which is actually "flat" (C, no classes) Steam API corresponding to the C++ API
 
   // ISteamClient
+  SteamAPI_ISteamClient_BReleaseSteamPipe: function (SteamClient: Pointer; hSteamPipe: HSteamPipe): LongBool; CDecl;
   SteamAPI_ISteamClient_SetWarningMessageHook: procedure (SteamClient: Pointer; WarningMessageHook: SteamAPIWarningMessageHook); CDecl;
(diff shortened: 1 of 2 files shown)


michaliskambi:
Than you, I'll look and test it soon!

peardox:
It's only 5 lines - not the biggest PR :)

michaliskambi:
Sorry for a delay with this, I have this PR on a list of TODOs for next week. Thanks for the patience :) 

( 
I know, it's a short PR, but I do need to test whether it doesn't cause any regressions on the existing platforms, Linux, Windows, macOS.

If you want to help me speed it up, I will appreciate if you add a comment justifying the need for this to the source code , before the `SteamAPI_ISteamClient_BReleaseSteamPipe`. I recall you described it in one of the previous PRs, let's copy/adjust that description into a Pascal comment, so that it stays forever recorded. Like 

```delphi
  if Enabled then
  begin
    { Without doing SteamAPI_ISteamClient_BReleaseSteamPipe , an invalid behavior occurs when X. 
       Reproduce it by doing Y ...  }
    SteamAPI_ISteamClient_BReleaseSteamPipe(SteamClient, SteamPipeHandle);
    SteamAPI_Shutdown();
  end;
```

Thank you!
)

PR #658, opened 2025-02-17
[Experimental][Draft] Steam Subsystems

peardox:
Moved every subsystem to it's own unit

Works under Delphi, Breaks under FPC

Includes many untested Steam Input functions (WIP)


michaliskambi:
Closing this PR -- I understand you're not working on this now. 

The PR does contain some wanted + accidental changes, like log name in `examples/steam/code/gameviewmain.pas` (we want to keep using default log locations in CGE examples).

The overall goal of this may be still something we want. We may want to "break down" both our public API (`CastleSteam` unit) and internal Steam API translation (`CastleInternalSteamApi` unit) into more files. But let's not over-engineer it too soon, i.e. let's not create a complicated solution without focusing on a concrete need. I would say to instead focus on exposing some particular Steam feature we need , from TODOs on https://castle-engine.io/steam : 

- support joysticks through Steam API (should be very nice with new https://castle-engine.io/controllers API)
- SteamDeck stuff

Let's keep extending the API in the simplest way to achieve these goals. Let's change the architecture of `CastleSteam` / and/or `CastleInternalSteamApi` when we will need it, and then we will also have a better idea "how" to make this split best (units? classes? include files?).

PR #668, opened 2025-03-25
Add function to set MultField value to library interface

janadamec1:
I've added a new function to (deprecated) library interface: CGE_SetNodeMultFieldValue

There is already a function to set SFxxx values, I need a function to set MFxxx values too. So this solves my problem and I can dynamically update the such fields too (tested).

Please merge it to master soon 🙏, so that I can merge it to other library-related branches (URL processing).

### DIFF
--- a/src/deprecated_library/castleengine.h
+++ b/src/deprecated_library/castleengine.h
@@ -379,6 +379,7 @@ extern void CGE_SetNodeFieldValue_MFVec2d(const char *szNodeName, const char *sz
 extern void CGE_SetNodeFieldValue_MFVec3d(const char *szNodeName, const char *szFieldName, int iCount, double *values);  // we expect "3 * iCount" doubles in the array "values"
 extern void CGE_SetNodeFieldValue_MFVec4d(const char *szNodeName, const char *szFieldName, int iCount, double *values);  // we expect "4 * iCount" doubles in the array "values"
 extern void CGE_SetNodeFieldValue_MFRotation(const char *szNodeName, const char *szFieldName, int iCount, float *values);  // we expect "4 * iCount" floats in the array "values"
+extern void CGE_SetNodeFieldValue_MFString(const char *szNodeName, const char *szFieldName, int iCount, char **values);  // We expect array of "iCount" char* pointers to null-terminated UTF-8 strings
 
 extern void CGE_IncreaseSceneTime(float fTimeS);    // set time in the scene, useful when viewport paused
 
--- a/src/deprecated_library/castleengine.h
+++ b/src/deprecated_library/castleengine.h
@@ -379,7 +379,7 @@ extern void CGE_SetNodeFieldValue_MFVec2d(const char *szNodeName, const char *sz
 extern void CGE_SetNodeFieldValue_MFVec3d(const char *szNodeName, const char *szFieldName, int iCount, double *values);  // we expect "3 * iCount" doubles in the array "values"
 extern void CGE_SetNodeFieldValue_MFVec4d(const char *szNodeName, const char *szFieldName, int iCount, double *values);  // we expect "4 * iCount" doubles in the array "values"
 extern void CGE_SetNodeFieldValue_MFRotation(const char *szNodeName, const char *szFieldName, int iCount, float *values);  // we expect "4 * iCount" floats in the array "values"
-extern void CGE_SetNodeFieldValue_MFString(const char *szNodeName, const char *szFieldName, int iCount, char **values);  // We expect array of "iCount" char* pointers to null-terminated UTF-8 strings
+extern void CGE_SetNodeFieldValue_MFString(const char *szNodeName, const char *szFieldName, int iCount, const char **values);  // We expect array of "iCount" char* pointers to null-terminated UTF-8 strings
 
 extern void CGE_IncreaseSceneTime(float fTimeS);    // set time in the scene, useful when viewport paused
 
--- a/src/deprecated_library/castleengine.h
+++ b/src/deprecated_library/castleengine.h
@@ -1,5 +1,5 @@
 /*
-  Copyright 2013-2024 Jan Adamec, Michalis Kamburelis.
+  Copyright 2013-2025 Jan Adamec, Michalis Kamburelis.
 
   This file is part of "Castle Game Engine".
 
@@ -357,6 +357,8 @@ extern void CGE_SetVariableInt(int /*ECgeVariable*/ eVar, int nValue);
 extern int CGE_GetVariableInt(int /*ECgeVariable*/ eVar);
 
 extern void CGE_SetNodeFieldValue(const char *szNodeName, const char *szFieldName,
+                                  float fVal1, float fVal2, float fVal3, float fVal4);
+extern void CGE_SetNodeMultFieldValue(const char *szNodeName, const char *szFieldName, int iIndex,
                                   float fVal1, float fVal2, float fVal3, float fVal4);
 
 extern void CGE_IncreaseSceneTime(float fTimeS);    // set time in the scene, useful when viewport paused
--- a/src/deprecated_library/castleengine.h
+++ b/src/deprecated_library/castleengine.h
@@ -356,10 +356,29 @@ extern void CGE_SetWalkNavigationMouseDragMode(int /*ECgeMouseDragMode*/ eMode);
 extern void CGE_SetVariableInt(int /*ECgeVariable*/ eVar, int nValue);
 extern int CGE_GetVariableInt(int /*ECgeVariable*/ eVar);
 
-extern void CGE_SetNodeFieldValue(const char *szNodeName, const char *szFieldName,
-                                  float fVal1, float fVal2, float fVal3, float fVal4);
-extern void CGE_SetNodeMultFieldValue(const char *szNodeName, const char *szFieldName, int iIndex,
-                                  float fVal1, float fVal2, float fVal3, float fVal4);
+extern void CGE_SetNodeFieldValue_SFFloat(const char *szNodeName, const char *szFieldName, float value);
+extern void CGE_SetNodeFieldValue_SFDouble(const char *szNodeName, const char *szFieldName, double value);
+extern void CGE_SetNodeFieldValue_SFInt32(const char *szNodeName, const char *szFieldName, int value);
+extern void CGE_SetNodeFieldValue_SFBool(const char *szNodeName, const char *szFieldName, int value);
+extern void CGE_SetNodeFieldValue_SFVec2f(const char *szNodeName, const char *szFieldName, float val1, float val2);
+extern void CGE_SetNodeFieldValue_SFVec3f(const char *szNodeName, const char *szFieldName, float val1, float val2, float val3);
+extern void CGE_SetNodeFieldValue_SFVec4f(const char *szNodeName, const char *szFieldName, float val1, float val2, float val3, float val4);
+extern void CGE_SetNodeFieldValue_SFVec2d(const char *szNodeName, const char *szFieldName, double val1, float val2);
+extern void CGE_SetNodeFieldValue_SFVec3d(const char *szNodeName, const char *szFieldName, double val1, float val2, float val3);
+extern void CGE_SetNodeFieldValue_SFVec4d(const char *szNodeName, const char *szFieldName, double val1, float val2, float val3, float val4);
+extern void CGE_SetNodeFieldValue_SFRotation(const char *szNodeName, const char *szFieldName, float axis1, float axis2, float axis3, float rotation);
+extern void CGE_SetNodeFieldValue_SFString(const char *szNodeName, const char *szFieldName, const char *value);
+extern void CGE_SetNodeFieldValue_MFFloat(const char *szNodeName, const char *szFieldName, int iCount, float *values);
+extern void CGE_SetNodeFieldValue_MFDouble(const char *szNodeName, const char *szFieldName, int iCount, double *values);
+extern void CGE_SetNodeFieldValue_MFInt32(const char *szNodeName, const char *szFieldName, int iCount, int *values);
+extern void CGE_SetNodeFieldValue_MFBool(const char *szNodeName, const char *szFieldName, int iCount, float *values);
+extern void CGE_SetNodeFieldValue_MFVec2f(const char *szNodeName, const char *szFieldName, int iCount, float *values);  // we expect "2 * iCount" floats in the array "values"
+extern void CGE_SetNodeFieldValue_MFVec3f(const char *szNodeName, const char *szFieldName, int iCount, float *values);  // we expect "3 * iCount" floats in the array "values"
+extern void CGE_SetNodeFieldValue_MFVec4f(const char *szNodeName, const char *szFieldName, int iCount, float *values);  // we expect "4 * iCount" floats in the array "values"
+extern void CGE_SetNodeFieldValue_MFVec2d(const char *szNodeName, const char *szFieldName, int iCount, double *values);  // we expect "2 * iCount" doubles in the array "values"
+extern void CGE_SetNodeFieldValue_MFVec3d(const char *szNodeName, const char *szFieldName, int iCount, double *values);  // we expect "3 * iCount" doubles in the array "values"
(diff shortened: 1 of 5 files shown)


michaliskambi:
Thank you! 

Code looks good, and the API you propose is consistent with how setting single-value fields work in `CGE_SetNodeFieldValue`.

However,  I see some issues with this definition of `CGE_SetNodeMultFieldValue(szNodeName, szFieldName: pcchar; iIndex: cInt32; fVal1, fVal2, fVal3, fVal4: cFloat)`:

- This approach offers no way to get or set the count of multi-value field. Code has to assume that existing field has enough length. The `ItemsSafe` setters will make sure to make a warning (not crash) when invalid index is given, but the code still doesn't have a chance to actually query (or change) this count beforehand.
- This approach is not efficient for setting larger multi-value fields (like indexes or 3D coordinates of meshes, `TCoordinateNode.FdCoord`), because change to every index makes one `TMFxxx(aField).Send` call that resends whole modified array. We don't recommend to use our Pascal API this way, we shall also not recommend this approach for C developers. When possible, set the whole value of the field, from C array.
- Also, not every type can be expressed as "4 floats". So e.g. this doesn't allow to set `MFString` or `MFMatrix4f` etc. I know, this concerns was also present for single-value `CGE_SetNodeFieldValue`.

Proposition:

I would propose instead a set of `CGE_SetNodeFieldValue_Xxx` that are dedicated to each field type, and take an array of given C types. Like this:

```c
// Set MFFloat.
void CGE_SetNodeFieldValue_MFFloat(const char *szNodeName, const char *szFieldName, int count, float* values);
// Set MFVec3f. We expect "3 * count" floats in the array "values"
void CGE_SetNodeFieldValue_MFVec3f(const char *szNodeName, const char *szFieldName, int count, float* values);
// Set MFVec3f. We expect "4 * count" floats in the array "values"
void CGE_SetNodeFieldValue_MFVec4f(const char *szNodeName, const char *szFieldName, int count, float* values);
// Set MFString. We expect each C string (char*) to encode international characters using UTF-8.
void CGE_SetNodeFieldValue_MFString(const char *szNodeName, const char *szFieldName, int count, char** values);
....
```

You could also introduce single-value versions:

```c
void CGE_SetNodeFieldValue_SFFloat(const char *szNodeName, const char *szFieldName, float value);
void CGE_SetNodeieldValue_SFVec3f(const char *szNodeName, const char *szFieldName, float x, float y, float z);
void CGE_SetNodeFieldValue_SFVec4f(const char *szNodeName, const char *szFieldName, float x, float y, float z, float w);
void CGE_SetNodeFieldValue_SFString(const char *szNodeName, const char *szFieldName, char* value);
....
```

I know it's more work to define things like this, we will introduce a lot of new `CGE_SetNodeFieldValue_Xxx` routines. But it seems like a more "future-proof" solution:

- You will have an API to set any type supported by CGE, including strings, matrices.
- You avoid current conversions like `Round(fVal1)` when someone uses `CGE_SetNodeMultFieldValue` to set MFInt32 / MFLong.
- You can pass `double` precisely too, C `double` is Pascal `Double` and goes to `SFDouble` or `MFDouble` in X3D.
- For multi-value fields, you have efficient API that sets the count, *and* provides all the values in one go. E.g. setting something like `TCoordinateNode.FdCoord` with `CGE_SetNodeFieldValue_MFVec3f` will make sense, it will be efficient, so C developers can can animate mesh coordinates.
- It also seems more obvious to me, looking at it from the perspective of C developer. The routines take simple C arrays of the corresponding type.

To be clear, I'm not saying that it's required to define this for all possible X3D types for the start. Just start with a few X3D types you actually need, and extend it as you need more. ( Through with the help of GitHub's Copilot, defining a lot of `CGE_SetNodeFieldValue_Xxx` may be easy, it works nicely in my experience when it has to generate a code that "repeats a recognizable pattern", like in this case).

P.S. BTW, sorry for still not applying your past PRs -- I have them on my TODO, and I keep moving it, due to having too many TODOs.... I really really will get around and merge everything soon. Thank for the patience!

janadamec1:
OK, as you wish. I agree, it will certainly be way more cleaner.
But adding all those new functions is a tedious task, as I need to add each to like 8 places with different forms, and must not make a mistake. And now you want me to add 20 new functions...

michaliskambi:
> But adding all those new functions is a tedious task, as I need to add each to like 8 places with different forms, and must not make a mistake. And now you want me to add 20 new functions...

Oh, I understand my proposal is more work now, but I think the resulting solution will be more useful, so it will require less work in the future to extend it.

You will be able to account for any X3D type (like strings, to set X3D `Text.string`), you will enable setting count for multiple-value fields, you will have efficient API (that works also for e.g. setting coordinates for a mesh like X3D `Coordinate.coord`). The current solution would likely need rework in the future -- as soon as you would need to change count, or set `MFString`, or even set `MFDouble` with precision, the current API would not be good enough.

Thank you!

src/deprecated_library/castleengine.h
```diff
@@ -356,8 +356,30 @@ extern void CGE_SetWalkNavigationMouseDragMode(int /*ECgeMouseDragMode*/ eMode);
...
+extern void CGE_SetNodeFieldValue_SFInt32(const char *szNodeName, const char *szFieldName, int value);
+extern void CGE_SetNodeFieldValue_SFBool(const char *szNodeName, const char *szFieldName, int value);
+extern void CGE_SetNodeFieldValue_SFVec2f(const char *szNodeName, const char *szFieldName, float val1, float val2);
+extern void CGE_SetNodeFieldValue_SFVec3f(const char *szNodeName, const char *szFieldName, float val1, float val2, float val3);
+extern void CGE_SetNodeFieldValue_SFVec4f(const char *szNodeName, const char *szFieldName, float val1, float val2, float val3, float val4);
+extern void CGE_SetNodeFieldValue_SFVec2d(const char *szNodeName, const char *szFieldName, double val1, float val2);
+extern void CGE_SetNodeFieldValue_SFVec3d(const char *szNodeName, const char *szFieldName, double val1, float val2, float val3);
+extern void CGE_SetNodeFieldValue_SFVec4d(const char *szNodeName, const char *szFieldName, double val1, float val2, float val3, float val4);
+extern void CGE_SetNodeFieldValue_SFRotation(const char *szNodeName, const char *szFieldName, float axis1, float axis2, float axis3, float rotation);
```
michaliskambi, line 369: Call `axis1`, `axis2`, `axis3` -> `axisX`, `axisY`, `axisZ`. This makes the purpose clear, this is always an XYZ of the axis.

src/deprecated_library/castleengine.h
```diff
@@ -356,8 +356,30 @@ extern void CGE_SetWalkNavigationMouseDragMode(int /*ECgeMouseDragMode*/ eMode);
 extern void CGE_SetVariableInt(int /*ECgeVariable*/ eVar, int nValue);
 extern int CGE_GetVariableInt(int /*ECgeVariable*/ eVar);
 
-extern void CGE_SetNodeFieldValue(const char *szNodeName, const char *szFieldName,
-                                  float fVal1, float fVal2, float fVal3, float fVal4);
+extern void CGE_SetNodeFieldValue_SFFloat(const char *szNodeName, const char *szFieldName, float value);
+extern void CGE_SetNodeFieldValue_SFDouble(const char *szNodeName, const char *szFieldName, double value);
+extern void CGE_SetNodeFieldValue_SFInt32(const char *szNodeName, const char *szFieldName, int value);
+extern void CGE_SetNodeFieldValue_SFBool(const char *szNodeName, const char *szFieldName, int value);
```
michaliskambi, line 362: Can we use `bool` in C, looks more obvious than `int`? 

I understand from https://www.geeksforgeeks.org/bool-in-c/ , https://github.com/gcc-mirror/gcc/blob/master/gcc/ginclude/stdbool.h that `#include <stdbool.h>` defines it. 

I know it's still just `int` underneath, but calling it `bool` makes the purpose obvious for anyone reading the C header.

I see the `CTypes` in Pascal ( https://www.freepascal.org/docs-html/current/rtl/ctypes/index-3.html ) also defines `CBool` to match C `bool`.

src/deprecated_library/castleengine.h
```diff
@@ -356,8 +356,30 @@ extern void CGE_SetWalkNavigationMouseDragMode(int /*ECgeMouseDragMode*/ eMode);
...
+extern void CGE_SetNodeFieldValue_SFVec2d(const char *szNodeName, const char *szFieldName, double val1, float val2);
+extern void CGE_SetNodeFieldValue_SFVec3d(const char *szNodeName, const char *szFieldName, double val1, float val2, float val3);
+extern void CGE_SetNodeFieldValue_SFVec4d(const char *szNodeName, const char *szFieldName, double val1, float val2, float val3, float val4);
+extern void CGE_SetNodeFieldValue_SFRotation(const char *szNodeName, const char *szFieldName, float axis1, float axis2, float axis3, float rotation);
+extern void CGE_SetNodeFieldValue_SFString(const char *szNodeName, const char *szFieldName, const char *value);
+extern void CGE_SetNodeFieldValue_MFFloat(const char *szNodeName, const char *szFieldName, int iCount, float *values);
+extern void CGE_SetNodeFieldValue_MFDouble(const char *szNodeName, const char *szFieldName, int iCount, double *values);
+extern void CGE_SetNodeFieldValue_MFInt32(const char *szNodeName, const char *szFieldName, int iCount, int *values);
+extern void CGE_SetNodeFieldValue_MFBool(const char *szNodeName, const char *szFieldName, int iCount, float *values);
```
michaliskambi, line 374: The `values` type should match the type for `SFBool` version. So eithe `int *values` or `bool *values` (see my comment at `SFBool` above). We do not need to pass booleans as floats :)

src/deprecated_library/castleengine.lpr
```diff
@@ -910,53 +911,431 @@ function CGE_GetVariableInt(eVar: cInt32): cInt32; cdecl;
...
+procedure CGE_SetNodeFieldValue_SFFloat(szNodeName, szFieldName: pcchar; value: cFloat); cdecl;
 var
   aField: TX3DField;
 begin
   try
-    if not CGE_VerifyScene('CGE_SetNodeFieldValue') then exit;
+    if not CGE_VerifyScene('CGE_SetNodeFieldValue_SFFloat') then exit;
+    aField := MainScene.Field(PChar(szNodeName), PChar(szFieldName));
+    if (aField <> nil) and (aField is TSFFloat) then
+       TSFFloat(aField).Send(value);
```
michaliskambi, line 922: We need to have a way to notify developer in case the type is incorrect. I.e. right now the `CGE_SetNodeFieldValue_SFFloat` will just silently not do anything (no warning, no error), if it will be used for X3D type other than `SFFloat`.

My proposal: just use "as" cast, and let it raise an exception (which you catch around this block already, so it will result in a nice message) when the type is incorrect.

So instead of 

```delphi
    if (aField <> nil) and (aField is TSFFloat) then
       TSFFloat(aField).Send(value);
```

->

```delphi
    if aField <> nil then
       (aField as TSFFloat).Send(value);
```

src/deprecated_library/castleengine.lpr
```diff
@@ -910,53 +911,431 @@ function CGE_GetVariableInt(eVar: cInt32): cInt32; cdecl;
...
+  aItemList: TSingleList;
+  i: cInt32;
+begin
+  try
+    if not CGE_VerifyScene('CGE_SetNodeFieldValue_MFFloat') then exit;
+    aField := MainScene.Field(PChar(szNodeName), PChar(szFieldName));
+    if (aField = nil) or not (aField is TMFFloat) then Exit;
+
+    aItemList := TSingleList.Create;
+    for i:=0 to iCount-1 do
```
michaliskambi, line 1094: Copy the contents in more efficient way, knowing that the memory layout of C and Pascal types matches precisely in this case. Instead of

```delphi
    for i:=0 to iCount-1 do
      aItemList.Add(values[i]);
```

you should be able to just do

```delphi
aItemList.Count := iCount;
Move(Values^, aItemList.L^, SizeOf(Single) * iCount);
```

This makes single allocation of memory for Pascal list, and single fast copy of the contents.

This optimization (and also a bit simpler code) should be applied to all multi-value setters (**except MFString**, where the normal iteration as you do already is good and safe; for strings, C and Pascal types are not equivalent).

src/deprecated_library/castleengine.lpr
```diff
@@ -910,53 +911,431 @@ function CGE_GetVariableInt(eVar: cInt32): cInt32; cdecl;
...
+  aItemList: TVector2List;
+  i: cInt32;
+begin
+  try
+    if not CGE_VerifyScene('CGE_SetNodeFieldValue_MFVec2f') then exit;
+    aField := MainScene.Field(PChar(szNodeName), PChar(szFieldName));
+    if (aField = nil) or not (aField is TMFVec2f) then Exit;
+
+    aItemList := TVector2List.Create;
+    for i:=0 to iCount-1 do
```
michaliskambi, line 1179: Similarly to `MFFloat`, this can also be speed up, making it even simpler.

```delphi
aItemList.Count := iCount;
Move(Values^, aItemList.L^, SizeOf(TVector2) * iCount);
```

michaliskambi (changes requested):
Thank you! This looks good, I hope it will be useful to you and others who want to access our API from C.

I point a few changes in the review below, after we solve them -> I think it's perfect to merge :)

janadamec1:
I hope it is better now.

Just one note, as I write in one sub-comment, for person not knowing what's beyond TList, I am very reluctant to use Move(...) function to copy values. I preallocated TList with .Count := iCount, but Move(...) seems too dangerous to me (maybe some future implementation of TList will not store values in one memory block).

src/deprecated_library/castleengine.lpr
michaliskambi: >but Move(...) seems too dangerous to me (maybe some future implementation of TList will not store values in one memory block). 

Oh, in general, it's sure a valid concern to be "resistant" to future implementation changes.

But in this case I think it's really OK -- reasons below in AD 1, AD 2. And the benefits are substantial, see AD 3. So I would still say to do this. i.e. rework code to use `Move`. Otherwise, you will cause slowdowns when trying to do some reasonable operations, like animating a mesh.

Details:

1. It is really guaranteed: 
    - `TList<T>` is guaranteed to store the values in a linear, consecutive memory layout, just like `packed array of T`. Both FPC and Delphi keep this assertion. 
    - Because breaking this -> would mean they have to also break API that is exposed to developers, like their `FItems` (FPC) / `TList.List` ([Delphi](https://docwiki.embarcadero.com/Libraries/Sydney/en/System.Generics.Collections.TList.List)). 

3. In CGE, we already rely on this in many places. Some examples:
    - Various places of https://github.com/castle-engine/castle-engine/blob/master/src/base/castleutils_struct_list.inc rely on it, like definitions of our `L` and our `List` and `Ptr`. These things would be unimplementable without relying on certain memory layout.
    - Passing it to OpenGL in some places relies on this, like in  https://github.com/castle-engine/castle-engine/blob/master/src/base_rendering/castlerenderprimitives_render_unlit_mesh.inc . 
    - Things like `TVector2List.AddSubRange` rely on this too, using `Move`, see https://github.com/castle-engine/castle-engine/blob/master/src/base/castlevectors_lists.inc#L240 .

    Practically speaking, if FPC or Delphi will break the `TList<T>` memory layout (which they really should not ever do, see AD 1) -> then in CGE we will stop using them, and invent our own structure for these lists. We need them to be just linear data in memory, to enable passing to other libs, and fast processing (using `Move`, but also friendly for CPU caches when iteration is necessary).

4. It's may be really beneficial for performance (which is also the reason why in CGE, in examples given in AD 2, we rely on this too).

    I'm thinking here about animating meshes, which can easily have thousands vertexes, and sometimes have millions. So `CGE_SetNodeFieldValue_MFVec3f` must be efficient to allow animating a mesh by setting `Coordinate.coord` every frame . One `Move` is much faster than thousands separate assignments.

michaliskambi:
Note: I broke yesterday GitHub Actions checks by committing to master a change that isn't valid for Delphi, sorry -> that's why GHA for this PR fails too, it's completely unrelated to your changes. The fix is already in master (and should pass checks soon).

src/deprecated_library/castleengine.lpr
janadamec1: OK, done.

michaliskambi (approved):
Looks great, thank you very much - merging!

PR #670, opened 2025-04-23
User defined Disable of media formats, allows to shrink binary file

phomm:
This is only a suggestion, I don't expect that it would be merged to upstream, but let's see if @michaliskambi sees benefits of it.

For example I use only GLTF and SpriteSheets in my game, when I checked the castle-engine-output, I see big compiled files like castleicf and other similar from src/scene/load, I don't need support of those formats in Release build, and I know linker is not so smart to remove unused code, so I did it via conditional compilation (and it works as those modules are just pluggable, and no other module depend on them).
comparison (I defined all other formats disabling, except gtlf and spritesheet):
image
so the benefit is 2.5 mb, which is 20-25% of exe file size.

Ofc, this is not intended to be enabled by default, only by user who knows the outcome, and it makes sense only for release I think.
P.S. also a bit corrected gitignore to ignore embedded fpc/lazarus, as I'm tired already by constantly manipulating while working with sources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,14 +67,50 @@ castle-game-engine-*.fpm
 /fpmake
 /units/
 
-# alphabetic list of specific subdirs (and files inside)
-
+# manpages output
 /doc/man/man1/*.1
+
+# PasDoc output
 /doc/pasdoc/cache/
 /doc/pasdoc/html/
 /doc/pasdoc/latex/
 /doc/pasdoc/php/
 /doc/reference/
+
+# Bundled FPC (ignore it, in case someone puts it in CGE source tree)
+/tools/contrib/fpc
+
+# Proprietary Android libraries
+# (it may be easy to put them into CGE source tree, but don't commit).
+/tools/build-tool/data/android/integrated-services/chartboost/app/libs/*.jar
+/tools/build-tool/data/android/integrated-services/game_analytics/app/libs/*.jar
+/tools/build-tool/data/android/integrated-services/startapp/app/libs/*.jar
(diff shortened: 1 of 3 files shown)


michaliskambi:
I pushed small modifications that address some of my own comments yesterday,

- We ignore `tools/contrib/fpc`.
- I changed the comments in `castleconf.inc` to recommend to use `CASTLE_xxx_SUPPORT_DISABLE` for both debug and release modes, and to define these symbols at project level (not modifying the CGE sources).

Merging, to avoid my own "decision paralysis" :) Maybe I will decide to address this need in some other way in the future.... but for now this is a proper improvement. This breaks nothing, is trivial to maintain, and it makes the improvement (you can have smaller exes if you don't need IFC etc.).

Thank you!

phomm:
I agree, that switching from something like defines to something more user-friendly (maybe even with setup from editor and writing into CastleEngineManifest.xml for auto-regenerate), and in result it could be automatically (or by user) written into uses clause of standard GameInitialize.pas, would be definitely better and usable. I would try this locally (transfer units from X3DLoad to main game unit) and see if it could be put as PR

About Release and Debug - this is ofc personal preference, so I'm flexible during development (if I would add something that I've disabled, so it would work under debug at least, I know this is doubtful, but still)

About gitignore (and lazarus in it) - this is also a bit personal and I totally agree about putting FPC only to gitignore, and I will also try exclude, that you suggeted.

In total I think this PR is then not really useful, just as a point of discussion maybe :) I will close it

PR #674, opened 2025-05-29
Fixes around RenderToTexture on iOS, enable OpenGL ES 3 context on iOS

janadamec1:
This PR fixes the screen effects with depth (like SSAO) on iOS. Also fixes the mirrors and shadows rendering on iOS.

To do so, we enabled OpenGL ES 3 context on iOS, and changed the way the context format is set up (mainly to use Depth24Stencil8 format that is one of few supported formats on iOS).

The code changes are implemented in iOS glue code in `tools/build-tool` (i.e. check results in castle-model-viewer-mobile) and `examples/deprecated_library/ios_tester`.


src/scene/castlescreeneffects.pas
```diff
@@ -460,10 +461,16 @@ procedure TCastleScreenEffects.Render;
     end;
     if Depth then
     begin
+      CheckGLErrors('At CreateScreenEffectTexture before creating depth');
```
michaliskambi, line 464: Please remove the `CheckGLErrors` calls from the PR. 

I know, they are useful for debugging, I also still sometimes just temporarily add them and then remove. But:

- We have a better debugging feature, [TGLFeatures.Debug](https://castle-engine.io/apidoc/html/CastleGLUtils.TGLFeatures.html#Debug), using [KHR_debug](https://www.khronos.org/opengl/wiki/Debug_Output). As that page says,

    > Note that catching GL errors with this method is much easier than by using [glGetError() error checking](https://www.khronos.org/opengl/wiki/Error_Checking), as there is no need to sprinkle [glGetError()](https://www.khronos.org/opengl/wiki/GLAPI/glGetError) calls throughout the code. Even better, on some GL drivers, when a GL error occurs the message callback function is invoked on the same thread and within the very same call stack as the GL call that triggered the GL error (or performance warning).

    And it's trivial to use, just do `TGLFeatures.Debug := true` somewhere early in application initialization.

    Disclaimer: I'm not sure does Apple/iOS support `KHR_debug`.

- They also make the code harder to read. If we continue to add more and more `CheckGLErrors`, we will have rather complicated OpenGL code. So we prefer to remove them, before merging.

- Calling `glGetError()` in release builds is a performance hit. So if anything, they should not be done unconditionally.

tools/build-tool/data/ios/services/photo_service/cge_project_name/photo_service/PhotoService.m
```diff
@@ -27,7 +27,7 @@ @implementation PhotoService
 
 - (bool)messageReceived:(NSArray* )message
 {
-    if (message.count == 2 &&
+    if (message.count >= 2 &&
```
michaliskambi, line 30: Please make it exactly `== 3` (not only `>= 2`) and add a comment like

```c
// We also get MIME type from Pascal, though we don't use it now
// NSString* mimeType = [message objectAtIndex: 2];
```

This way it is self-explanatory why we have 3 arguments :)

src/base_rendering/castleglshaders.pas
```diff
@@ -1780,7 +1780,11 @@ procedure TGLSLProgram.AttachShader(const ShaderType: TShaderType; S: string);
         if (not HasLine('precision mediump float;', S)) and
            (not HasLine('precision lowp float;', S)) and
            (not HasLine('precision highp float;', S)) then
+          {$ifdef iOS}
```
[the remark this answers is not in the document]
michaliskambi: Sorry, I just read your PR description carefully, you already said explicitly _"TODO: when the scene uses shadows, there may be black noise on some surfaces. I tried increasing the float precision in the shader, which decreased the noise heavily, but it is still there. So please don't merge this PR now, it is here mainly to track the changes and this topic."_ .

OK, I know this is WIP. Let's find the fix for the shadow maps issues and then decide what precision/why is necessary.

tools/build-tool/data/ios/services/photo_service/cge_project_name/photo_service/PhotoService.m
janadamec1: Well, the actual code needs just 2 parameters and ≥ is more future proof (e.g. when somebody forgets to update all TPhotoService derivatives ;-)). But OK, you are the boss, changed to == 3 as in Android code.

src/scene/castlescreeneffects.pas
janadamec1: There is no KHR_debug on iOS, `TGLFeatures.Debug := true` does not help. glGetError and output to log is the only way to catch errors early enough (usually there is only the check at the end of getting OpenGL info). I had hard time to catch the proper place of the OpenGL error. OK, commented out in the next commit.

janadamec1:
So after proper testing shadows on Android, I can see the same issue as on iOS (but with much less intensity). At certain angles, there is a black noise over the top yellow face.

Android

Also the shadow completely disappeared after switching from Castle Model Viewer to another app (Photos) and returning back.

michaliskambi:
Thanks, in a way that's good you can reproduce the issue on Android, so things are less strange :) Possibly this can be fixed by just different scale / bias on shadow map? They depend on `glPolygonOffset` and it may work differently on different implementations (though in past I was able to find options that seemed to work on various platforms reliably...). Like

```
    bias 4
    scale 4
```

?

Note that I also did one shadow map fix yesterday, see https://forum.castle-engine.io/t/shadow-maps-in-2025/1923/23?u=michalis . It's probably unrelated to you (it doesn't change the quality of shadow maps; just is some cases shadow maps were not activated at all), but please upgrade your branch to be sure :)

As for Android building and `ppcarm`: If only building Android 32-bit is an issue, you can just build only for Android 64-bit. For development, it's naturally good enough, assuming your testing device is 64-bit. To do this, instead of

```
castle-engine package --target=android
```

just execute

```
castle-engine package --os=android --cpu=aarch64
```

This will make APK with only 64-bit (Aarch) architecture.

janadamec1:
I've tested it on a few other scenes (trees_river, projected_spotlight), and it works. It seems the most problematic is the shadow_spot_simple, probably because the light source is aligned to geometry. Other scenes work even with bias=4. Will also check shadows in my app with the library interface.

I have difficulties opening the complex scenes prepared in .zip (prototypes, textures) at the moment with ModelViewer-mobile on iOS, this will need to fixed in the code.

janadamec1:
You were right. The only variable that needs to be highp is the "shadowMapCoord". So I did StringReplace to  all occurrences of "const vec4 shadowMapCoord" and "const vec3 shadowMapCoord" to "const highp vec4 shadowMapCoord" and "const highp vec3 shadowMapCoord".

With this, I think RenderToTexture fix on iOS may be complete.

michaliskambi:
GitHub Actions should be good now, merged to master and to this PR (`ios-render-to-texture-fix` branch) necessary fixes.

michaliskambi (approved):
Thank you for the PR and patience :) I'm back on this, did some minimal changes, and want to do some final testing tomorrow and apply!

janadamec1 (commented):
I think this restoration of CGEApp_ContextProperties makes things worse again. I will have to test if the required context with Depth24-Stencil8 is set or not. If I remember correctly, this was a must, and that's why there is a repaint afterwards.

So please don't merge this now.

Why we have CGEApp_ContextProperties anyway? Context was created by the view, parameters read in CGE, then sent to the view via  CGEApp_ContextProperties again?

michaliskambi:
> I think this restoration of CGEApp_ContextProperties makes things worse again. I will have to test if the required context with Depth24-Stencil8 is set or not. If I remember correctly, this was a must, and that's why there is a repaint afterwards.

I have indeed a TODO to test whether I didn't make regression by my restoration of `CGEApp_ContextProperties` (that I've done in https://github.com/castle-engine/castle-engine/pull/674/commits/e246e99c6f8db091f549d904734ab74f3fb9db91 ).

But I didn't see explanation in code/PR why this was removed, so I assumed yesterday this was just a mistake.

And the default situation, if user didn't customize any `Window.Xxx` properties, should result in exactly the same values as you had hardcoded.

> Why we have CGEApp_ContextProperties anyway? Context was created by the view, parameters read in CGE, then sent to the view via CGEApp_ContextProperties again?

We don't want to just hardcode the context creation parameters. We want them to be controlled by properties of `TCastleWindow`, like:

 - `TCastleWindow.RedBits`,
 - `TCastleWindow.GreenBits`,
 - `TCastleWindow.BlueBits`,
 - `TCastleWindow.AlphaBits`,
 - `TCastleWindow.DepthBits`,
 - `TCastleWindow.StencilBits`,
 - `TCastleWindow.MultiSampling` (and thus `TCastleWindow.AntiAliasing`)

By removing usage of `CGEApp_ContextProperties`, we would remove this functionality. And e.g. users could not request multi-sampling, of extra depth/stencil precision (if possible on any hardware).

Of course, if a backend, like iOS, would really support only _one_ reasonable configuration, then we would have no choice. But this is really a last resort :)

michaliskambi:
I pushed to https://github.com/castle-engine/castle-model-viewer-mobile/ repo 3 more built-in demos, that shall enable us to test these features easier on mobile:

- Chopper + Stonehenge (shadow volumes)
- Shinto Shrine (shadow maps, bump mapping)
- Simple Shadow Maps (somewhat updated demo that you also used for tests in this ticket).

They are copies of analogous models from demo-models.

michaliskambi:
Testing on iOS (iPhone SE 2nd gen):

- Version from this PR works perfectly :) Screen effects (2 demos mentioned above), shadow maps (all 3 demos including "Simple Shadow Maps" that I committed to castle-model-viewer-mobile, see above) are perfect.
- On master, shadow maps don't work at all, we knew this. Rest (screen effects, shadow volumes) are OK.

Remaining for tomorrow:
- I will clone and test whether the "Unholy Society" FBO behavior changes. We had regression about it, which was holding OpenGLES at level 2 (not 3), in the past. 
    - Though likely this will not affect my decision whether to apply this PR. We never reproduced that specific "Unholy Society" problem in other applications / X3D files. And this PR fixes shadow maps in iOS, which is important. So even if applying this will cause "Unholy Society" regression -- at this point, it will be OK. I mean I will want to fix that regression, but it can wait, and in the meantime it's better to have CGE master that supports shadow maps on iOS.
- I will test does web target (it reuses some OpenGLES code) is still OK on this branch.
- @janadamec1  Of course I appreciate any testing from you!
    - As for the `CGEApp_ContextProperties`, I don't really see any change to it necessary -- it seems to behave exactly as it should, it sends correct values, and the `Window.XxxBits` should be honored, and the default is just equal to what you had hardcoded, from what I can see. Of course, any report of the contrary is welcome, maybe there's some bug I don't see and that is reproducible only on other devices.

michaliskambi:
> Remaining for tomorrow:

I decided to "push it" before going to sleep. Sorry for so many messages in this thread -- at various steps above, I thought "I will do this and then go to sleep", but then I decided to continue :) I hope to put Castle Model Viewer in TestFlight soon and finally push its release on AppStore.

1. Unholy test: Passed.

    - That is, the regression that I observed in the past (with migrating to OpenGLES 3)... is no longer there. Thanks to your improvements, I assume.

    - It is hard to make a precise test, because "Unholy Society" has now other problems with CGE master, so completely unrelated to this PR. We render some textures wrong, before in CGE master and after this PR. I will have to debug it (and it may be some Unholy-specific issue now, or just me doing sthg stupid now... it's been a while since I build that project :) ). Anyhow, this is surely unrelated to this PR changes, since problems are already on master.

2. Web building and run of various examples, including screen effects: all OK.

    - Note that shadow maps on web are still a TODO, this is known.

3. Added one more demo to castle-model-viewer-mobile: screen effect with SSAO (on deliberately "sterile" level where it's very noticeable).

    - Web: fails before and after this PR, which is OK, this is TODO on web for unrelated reasons :) The important thing is that this PR doesn't cause regression.
    - Android: OK before and after, it seems.
    - iOS: doesn't work before, works OK after this PR.

4. As the SSAO effect seemed very subtle on mobile, I also replicated it with dummy shader visualizing depth: 

    ```
   void main (void)
    {
      float d = screenf_get_original_depth();
      // no pow on OpenGLES
      d = d * d * d * d * d * d * d * d * d * d;
     gl_FragColor = vec4(d, d, d, 1.0);
   }
   ```

    With this,  I can confirm for 100% that screen effects with depth were broken on iOS before, and are now fixed :)

So, all OK according to my tests. Merging! 

As said above, I of course still welcome your testing. I don't _think_ (from my looking at logs, code, and tests above) that anything is broken, and `CGEApp_ContextProperties` works as expected in my tests.

janadamec1:
I've tested it now. OK, the shadows basically work. But there is an issue with noise again, that was treated using shadowMapPrecision to highp.

And you were right, CGEApp_ContextProperties is OK, and I am still getting the correct values from OpenGLInfo. Maybe it was needed only in iOS library tester.

michaliskambi:
> I will try to fix it over the weekend, and post news on https://castle-engine.io/wp/ .

I will postpone fixing it (SSAO shader on OpenGLES) -- no time this weekend, unfortunately.

I have created a dedicated issue about SSAO shader on OpenGLES, https://github.com/castle-engine/castle-engine/issues/692 . I will post further progress there (as it gets off-topic for comments in this PR). Anyone interested can subscribe there (and/or help me investigate it :) ).

Issue #678, opened 2025-06-23
tmssoftware/smartsetup integration


agallero:
Hi,

We are developing a tool for compiling packages in Delphi. It was done for our own cases, but some time ago we decided to open source it so if people want to use it for their own projects/components, they can.

I've been exploring many open source projects, we have a list of products already adapted here: https://github.com/tmssoftware/smartsetup-registry

And while I am not sure Castle Engine is really a match for this (As it really needs a more complex install that installs the castleEngine IDE), well, I wanted to try it. I had to make the following modifications to make it work:

1. In Delphi11 or older, I get:
```
R:\castle\Products\castle-engine\src\services\steam\castlesteam.pas(224): error E2056: String literals may have at most 255 elements
```
Delphi12 compiled fine, but all other delphis are not compiling. 
So, to fix it, I just broke the string in 2: instead of 'abcd', use 'ab'+'cd'

2. There is no libsuffix in the dpks. This is one that I am not sure if might break other stuff (is the CastleEngine IDE expecting castle_engine.bpl instead of castle_engine290.bpl?  But we need to have a libsuffix so we can install in more than one IDE at the same time. So I modified all Dpks to add a line with libsuffix:

```
{$DESCRIPTION 'Castle Game Engine - Delphi IDE design-time-only package'}
{$LIBSUFFIX AUTO}
{$DESIGNONLY}
{$IMPLICITBUILD ON}
```

Note that Libsuffix auto is only correctly supported in D11+  (which I am not sure matters, since as the code was now, it only compiled in D12) If you want to add support for older versions,create extra packages for each version supported. By now, I am just happy supporting D11 and newer 😄 

The good news is that with the fix it looks like it is working (I am no expert in castle engine, but I couldn't see anything wrong). I could get CastleEngine installed in D11 and D12 in 1 minute:

![Image](https://github.com/user-attachments/assets/a0a564ab-eae4-4c92-bb43-7d0ef38eae34)

And it looks ok:

![Image](https://github.com/user-attachments/assets/b39622b1-c4fe-441f-81c7-76d7c62aae41)

By the way, there is a point 3: As you can see in my previous screenshot, I compiled the packages for both 32 and 64 bit. But I had to skip the design package in 64 bits because it doesn't compile, so I couldn't install it in the Delphi 12.3 64 bit IDE.

I got the error:
```
[...]

PR #680, opened 2025-06-24
64bit ide

agallero:
Adds libsuffix


agallero:
Opps... I actually forgot to test it from the IDE itself, and it indeed doesn't work. After some investigation, the reason seems to be:

The dcc32 compiler is using the LIBSUFFIX we added in the dpk, so compiling from the command line works as expected. That's what I tested. But the IDE uses a `DLL_SUFFIX` entry in the dproj instead, so it believes it has no libsuffix. (you can see this in your screenshot:  

It says: castle_engine.bpl instead of castle_engine290.bpl as it should:



So my guess is: The ide believes there is no LIBSUFFIX, while the compiler sees a 290 LIBSUFFIX. This causes the IDE to not be able to find the package the compiler compiled in order to install it. (IDE thinks it compiled a castle_engine.bpl, but the compiler compiled castle_engine290.bpl).

Sadly I don't think this is going to work. Going back to see what the original author from where I got this idea did, I see that he provides 2 packages: One with LIBSUFFIX auto for D>=10.4, with both LIBSUFFIX in the dpk and DLLSUFFIX in the dproj. **but**, this is the important part, there is no  dproj provided for D<10.4: https://github.com/pyscripter/python4delphi/tree/master/Packages/Delphi/Delphi%2010.3-

But while this might work (Delphi recreates the dprojs if they don't exist), it is not a nice solution because you lose all the settings in the dproj. 

So this isn't really a feasible solution. 

The good news, is that in the meantime since I wrote my last post, we added support in SmartSetup for automatically adding libsuffix if it is not in the dpk/dproj when we compile (so we can support more libraries), and so we can still support castle packages if you don't add the LIBSUFFIX thing. (you would still have to merge the second part of this request, for 64 bit delphi ide. My idea was to do pull requests, but I mixed it up and both commits ended up in the same one).

Still, I don't know what to say: If we support it by adding LIBSUFFIX when you compile from smartsetup, and Castle's packages don't, you might end up with a mess of castle_engine.bpl and castle_engine280 in the disk. 

Also, irrespective of SmartSetup, Castle Packages should have a LIBSUFFIX, that's the way things are done in Delphi since Delphi 6, and the only way to install Castle Packages in say both D11 and D12 at the same time. (See for example this excellent post by Vincent Parrett about it: https://www.finalbuilder.com/resources/blogs/advice-for-delphi-library-authors )

But to support LIBSUFFIX, you would have to provide at least 4 packages (*):
D12.2
D12.3
D12.4
D11+

(*) You could get away with 3, because support for libsuffix auto was added in D12.4.1, so you could have a D12.4.1+ but that would drop support for D12.4.0 users, and also 10.4.1 was buggy, see
https://en.delphipraxis.net/topic/3427-delphi-1041-libsuffix-auto/

I don't see any good solution here, I don't know if you see anything else.

michaliskambi:
Same as in https://github.com/castle-engine/castle-engine/issues/678#issuecomment-3054358047 , just wanted to say for now "thank you" and I have this in TODO for this weekend to write proper answer.

>  (you would still have to merge the second part of this request, for 64 bit delphi ide. My idea was to do pull requests, but I mixed it up and both commits ended up in the same one).

No worries, change to add 64-bit design-time is in separate commit ( https://github.com/castle-engine/castle-engine/pull/680/commits/5c96762a58c16fcd7b089128466a89e01316a371 ), I can cherry-pick it without problems. That's also something I want to test around this weekend.

PR #684, opened 2025-07-28
Comment about QOpenGLWindow vs QOpenGLWidget in Qt library example.

janadamec1:
Improve comment about the difference when using QOpenGLWindow and QOpenGLWidget in Qt library example.

This just clarifies the changes made in https://github.com/castle-engine/castle-engine/pull/682


michaliskambi (approved):
Thank you!

PR #686, opened 2025-08-05
Implements UISceneDelegate in iOS targets

janadamec1:
It's about the UIScene lifecycle (added in iOS 13 in 2019). Xcode started issuing the following warning:

`CLIENT OF UIKIT REQUIRES UPDATE: This process does not adopt UIScene lifecycle. This will become an assert in a future version.`

Basically, in the future AppDelagate will be deprecated, removed. Apple lets to support multiple app instances (via UIScene, not possible with AppDelegate), so that's we they want to remove the logic in AppDelegate.

What is related to CGE (Services and AppDelegate):
- applicationDidEnterBackground
- applicationDidBecomeActive
- application:didFinishLaunchingWithOptions:
- application:openURL:options:

Once implementing `UISceneDelegate`, the above functions are not called on AppDelegate, but on SceneDelegate. So in this Pull Request, all changes are related to this.

As CGE supports only 1 window now, there were no problems with background-foreground transitions, we can keep the same logic.

To keep the compatibility with iOS < 13, I introduced the new `openURL` function on Services (with different parameters). So each service must implement both. I did this in FacebookService, as this is the only service using this.

The main problem was with `application:didFinishLaunchingWithOptions:`. We create the services with viewController on parameter. When using SceneDelegate, the main viewController is created after application:didFinishLaunchingWithOptions:, se we cannot call application:didFinishLaunchingWithOptions: on them (as the services instances don't exist).

I solved this by calling application:didFinishLaunchingWithOptions: just after the window becomes active in SceneDelegate. The proper functionality of FacebookService should be tested after this change.


michaliskambi:
> I will make a test of the Facebook service in action (and maybe add a test application for it to CGE) around this weekend and then apply.

( I mean, unless you or anyone else wants to do this instead of me :) If you will tell me that you tested iOS application + Facebook service and it works, and "Facebook Analytics" reflects usage of app (that is the main purpose of [facebook integration](https://github.com/castle-engine/castle-engine/blob/master/tools/build-tool/data/ios/services/facebook/README.adoc) so far) and `TFacebook.LoginButton` works -> then I will be happy to merge it, without testing on my side. )

janadamec1:
I understand. And sorry, I have no experience with FB SDK, Analytics, Login, etc. I think the only app implementing this, is one of yours. (BTW Facebook has completely new SDK, so I am not sure if the old SDK is even supposed to work at all].

> ( I mean, unless you or anyone else wants to do this instead of me :) If you will tell me that you tested iOS application + Facebook service and it works, and "Facebook Analytics" reflects usage of app (that is the main purpose of [facebook integration](https://github.com/castle-engine/castle-engine/blob/master/tools/build-tool/data/ios/services/facebook/README.adoc) so far) and `TFacebook.LoginButton` works -> then I will be happy to merge it, without testing on my side. )

michaliskambi:
> I understand. And sorry, I have no experience with FB SDK, Analytics, Login, etc. I think the only app implementing this, is one of yours. (BTW Facebook has completely new SDK, so I am not sure if the old SDK is even supposed to work at all].

That may be (that I'm the only one with application actually using it), but then I should fix it :) Looking back, we (we = "Cat-astrophe Games" company) used this for our old "Escape from the Universe" game (mobile version), though it's not available in AppStore anymore, it's only for Nintendo Switch now. In our later games, we didn't use it.

I'll create `examples/mobile/facebook/` to show and test this easily, with docs for everyone how to create FB page and connect with it. And I'll see if it works at all (also before your changes).

PR #704, opened 2026-01-04
removed dupe EGLContextNotPossible, it caused wrong reinit flow

phomm:
correct Exception definition is  in CastleInternalContextBase, and thus a name clash, which caused 
procedure TCastleWindow.Open(const Retry: TGLContextRetryOpenFunc);
to skip processing of Retry with lowered requirements, when exception was fired from 
CastleInternalContextWgl => TGLContextWGL.InitializeCore => SetPixelFormat_WGLChoose


phomm:
tested on windows, with all engine code recompile, not 100% sure if this is suitable fix for other OSes, but let's see what CICD could catch

PR #715, opened 2026-04-20
Android http error responses

phomm:
Corrected Android error handling, Requests with Http response status 400+ are treated as TCastleDownload.DownloadStatus = dsError

also a small compilation fix after Web HttpRequests branch changes


michaliskambi:
Thank you! Applying.

I see you also added rewinding `Contents.Position` and updating `DownloadedBytes` / `TotalBytes` on `dsError`, just like on `dsSuccess`, and that's indeed cool too, we want this to behave consistently.

PR #717, opened 2026-04-23
Support Windows on ARM (aarch64-win64)

janadamec1:
Castle Engine can be built using FPC>=3.3.1, so let's try to support this platform directly.

`castle-engine --os=win64 --cpu=aarch64 compile`

At the moment, I built all external libraries that we ship on Windows (ogg, vorbis, zlib, OpenSSL, ...).

Still we need to add test runner on GitHub and more to finalize this to be perfect. But theoretically, with those DLLs, the apps built with CGE may already work on Windows on ARM.

TODO:

- add a 'windows-11-arm' runner to GitHub Actions (_In principle, this just means adding "windows-11-arm" to the
"runner" list in ".github/workflows/test-and-pack-runner-native.yml"... and then likely dealing with all errors that appear:)_ )


michaliskambi:
I looked at code that is copying the DLLs alongside EXE -- we likely have a bug there, and it needs to be adjusted :)

This is in `TProjectDependencies.DeployFiles`, https://github.com/castle-engine/castle-engine/blob/effa2cba7559470acbf20175436812aa40f9d0ee/src/files/tools/castleinternaltools.pas#L341 , used by both build tool (called from `tools/build-tool/code/toolproject.pas`) and inside Delphi IDE. Our code says there:

```delphi
procedure TProjectDependencies.DeployFiles(const DeployPlatform: TDeployFilesPlatform;
  const Files: TStrings);
var
  Prefix: String;
begin
  case DeployPlatform of
    dpWin32:
      begin
        { Matches CPUToString(CPU) + '-' + OSToString(OS) for win32. }
        Prefix := 'external_libraries/i386-win32/';
        ....
      end;

    dpWin64:
      begin
        { Matches CPUToString(CPU) + '-' + OSToString(OS) for win32. }
        Prefix := 'external_libraries/x86_64-win64/';
        ....
      end;
  end;
end;
```
 
This is wrong now (assumes the only CPU on Win64 is x86_64). And it receives only `TDeployFilesPlatform`  which can be `(dpWin32, dpWin64)`, not full OS / CPU info. Our `TDeployFilesPlatform` has a TODO about it

```
    TODO: In the future, we could use TOS and TCPU and make it general.
    For now, this is practically only used by Windows platforms that need
    some DLLs, on other platforms we're fine with just using system libraries. }
```

So, I think we have to address that TODO :) Possibly the `TDeployFilesPlatform` enum should be just thrown away, it's not flexible enough. And `TProjectDependencies.DeployFiles` should receive `OS`, `CPU` as 2 parameters (just like various build tool API does) and then it should just set 

```delphi
Prefix := 'external_libraries/' + CPUToString(CPU) + '-' + OSToString(OS);
```

and the `case` inside `TProjectDependencies.DeployFiles` should copy the DLLs suitable for Aarch64 (the same as for x86_64, from what I understand).

janadamec1:
OK, done.

> I looked at code that is copying the DLLs alongside EXE -- we likely have a bug there, and it needs to be adjusted :)
> ...
> So, I think we have to address that TODO :) Possibly the `TDeployFilesPlatform` enum should be just thrown away, it's not flexible enough. And `TProjectDependencies.DeployFiles` should receive `OS`, `CPU` as 2 parameters (just like various build tool API does) and then it should just set

PR #728, opened 2026-08-29
Set default Android API LVL to 36 required by Google Play

and3md:
Hi, small patch to set default API level to 36 required by Google Play from 31 August 2026. Tested in my mobile game, everything seems to be working.


michaliskambi:
Thank you, merging! :)

PR #729, opened 2026-09-13
admob: user consent (UMP), and ads SDK off a sunset version

kabiri:
Two things the admob service could not do.

## 1. There was no consent flow at all

Google requires a consent management platform for users in the EEA and the UK,
and requires that no ad is requested before such a user has answered. A search
of the service for `ConsentInformation`, `UMP` or `consent` returned nothing.

`TAds.InitializeAdMobWithConsent` gathers it with Google's own User Messaging
Platform: the form is shown when it is required, and the ads SDK is initialized
only once the answer allows requesting ads. Where no consent is required, e.g.
outside Europe, nothing is shown to the user and the ads initialize
immediately.

Everything that touches the ads SDK now waits for that — including the first
interstitial and rewarded load, and a banner — because loading an ad is already
requesting one. In `ServiceAdMob.java` this is `initializeMobileAdsIfAllowed`,
called both when the Pascal side initializes and when the consent flow ends,
since those two can arrive in either order.

**Backwards compatible.** `InitializeAdMob` behaves exactly as before: no
consent gathering, no waiting, everything initialized on the spot. Only a
caller that asks for the new method gets the new path.

New API:

- `TAds.InitializeAdMobWithConsent` (two overloads, the second one taking the
  debug-geography parameters)
- `TAds.OnConsentGathered`
- `TAds.CanRequestAds`
- `TAds.PrivacyOptionsRequired`
- `TAds.ShowConsentPrivacyOptions` — the entry point Google requires the
  application to offer when the CMP says so
- `TAds.ResetConsent` — testing only

## 2. play-services-ads was on a sunset version

22.1.0, which Google sunset on 2026-06-30: *"ad traffic from that version is at
risk of receiving automatic no fill due to stopped ad serving"*
(https://developers.google.com/admob/android/deprecation). Now 25.4.0, which is
supported until 2028.

That version requires `minSdkVersion 23`, so a project using this service must
declare `<android min_sdk_version="23" />`. Documented in the service README.

## Verified on a device

Galaxy A32, Android 13. With a debug EEA geography the form appears, is
answered, and the answer is stored as a real TCF string:

Receive consent action: ?action=dismiss&args={"status":"CONSENT_SIGNAL_SUFFICIENT"}
Writing to storage: [IABTCF_TCString] CQqbdwAQqbdwAEsACBENCwFoAP_gAEPgACaI...
Ads: Consent gathered. Ads allowed: True. Privacy options required: True.

The ordering — the part that actually had to be proven — is visible in
Android's own `ConnectivityManager` stack trace, which shows `MobileAds.initialize`
running from inside the form's dismissal callback:

UserMessagingPlatform.loadAndShowConsentFormIfRequired(ump@@4.0.0)
-> ServiceAdMob$2$1.onConsentFormDismissed(ServiceAdMob.java:209)
-> ServiceAdMob.consentGathered(248)
-> ServiceAdMob.initializeMobileAdsIfAllowed(155)
-> MobileAds.initialize(play-services-ads@@25.4.0)

Outside the EEA (tested in Turkey, without the debug geography) no form is
shown and the ads initialize immediately, with `IABTCF_gdprApplies: 0`.

## Notes

- `play-services-ads` 25.4.0 is compiled with Kotlin 2.3.0 while the build
  tool's Gradle setup uses 2.1.0. Gradle prints a metadata warning and the
  build succeeds. A Kotlin bump may be worth a separate look.
- The service README documents the consent flow, the `min_sdk_version`
  requirement, and how to test the form outside the EEA
  (`ConsentDebugSettings` plus a device hash from logcat).
- `AdSize.getCurrentOrientationAnchoredAdaptiveBannerAdSize` is deprecated in
  the new ads SDK. Left as it is here — it still works, and replacing it is a
  banner change rather than a consent one.

## Open to your judgement

- **minSdk.** Raising the requirement to 23 affects every project using this
  service. The alternative is pinning 23.6.0, which keeps minSdk 21 but is
  itself deprecated with a 2027-06-30 sunset. I took the supported version;
  happy to change it.
- I have not tested what a project with `min_sdk_version="21"` sees — this
  project set 23 from the start. I expect a manifest merger error rather than
  anything subtle, but I have not confirmed it.


kabiri:
Following up on the last point of the description: I have now tested it.

A project that keeps `min_sdk_version="21"` gets a Gradle failure, not
anything subtle, and the message names both the cause and the fix:

> Manifest merger failed : uses-sdk:minSdkVersion 21 cannot be smaller than
> version 23 declared in library [com.google.android.gms:play-services-ads:25.4.0]
> as the library might be using APIs not available in 21
>
> Suggestion: use a compatible library with a minSdk of at most 21,
> or increase this project's minSdk version to at least 23

So the upgrade path for an existing project using this service is one line in
`CastleEngineManifest.xml`, and the build tells them exactly that line. No
silent breakage, and nothing at runtime.

Worth weighing against the alternative: pinning play-services-ads 23.6.0 keeps
minSdk 21 and would need no change from anyone, but 23.x is already deprecated
with a 2027-06-30 sunset, so it moves the same decision a year out.

src/services/castleads.pas
```diff
@@ -336,22 +446,45 @@ procedure TAds.TAdNetworkHandler.FullScreenAdClosed(const WatchedStatus: TAdWatc
...
 
 procedure TAds.TAdMobHandler.ReinitializeJavaActivity(Sender: TObject);
 begin
+  { The consent has to be gathered before the ads SDK is initialized,
+    so this message has to be sent before the initialize message.
+    The consent is gathered again after the Java activity was recreated,
+    as the Java side state is gone then. }
+  if FGatherConsent then
+    Messaging.Send(['ads-' + Name + '-consent-request',
+      IfThen(FConsentDebugForceEea, 'true', 'false'),
```
michaliskambi: Use `TMessaging.BoolToStr` which is guaranteed to work with Java counterpart `stringToBoolean`.

michaliskambi:
Thanks! Looks good!

Summary:

- I did 3 small changes described below. I pushed them to your PR branch, please test them on your side (as I admittedly don't have any test project using ads now; so, please test that I didn't break any functionality in your use-case :) ). See point 1 below for overview what/why I changed.

- Let's introduce a way for service to bump the minimum SDK version, automatically. This would prevent any compat-breaking by merging this (people using the service would just automatically have minimum SDK requirements raised, which is acceptable). See point 2 below for details.

Details:

1. I made small commits on top.

1.1. use `TMessaging.BoolToStr`, `TMessaging.MessageToBoolean`

1.2. use `ServiceAbstract.splitString` with `allowEmptyList=true`

- This is important, as Pascal sends '' when there are no hashes, and in Java `"".split(",")` was giving an array of length 1 with one zero-length string. This meant that `debugDeviceHashes.length != 0` was always `true` in your version. So a condition in `consentRequest` was always building debug `ConsentDebugSettings`.

- Adding empty hash was protected by `!deviceHash.equals("")`, but it's cleaner to just make `debugDeviceHashes` list empty, when Pascal code passes empty list:)

1.3. I adjusted a bit comments, to explain better why order is important. Consent gathering happens asynchronously from `ads-admob-consent-request`, so when we call `ads-admob-initialize` -> consent is still not gathered. But it's OK, because

- processing `ads-admob-consent-request` sets `consentGathering` to true, and this prevents `initializeMobileAdsIfAllowed` from acting.
- And later `initializeMobileAdsIfAllowed` will be called again at the end of `consentGathered`.
- I changed comments to express this better hopefully.

2. For backward compatibility: People using AdMob right now -> will have to adjust, adding the `<android min_sdk_version="23" />`. You documented it clearly above and in README, this is very good, but still I think we can do even better:) Let's make old projects "just work". Bumping the required min SDK version (with proper message) is acceptable, and better than breaking, in this case.

- Note that we already had/have cases of services doing this. `game_analytics` also requires this, requiring 26. So it would benefit from this possibility too.

- I propose adding to `CastleEngineService.xml` (optional in top-level of service directory) an option like `min_min_sdk_version` (I know, double `min` feels weird, but it's indeed a minimum of a minimum :) ). So you would then create file `tools/build-tool/data/android/services/admob/CastleEngineService.xml` with contents like this:

    ```
    <?xml version="1.0" encoding="utf-8"?>
    <service min_min_sdk_version="23">
    </service>
    ```

    and make build tool obey this. To do this:

    - Reading services XML files is in `tools/build-tool/code/toolservicesoperations.pas`, so you need to add there reading `min_min_sdk_version`

    - and apply this in `TCastleProject.AddMacrosAndroid`, that right now uses `AndroidMinSdkVersion` which just looks at manifest value, but should also check all services for Android. (I don't have an opinion whether this change should happen inside `AndroidMinSdkVersion` implementation or when the `AndroidMinSdkVersion` is called, both options seem good).

    - There is a catch here: right now `PackageAndroid` calls first `GenerateFromTemplates` and only later `PackageServices`, and the latter loads actually `CastleEngineService.xml`. You will have to split it, like add `LoadServices` that is called before `GenerateFromTemplates`, so that we can calculate proper min SDK number when generating manifest.
    - Also make minimal reporting (_Bumping project's Android min SDK version to XXX because service YYY requires it_) and check that minimum is not > `target_sdk_version`.

- Once it works, you can simplify docs of `admob` and `game_analytics`: developers using them will not need to do anything.

Let me know if you're OK doing AD 2 (`min_min_sdk_version`) above:) If not, I can also do it, though it may take me longer then:)

kabiri:
Thanks for the review and the fixes!

1. Your commits. I pulled them and applied them to my engine copy. Good catch on "".split(","): I hadn't noticed the debug consent settings were built every time. I'll rebuild my game with your changes and test the consent flow again on a real device (Galaxy A32, Android 13): the form, the ads starting after consent, and the privacy options. I'll report back here.

2. min_min_sdk_version. Yes, happy to do it, and it's already implemented. I'll push it to this branch once testing is done:

toolservicesoperations.pas: new LoadServices, which reads each service's CastleEngineService.xml once, including the new optional min_min_sdk_version attribute on <service>. PackageServices now takes that loaded list, so the XML isn't read twice. iOS uses the same path, with no change in behaviour.
New ApplyServicesAndroidMinSdkVersion. It is called in PackageAndroid before GenerateFromTemplates, so ${ANDROID_MIN_SDK_VERSION} is already raised when the manifest and build.gradle are generated. It prints Bumping project's Android min SDK version to X (from Y) because service "Z" requires it. It fails with a clear message if the result is higher than target_sdk_version.
TCastleProject.AndroidMinSdkVersion returns the higher of the manifest value and the new AndroidMinSdkVersionFromServices. I put the logic there so every user of the macro gets it.
New CastleEngineService.xml for admob (23) and game_analytics (26). Both READMEs now say the build tool handles it and nothing needs to be done.
Tested with a minimal project using admob and no min_sdk_version: [result: the bump message, and minSdkVersion 23 in the generated build.gradle].

kabiri:
Thanks for the review!

1. Your commits: tested on a Galaxy A32 (Android 13), debug geography EEA, app data cleared. Ads wait for consent, the form shows, AdMob initialized comes only after it is answered, and the rewarded ad plays. With an empty hash list no debug settings are built. Good catch on "".split(",").

One note for testers: without adb shell pm clear first, AdMob initialized can appear before the form is answered. canRequestAds() uses the consent saved in the previous session. That is expected behaviour.

2. min_min_sdk_version: pushed (4efa055d4b).

LoadServices reads every service's CastleEngineService.xml before GenerateFromTemplates, and PackageServices reuses that list.
TCastleProject.AndroidMinSdkVersion takes the higher of the manifest value and what the services need. The bump is reported, and it fails if the result is above target_sdk_version.
admob requires 23 and game_analytics 26, and their READMEs are simplified.
Tested with a project using admob and no min_sdk_version: Bumping project's Android min SDK version to 23 (from 21) because service "admob" requires it, minSdkVersion 23 in build.gradle, and the APK builds.

kabiri:
Just a heads-up: I have a few more AdMob changes ready (letting the game know when a rewarded ad is loaded, and a fix for a rewarded ad that fails to show never reporting that it closed).
I'll send them as a separate PR once this one is merged, so the two don't get mixed up.

michaliskambi (approved):
All looks good, thank you -- applying! 

I in particular like `Project.AndroidMinSdkVersionFromServices` as an independent value from `Project.AndroidMinSdkVersion`,  this allows to display the `Bumping project...` message whenever any service has requirements larger than the project min SDK. So we will see all "bumping..." messages from services that do this, regardless of the order in which services are declared.

P.S. I can see you use Claude for this, but I have no objections -- this is good work, and I believe it follows our [AI guidelines](https://castle-engine.io/ai). Thanks!

kabiri:
Yes, I got some help from Claude with the Java part. That’s why I added the following line at the end of the commit message:
`Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>`
I’m always transparent about using AI in all my work.
However, since this is your repository, I should have asked for permission first. Sorry about that.

michaliskambi:
> However, since this is your repository, I should have asked for permission first. Sorry about that.

It's all good, no need to apologize, and no worries -- you didn't need to ask before using AI :) I wrote our [guidelines](https://castle-engine.io/ai) to state how we should use / not use AI, and I believe your work follows this, the result is solid.

